PR #3, opened 2022-07-10
Add extra steps for monterey

Willyfrog:
Added pyenv file for helping install a working version
Added notes for readme regarding use of nvm and pyenv
Upgraded minimum version to node 18


zefhemel (approved):
Super awesome, thanks!

PR #4, opened 2022-07-10
Add link command

Willyfrog:
Add links with cmd/ctrl+shift+k (feel free to change it afterwards)

It tries to detect if the selected text should be wrapped as url or as display text. If no text is selected, then it adds an empty link.

Future improvements: 
- better/flexible url detection (right now it requires having a protocol)
- maybe use a modal as other tools do that show the display text and the url and allows for editing
- ~~move caret to first empty section.~~ Done

### DIFF
--- a/packages/plugs/core/core.plug.yaml
+++ b/packages/plugs/core/core.plug.yaml
@@ -215,6 +215,12 @@ functions:
     path: ./text.ts:numberListifySelection
     command:
       name: "Text: Number Listify Selection"
+  linkSelection:
+    path: ./text.ts:linkSelection
+    command:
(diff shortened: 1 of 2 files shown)


zefhemel:
Nice! Tested it, seems to work. Thanks!

By the way, not exactly the same, but try pasting a link when you have text selected, it will auto link that text.

Willyfrog:
> Nice! Tested it, seems to work. Thanks!
> 
> By the way, not exactly the same, but try pasting a link when you have text selected, it will auto link that text.

I noticed, but I realized I had a display text without the link so I thought it would be nice to have that.

PR #45, opened 2022-07-23
Enable writing on the settings page from plugs

Willyfrog:
The feature to read a settings is super nice, but wouldn't it be nice to allow plugs to write their own settings?

Also in this PR:
- Added an example use with the preview panel.
- Fixed a bug preventing having a setting to false


packages/plugs/lib/settings_page.ts
```diff
@@ -10,13 +11,16 @@ import { readYamlPage } from "./yaml_page";
...
+const SETTINGS_TEMPLATE = `This page contains settings for configuring SilverBullet and its Plugs:\n\`\`\`yaml\n\`\`\``; // might need \r\n for windows?
+
 export async function readSettings<T extends object>(settings: T): Promise<T> {
   try {
-    let allSettings = (await readYamlPage("SETTINGS", ["yaml"])) || {};
+    let allSettings = (await readYamlPage(SETTINGS_PAGE, ["yaml"])) || {};
     // TODO: I'm sure there's a better way to type this than "any"
     let collectedSettings: any = {};
     for (let [key, defaultVal] of Object.entries(settings)) {
-      if (allSettings[key]) {
```
Willyfrog, line 19: this was driving me crazy for a bit as it was preventing any value to be falsy and using the default

Willyfrog:
if this is fixed, I'll use it for fixing https://github.com/Willyfrog/silverbullet-backlinks/issues/3

packages/plugs/markdown/markdown.plug.yaml
```diff
@@ -14,3 +14,7 @@ functions:
       - editor:updated
       - editor:pageLoaded
       - editor:pageReloaded
+  switchSide:
+    path: "./preview.ts:switchSide"
+    command:
+      name: "Swith Preview to other Sidebar"
```
Willyfrog, line 20: Not super useful, but I used it as a test tool for the function and provides a simple example on how to use read and write settings function within the main repo.

packages/plugs/lib/settings_page.ts
```diff
@@ -31,3 +35,24 @@ export async function readSettings<T extends object>(settings: T): Promise<T> {
...
+ * If the SETTiNGS page doesn't exist it will create it.
+ * @param settings 
+ */
+export async function writeSettings<T extends object>(settings: T) {
+  let readSettings = {};
+  try {
+    readSettings = (await readYamlPage(SETTINGS_PAGE, ["yaml"])) || {};
+  } catch (e: any) {
+    console.log("Couldn't read settings, generating a new settings page");
+    flashNotification("Creating a new SETTINGS page...", "info");
```
zefhemel, line 50: I wouldn't put UI stuff in something generally usable (from both client and server). It will break when using this API in a server context (there's no implementation of `flashNotification` there). Since you are not using `await` here it will silently fail (I think), but still.

packages/plugs/lib/yaml_page.ts
```diff
@@ -40,3 +40,42 @@ export async function readYamlPage(
...
+  } catch {
+    // page doesn't exist, so let's create one.
+    text = templateOnEmpty;
+  }
+  const tree = await parseMarkdown(text);
+  // if we use any other language than yaml... how to create it?
+  const doc = new YAML.Document();
+  doc.contents = properties;
+  // generate a new node for the properties
+  const newCode = `\`\`\`yaml\n${doc.toString()}\n\`\`\``;
```
zefhemel, line 62: I think we can roughly take two approaches to this `writeYamlPage` API:

1. "Let's surgically replace settings and leave everything as is" that means: let's keep the page text in place, comments in YAML, ideally, indentation etc. I'm not sure the yaml package we use is setup for this (preserving whitespace etc.) so this could be quite a challenge.
2. Let's just blow up the whole page and write it from scratch.

Since (1) is hard and not super important right now, I'd take the second approach. In which case you can get away without any fancy markdown parsing and rewriting, and go straight to the `writePage(pageName, newCode)` of this `newCode` variable (except maybe base it on the passed in template), because that's the core of it.

Or am I missing something?

packages/plugs/markdown/markdown.plug.yaml
zefhemel: Yeah it's a bit silly, but  also don't have a better sample use case right now.

zefhemel (changes requested):

packages/plugs/lib/yaml_page.ts
Willyfrog: Yeah, I don't expect many people to write their own config, and also this will fail on a specific use case where the user documents the settings, as it is only considering one single yaml section. Eventually I'd love to support that, but it will probably make things much more complex so I'm ok with rewriting it.

I'd then also add a warning in the template that any unexpected node will be deleted.

packages/plugs/lib/settings_page.ts
Willyfrog: > there's no implementation of flashNotification there.

I still think it is important to let the user know that something has gone wrong. In this one is about creating a new settings page so user is not surprised and could be removed (although I prefer to let them know that there is a new page in their space). But a couple of lines later I use the same mechanism for error/success. if a setting fails to update I'd like the user to know.

is there a way to tell if this is client/server code? and once we know we are in the server, how do we tell the user about anything happening? (other than throwing errors)

packages/plugs/markdown/markdown.plug.yaml
Willyfrog: If we find a better example, we should remove this one :)

packages/plugs/lib/util.ts
```diff
@@ -12,3 +14,21 @@ export async function replaceAsync(
...
+export function isBrowser() {
+  return !isServer();
+}
+
+export async function notifyUser(message: string, type?: "info"|"error") {
+  if (isBrowser()) {
+    return flashNotification(message, type);
+  }
+  const log = type === "error" ? console.error : console.log;
+  log(message); // we should end up sending the message to the user, users dont read logs.
```
Willyfrog, line 32: not super happy with this, but at least it will prevent breaking.

packages/plugs/lib/settings_page.ts
zefhemel: There’s no mechanism right now to know if you’re being run on the client or server (although you can mark a function with `env: server` in the yaml file to make sure that it does only run in one env). What I usually do is throw an exception, these exceptions transfer across client-server boundaries. So in an API like settings_page, I’d simply throw an exception when things go wrong.

Then on the client side I can catch this exception, even if it occurs on the server when I invoke it remotely via `invokeFunction(“server”, …)` I can put a try-catch around it on the client side, and use a `flashNotification` in the catch-clause to show the error.

packages/plugs/lib/util.ts
zefhemel: See my other comment, I’d just `throw Error(“FAIL”)` and catch it on the call site.

packages/plugs/markdown/markdown.plug.yaml
zefhemel: You can remove this and just use it in your backlinks plugin :)

packages/plugs/lib/settings_page.ts
zefhemel: You’re welcome 🤡

packages/plugs/lib/settings_page.ts
```diff
@@ -31,3 +36,22 @@ export async function readSettings<T extends object>(settings: T): Promise<T> {
...
+ * Convenience function to write a specific set of settings from the `SETTINGS` page.
+ * If the SETTiNGS page doesn't exist it will create it.
+ * @param settings 
+ */
+export async function writeSettings<T extends object>(settings: T) {
+  let readSettings = {};
+  try {
+    readSettings = (await readYamlPage(SETTINGS_PAGE, ["yaml"])) || {};
+  } catch (e: any) {
+    await notifyUser("Creating a new SETTINGS page...", "info");
```
zefhemel, line 50: Ok, now I see why you need the `notifyUser` thing, not just for errors but for "success" messages too. Alright. Ideally we would have a way to "push" notifications to the client, but this is not really possible right now so maybe switching between a console.log/error or flashNotification makes sense. Let's go with it for now.

zefhemel (approved):

PR #48, opened 2022-07-23
Add GHR protocol to plug manager

Willyfrog:
After writing the discussion on https://github.com/silverbulletmd/silverbullet/discussions/47

I figured at least the protocol should be easy to do and here it is. I modified the backlinks release to conform to that one so once this is merged you can install silverbullet-backlinks by adding:
```yaml
- ghr:willyfrog/silverbullet-backlinks
```
or specify a version
```yaml
- ghr:willyfrog/silverbullet-backlinks/v0.2
```

I haven't properly managed what happens on unresolved plugs, as I think it is out of scope. Right now it is throwing an error the same way an unresolved https plug does
This is it in action
CleanShot 2022-07-23 at 21 04 38
.

### DIFF
--- a/packages/plugs/core/core.plug.yaml
+++ b/packages/plugs/core/core.plug.yaml
@@ -280,7 +280,10 @@ functions:
     path: "./plugmanager.ts:getPlugGithub"
     events:
       - get-plug:github
-
+  getPlugGithubRelease:
+    path: "./plugmanager.ts:getPlugGithubRelease"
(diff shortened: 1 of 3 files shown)


zefhemel:
Very nice! Took me a while to figure out what `ghr` stood for though :D

Willyfrog:
Feel free to change the protocol to something else. I wasn’t sure what to use

Issue #49, opened 2022-07-26
Renaming page command deletes page if default prompt input is not changed


m1lt0n:
While trying to rename a page, if the user hits "OK" in the prompt and submits the same value (since the prompt has as its default the current page name), the page is deleted.

Suggested change: check for old vs new value and make the rename a no-op in case they are equal (same behavior as when newName is falsy).

I've prepared a small PR for this https://github.com/silverbulletmd/silverbullet/pull/50

zefhemel:
Closing after merging #50

PR #50, opened 2022-07-26
Fix deletion of page on renaming to current name

m1lt0n:
Relevant issue: https://github.com/silverbulletmd/silverbullet/issues/49


zefhemel:
Nice catch, thanks!

Issue #43, opened 2022-07-23
Documentation: data objects


zefhemel:
Fully undocumented, but implemented exist:

    ```data
    name: John
    age: 50
    ---
    name: Jane
    age: 53
    ```

Indexed and queryable across the entire space, e.g.:

```
<!-- #query data where age > 50 -->

<!-- /query -->
```

sinansonmez:
General question about issues with documentation label: Where these documentation info should be placed? Under [website/feature](https://github.com/silverbulletmd/silverbullet/blob/main/website/🤯%20Features.md) ?

zefhemel:
Good question. I didn't think this through fully yet. In case of functionality coming from plugs (like documented here), I think it would be best to add it to "plug docs" like I started one for query here: https://github.com/silverbulletmd/silverbullet/blob/main/website/%F0%9F%94%8C%20Query.md For others it may indeed be something under a page under `website` indeed, could be `feature` we can always move it later.

davidbeijinho:
Hey just to add that I think data object can be very powerfull if we could have multiple in one page and "attached" to a node so we could have 2 paragraphs , in one page each representing a book, and add the values like we do in front matter

lylebrown:
@davidbeijinho I actually just learned something that will work for my use case, and might help you out too -

when you add `data` to a page, it saves the page the data was placed on as a column. So the following query will filter out `data` to just what's on the current page:
`<!-- #query data where page = "{{page}}" -->`

I'm also considering adding a "type" property to all of my data, so I can have multiple types of data (with different columns/properties), and filter by that type.

davidbeijinho:
I was trying to create an example page called books

---

Chamber of Secrets
```data
type: book
year: 1998
```

Philosopher's Stones
```data
type: book
year: 1997
```

  

|type|year|page |pos|
|----|----|-----|-|
|book|1997|Books|0|



But the query does not return the expected results, maybe because the data objects are not to be used like this @zefhemel

zefhemel:
I'm fully open to changing data objects in whatever seems most useful. It's a feature I thought could be useful, without having a clear use case in mind (and in fact I'm not using it myself at the moment).

That said, this is how it works now:

* There is one global "data" source (`#query data`)
* You (should) be able to have as many `data` code blocks in a page as you like, each one becomes its own entry in the data "table"
* You can have multiple entries in a single data code block by separating them with `---`
* There are no ids for these, so you can only filter based on attributes
[...]

Issue #40, opened 2022-07-23
Documentation: #query


zefhemel:
The query mechanism isn't documented at all and should be:

* What is it
* What's the syntax (we got a proper [grammar definition for it](https://github.com/silverbulletmd/silverbullet/blob/main/packages/plugs/query/query.grammar)
* When does it kick in (how do you refresh the data)
* How do you use templates to render the body (spoiler alert: handlebars with some helpers)
* What sources are available 
* How do you implement your own data source

PR #53, opened 2022-07-29
Write documentation for Query and Data

sinansonmez:
This PR updates the document for `query` and usage of `data objects` with examples

Fixes: https://github.com/silverbulletmd/silverbullet/issues/40
Fixes: https://github.com/silverbulletmd/silverbullet/issues/43

### DIFF
--- a/website/🔌 Query.md
+++ b/website/🔌 Query.md
@@ -23,6 +23,7 @@ The query plug is a built-in plug implementing the `<!-- #query -->` mechanism. 
 ### How to run a query
 After writing the query, there are two options
 * Open the **command palette** and run **Materialized Queries: Update**
-* Use shortcut: for windows **Alt-q** and for mac **Option-q** 
+* Use shortcut: for windows **Alt-q** and for mac **Option-q**
+* Go to another page and come back to the page where query is located
 
 After using one of the options, the data will be displayed.
--- a/website/🔌 Query.md
+++ b/website/🔌 Query.md
@@ -72,4 +72,8 @@ country: Germany
 |Jane|53|Rome |Italy|🔌 Query       |2194|
 <!-- /query -->
  
-
+#### 4.3 Plugs’ data sources
+Certain plugs can also give you special options to query a certain data. Some examples are 
+* [[🔌 Github]] provides `gh-pull` to query PRs for selected repo
+* [[🔌 Mattermost]] provides `mm-saved` to fetch (by default 15) saved posts in Mattermost
+For complete list of data sources, please check plugs’ pages
--- a/website/🔌 Query.md
+++ b/website/🔌 Query.md
@@ -5,10 +5,10 @@ repo: https://github.com/silverbulletmd/silverbullet
 author: Silver Bullet Authors
 ```
 
-### What?
+### 1. What?
 The query plug is a built-in plug implementing the `<!-- #query -->` mechanism. You can use query plug to automatically receive information from your notes.
 
-### Syntax
+### 2. Syntax
 1. start with: `<!-- #query [QUERY GOES HERE] -->`
 2. end with: `<!-- /query -->`
 3. write your query: replace `[QUERY GOES HERE]` with any query you want using options below
@@ -20,10 +20,32 @@ The query plug is a built-in plug implementing the `<!-- #query -->` mechanism. 
    * `render`
 5. If you are a developer or have a technical knowledge to read a code and would like to know more about syntax please check out [query grammar](https://github.com/silverbulletmd/silverbullet/blob/main/packages/plugs/query/query.grammar)
 
-### How to run a query
+### 3. How to run a query?
 After writing the query, there are two options
 * Open the **command palette** and run **Materialized Queries: Update**
 * Use shortcut: for windows **Alt-q** and for mac **Option-q**
 * Go to another page and come back to the page where query is located
 
 After using one of the options, the data will be displayed.
+
+### 4. Data sources
+Available data sources can be categorized as
+1. Builtin data sources
+2. Data that can be inserted by users
+3. Plug’s data sourcer
+
+Best part about data sources: there is an auto completion 🎉. Start writing `<!— #query `, it will show you all available data sources.  
+
+#### 4.1 Builtin data sources
+* `page`: list of all pages
+* `task`: list of all tasks created with `[]` syntax
+* `full-text`: use it with `where phrase = "SOME_TEXT"`. List of all pages where `SOME_TEXT` is mentioned
+* `item`: list of ordered and unordered items such as bulleted lists
+* `tags`: list of all hashtags used in all pages
+* `link`: list of all pages giving a link to the page where query is written
+
+<!-- #query  -->
+
+<!-- /query -->
+ 
+
--- a/website/🔌 Query.md
+++ b/website/🔌 Query.md
@@ -6,15 +6,15 @@ author: Silver Bullet Authors
 ```
 
 ### 1. What?
-The query plug is a built-in plug implementing the `<!-- #query -->` mechanism. You can use query plug to automatically receive information from your notes.
+The query plug is a built-in plug implementing the `<!-- #query -->` mechanism. You can use query plug to automatically receive information from your pages.
 
 ### 2. Syntax
 1. _start with_: `<!-- #query [QUERY GOES HERE] -->`
 2. _end with_: `<!-- /query -->`
 3. _write your query_: replace `[QUERY GOES HERE]` with any query you want using options below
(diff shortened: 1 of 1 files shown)


sinansonmez:
Parts still to be added

1. How do you use templates to render the body (spoiler alert: handlebars with some helpers)
2. What sources are available
3. How do you implement your own data source

@zefhemel One question related to number 2. I was able to identify couple sources: 
- page
- data
- data coming from other plugs such as `gh-pull` from Github or `mm-saved` from Mattermost

What are the other sources (if any)?

Willyfrog (approved):

Willyfrog:
this is looking great, thanks!

zefhemel:
This is looking good already! Regarding 2: this depends on the enabled plugs. However, you should now get a full list when auto completing: `<!— #query ` should give you a drop down with all sources. Here’s the ones in the builtins:

* page
* task
* data
* full-text (using a `where phrase = ` clause to filter
* item (indexes all ordered and unordered items such as bulleted lists)
* tag (hash tags)

zefhemel:
One note on when materialized queries are updated: it also happens when you open a page. So another way to trigger a query update is to switch to another page, and switch back, for instance.

sinansonmez:
Thanks @zefhemel and @Willyfrog 

I updated the PR to include missing points and integrate @zefhemel's suggestions related to (i) data options and (ii) materializing queries. 

It is ready to review. Please let me know if I missed or misstated something

website/🔌 Query.md
```diff
@@ -5,4 +5,161 @@ repo: https://github.com/silverbulletmd/silverbullet
 author: Silver Bullet Authors
 ```
 
-The query plug is a built-in plug implementing the `<!-- #query -->` mechanism. 
\ No newline at end of file
+### 1. What?
+The query plug is a built-in plug implementing the `<!-- #query -->` mechanism. You can use query plug to automatically receive information from your notes.
```
zefhemel, line 9: Minor note (hah) on terminology, the term I try to consistently use is “pages” as opposed to “notes.”

website/🔌 Query.md
```diff
@@ -5,4 +5,161 @@ repo: https://github.com/silverbulletmd/silverbullet
...
+
+### 2. Syntax
+1. _start with_: `<!-- #query [QUERY GOES HERE] -->`
+2. _end with_: `<!-- /query -->`
+3. _write your query_: replace `[QUERY GOES HERE]` with any query you want using options below
+4. _available query options_: Usage of options is similar to general query language except special render option. Render option is to use display the data in a format that you created in a separate template
```
zefhemel, line 15: When refering to “general query language” I think you are refering to SQL, yes? Indeed, this is heavily (and largely compatible) with SQL. Except for `render` indeed, and some of the operators.

website/🔌 Query.md
```diff
@@ -5,4 +5,161 @@ repo: https://github.com/silverbulletmd/silverbullet
...
+   * `where`
+   * `order`
```
zefhemel, line 17: `order by`

website/🔌 Query.md
```diff
@@ -5,4 +5,161 @@ repo: https://github.com/silverbulletmd/silverbullet
...
+
+P.S.: If you are a developer or have a technical knowledge to read a code and would like to know more about syntax, please check out [query grammar](https://github.com/silverbulletmd/silverbullet/blob/main/packages/plugs/query/query.grammar).
+
+### 3. How to run a query?
+After writing the query, there are three options:
+* Open the **command palette** and run **Materialized Queries: Update**
+* Use shortcut: hit **Alt-q** (Windows, Linux) or **Option-q** (Mac)
+* Go to another page and come back to the page where query is located
+
+After using one of the options, the data will be displayed.
```
zefhemel, line 30: More precisely: the “body” of the query (between the two placeholders # query and / query) is replaced with the new results of the query.

website/🔌 Query.md
```diff
@@ -5,4 +5,161 @@ repo: https://github.com/silverbulletmd/silverbullet
...
+
+### 4. Data sources
+Available data sources can be categorized as:
```
zefhemel, line 33: You can categorize it this way, but really there’s only one category: plug data sources. Some of these plugs happen to be “builtin” (although that just means they’re distributed with SB by default), but they’re not special in anyway.

website/🔌 Query.md
```diff
@@ -5,4 +5,161 @@ repo: https://github.com/silverbulletmd/silverbullet
...
+1. Builtin data sources
+2. Data that can be inserted by users
```
zefhemel, line 35: You mean the `data` YAML blocks? This is really not any different than any of the builtin data sources.

website/🔌 Query.md
```diff
@@ -5,4 +5,161 @@ repo: https://github.com/silverbulletmd/silverbullet
...
+3. Plug’s data sources
+
+Best part about data sources: there is an auto completion. 🎉 
+
+Start writing `<!— #query `, it will show you all available data sources. 🤯
```
zefhemel, line 40: Or what I usually prefer is to use the `/query` slash command to insert a query template.

website/🔌 Query.md
```diff
@@ -5,4 +5,161 @@ repo: https://github.com/silverbulletmd/silverbullet
...
+|John|50|Milan|Italy|🔌 Query       |2148|
+|Jane|53|Rome |Italy|🔌 Query       |2194|
+<!-- /query -->
+ 
+#### 4.3 Plugs’ data sources
+Certain plugs can also provide special data sources to query a certain data. Some examples are 
+* [[🔌 Github]] provides `gh-pull` to query PRs for selected repo
+* [[🔌 Mattermost]] provides `mm-saved` to fetch (by default 15) saved posts in Mattermost
+
+For complete list of data sources, please check plugs’ pages.
```
zefhemel, line 83: It would be nice (but we can do this later) is to add two more sections:

* One on the operators to be used in `where` clauses, from the top of my head we support `=` (equals), `!=` (does not equal), `<` (less than), `<=` (less or greater than), `>` (greater than), `>=` (greater than or equals), and `=~` (to match against a regular expression) and `!=~` (does not match this regular expression). Further you can combine multiple of these with `and`. Example `prop =~ /something/ and prop != “something”`.
* The template used with a `render` clause, which is instantiated for every matching query result and uses [Handlebars](https://handlebarsjs.com/) syntax.

zefhemel (commented):
I left you a few suggestions. Honestly, this is already 100% better than what we had before (= nothing), so it’s up to you if you like to address them.

website/🔌 Query.md
sinansonmez: Good to know. Thanks it is updated

website/🔌 Query.md
sinansonmez: Makes sense 👨‍💻 updated

website/🔌 Query.md
sinansonmez: 'general query language' is replaced with 'SQL'

website/🔌 Query.md
sinansonmez: Indeed, all time I was using the slash command as well. I forgot to include it 😵‍💫

website/🔌 Query.md
sinansonmez: Okay. I merged number 1 and number 2 together. I kept number 3 as a separate category to provide a bit more info on that.

website/🔌 Query.md
sinansonmez: Yeah, I was referring to that. As mentioned in the previous comment, I merged them together.

website/🔌 Query.md
sinansonmez: I created [an issue](https://github.com/silverbulletmd/silverbullet/issues/55) for this to tackle in a separate PR. Please assign the issue to me. 

For the template, I used in one example below but we can provide more information on 
- how to create a template
- what is the syntax (provide a brief explanation and give a link to [Handlebars](https://handlebarsjs.com)
- how to use the template (already covered by the example I mentioned)  

If there is anything else lets discuss on the issue.

https://github.com/silverbulletmd/silverbullet/blob/9cb8be6d0309a2c8335cf149f01c5aae526c435f/website/%F0%9F%94%8C%20Query.md?plain=1#L148

zefhemel (approved):

zefhemel:
Awesome, I’ll merge this then. Thanks!

sinansonmez:
Thanks @zefhemel it was great to work with you

zefhemel:
And it’s already live: https://silverbullet.md/🔌_Query

PR #54, opened 2022-07-29
Improve error on formating plug sections

Willyfrog:
So I inadvertently added a space, which is valid YAML but not the one expected.
CleanShot 2022-07-29 at 18 24 57@2x

this should give a more meaningful message, I still think we need a better way to notify users, but for now this works.

### DIFF
--- a/packages/plugs/core/plugmanager.ts
+++ b/packages/plugs/core/plugmanager.ts
@@ -38,7 +38,12 @@ export async function updatePlugsCommand() {
 export async function updatePlugs() {
   let plugList: string[] = [];
   try {
-    plugList = await readYamlPage("PLUGS");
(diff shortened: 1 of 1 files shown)


Willyfrog:
Also, I'm not sure why TS is not complaning about casting from `any` to `string[]`

zefhemel:
Alright yeah, validation is not done very well right now, this makes it better.

zefhemel:
@Willyfrog regarding why TS is not complaining about casting from `any` to `string[]` I think this is by design. `any` could be anything, explicitly so, so casting is at your own risk. What would be nice is to get some sort of runtime warning here, but of course TS doesn’t do any runtime checking.

Willyfrog:
yeah, but I guess I was expecting something to be explicitly casted, not with just an assignation.

Issue #11, opened 2022-07-14
Image embedding


zefhemel:
As a first step toward attachment support, it would be good to investigate using [CodeMirriror’s decorations](https://codemirror.net/examples/decoration/) to show images inline into the code editor using the standard `![alt text](https://link.to/image.jpg)` syntax.

As a principle in SB I’d really like to avoid having people to turn on preview to see the full content of a page, so inline embedding of images would be great. But we have to see if will work reasonably.

PR #66, opened 2022-08-23
Show images inline into the editor

m1lt0n:
This PR adds support for displaying images inline in the editor, resolving issue: https://github.com/silverbulletmd/silverbullet/issues/11

### DIFF
--- a/packages/web/inline_image.ts
+++ b/packages/web/inline_image.ts
@@ -1,0 +1,70 @@
+import { syntaxTree } from "@codemirror/language";
+import { Range } from "@codemirror/state";
+import { Decoration, DecorationSet, EditorView, ViewPlugin, ViewUpdate, WidgetType } from "@codemirror/view";
+
+class InlineImageWidget extends WidgetType {
+  constructor(readonly url: string) {
+    super();
+  }
+
+  eq(other: InlineImageWidget) {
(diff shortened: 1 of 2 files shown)


packages/web/inline_image.ts
```diff
@@ -0,0 +1,70 @@
...
+    img.src = this.url;
+    img.style.display = 'block';
+
+    return img;
+  }
+}
+
+const inlineImages = (view: EditorView) => {
+  let widgets: Range<Decoration>[] = [];
+  const imageRegex = /!\[[^\]]*\]\((?<url>.+)\)/;
```
Willyfrog, line 25: from the look of it, it doesn't support [adding a title to an image](https://www.markdownguide.org/basic-syntax/#images-1). Is that on purpose?

m1lt0n: Thanks for that! I've added the title as an image title and alt attribute, too.

m1lt0n:
> Oh this is exciting, but cannot check it now. Screenshot please!

Added one in the PR description :)

zefhemel:
Excellent and relevant example image. This behaves reasonably as well for very large images etc?

zefhemel:
Just managed to test this. Works well. Merging!

PR #69, opened 2022-08-26
feat: dark theme

### LINKED ISSUE #2: Dark mode
A dark theme would be a great addition to the default (light) theme. Many ways to support it ([my go-to resource](https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/)) but in general use the `prefers-color-scheme` set at the OS/browser level or do more custom persistence via `localStorage`, cookies, or maybe the SQLite store.

Pinjasaur:
This probably closes #2.

Spent a bit more time tweaking the styling and then decided to go forward with persisting it via `localStorage` and a `data-theme` on the root `<html>` element. I was particularly thinking if using the app on a mobile device or via a PWA install then toggling the browser's `prefers-color-scheme` is probably less ideal.

I did end up fighting Parcel's service worker cache version key where it kept serving a cached copy of the index.html on page reloads. Restarting the watch script seemed to fix that.

I think this a good first-pass effort:
- the SCSS could probably be refactored into common variables, etc
- since the majority of SB presents as a fancy text editor, I'm curious if it's trivial or not to swap out the theme for just _that_ e.g. I'm a personal fan of Atom's One Dark theme—wondering if a port exists or if there would be too much CSS fighting

Light:
image

Dark:
image

### DIFF
--- a/packages/web/components/top_bar.tsx
+++ b/packages/web/components/top_bar.tsx
@@ -1,5 +1,6 @@
-import { faRunning, faHome } from "@fortawesome/free-solid-svg-icons";
+import { faRunning, faHome, faSun, faMoon } from "@fortawesome/free-solid-svg-icons";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { useEffect, useState } from "react";
 import { Notification } from "../types";
 
 function prettyName(s: string | undefined): string {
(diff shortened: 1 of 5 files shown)


mfossen (approved):

zefhemel:
Can't wait to try this out. Will have a look at it soon!

zefhemel:
Just tested this a bit on iPad and I’m sure we can make further improvements this is already pretty neat! Merging!

Issue #38, opened 2022-07-19
Serendipity plug: random notes


zefhemel:
Sometimes we just need randomness. Commands that may be useful:

* `Random page`
* `Random page with tag` (either using existing tags in the current page, or by asking for a tag)
* `Random page based on search`

zefhemel:
Implemented by @m1lt0n https://github.com/m1lt0n/silverbullet-serendipity

PR #73, opened 2022-09-09
[DRAFT] Serendipity built-in plug

m1lt0n:
Serendipity plug, implemented as a built-in plug (in the context of https://github.com/silverbulletmd/silverbullet/issues/38 ).

### DIFF
--- a/packages/plugs/serendipity/serendipity.ts
+++ b/packages/plugs/serendipity/serendipity.ts
@@ -1,0 +1,49 @@
+import { listPages } from "@silverbulletmd/plugos-silverbullet-syscall/space";
+import { navigate, prompt } from "@silverbulletmd/plugos-silverbullet-syscall/editor";
+import { dispatch } from "@plugos/plugos-syscall/event";
+import { fullTextSearch } from "@plugos/plugos-syscall/fulltext";
+
+export async function randomPage() {
+    const pages = await listPages();
+    const pageNames = pages.map(p => p.name);
+  
+    await navigate(
+      pageNames[Math.floor(Math.random() * pageNames.length)]);
+  }
+  
+  export async function randomPageWithTag() {
+    const tagName = await prompt('Random page should include the following tag:');
+  
+    if (!tagName) {
+      return;
+    }
+  
+    let pages: Array<Array<any>> = await dispatch(
+      'query:page',
+      { query:  { table: 'page', filter: [{ op: '=', prop: 'tags', value: tagName }] }},
+      10 * 1000
+    );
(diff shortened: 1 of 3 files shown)


packages/plugs/serendipity/serendipity.ts
```diff
@@ -0,0 +1,51 @@
+import { listPages } from "@silverbulletmd/plugos-silverbullet-syscall/space";
+import { navigate, prompt } from "@silverbulletmd/plugos-silverbullet-syscall/editor";
+import { dispatch } from "@plugos/plugos-syscall/event";
+import { fullTextSearch } from "@plugos/plugos-syscall/fulltext";
+
```
m1lt0n, line 5: This is a first stab at writing a plug, so many things may be written how they should be :)

packages/plugs/serendipity/serendipity.ts
```diff
@@ -0,0 +1,51 @@
...
+    );
+  
+    if (pages.length === 0) {
+      return;
+    }
+  
+    await navigate(pages[0].map(p => p.name)[Math.floor(Math.random() * pages.length)])
+  }
+
+  export async function randomPageFromSearch() {
```
m1lt0n, line 34: All other functions seem to work (`randomPage` and `randomPageWithTag`), but for `randomPageFromSearch`, I'm getting an error that the fulltext.search syscall is not registered (and verified it by logging in console all syscalls). I'm also getting the same error if I use the `queryProvider` from searcht.ts (instead of the `fullTextSearch` function)

Couldn't move past that, though, as it's weird why the syscall is not registered, since the exact same syscall is made in search.ts queryProvider. I'm probably missing something in the setup of the plug?

zefhemel: Right, I just checked and it looks like fulltext search is currently only registered on the server side. It's not very hard to expose this on the client as well, though. Created #74 for this

zefhemel:
By the way, the question is if you want to really merge this into SB itself or just keep it as a separate repo. I like this functionality, and we can add it to core, but it’s probably also quite niche so not sure if it should be included by default?

packages/plugs/serendipity/serendipity.ts
zefhemel: This should now work (latest commits in "main")

m1lt0n:
Yep, agree with moving it into a standalone plug given it's nice indeed. I've created https://github.com/m1lt0n/silverbullet-serendipity and working on it. I'll close this PR

PR #80, opened 2022-09-24
Add shortcut hint for command switcher

hanzei:
Add a hint about the keybord shortcut to the command switcher icon. Otherwise there is no way to discover it in the product.

image


zefhemel:
Nice, thank you!

PR #82, opened 2022-09-29
Document in query operator

hanzei:


zefhemel:
Let me check if the `in` operator is actually implemented 😃

hanzei:
At least the query grammar [mentions it](https://github.com/hanzei/silverbullet/blob/94e9aef6f496cdeefc4d0e098cfc5f7fd7e777ba/packages/plugs/query/query.grammar#L36) and in a quick test it did work.

zefhemel:
Yep, it's there. Thanks!

PR #109, opened 2022-11-03
Add nextWeek template variable

hanzei:


zefhemel:
Nice, we continue our arbitrary ad hoc template variable streak 🤣

Issue #16, opened 2022-07-15
Add proper authentication


zefhemel:
Currently Silver Bullet either runs unauthenticated (anybody access to your machine has full access) or in a super basic "password" mode when adding the `--password mypassword` flag.

This isn't great, we probably want to have proper authentication. As suggested by @Willyfrog [here](https://github.com/silverbulletmd/silverbullet/discussions/14#discussioncomment-3150591), it could make sense to use a library like [PassportJS](https://www.passportjs.org/) to add a proper authentication mechanism.

One thing to consider is: do we have a multi-user use case and do we want to support that (if so, quickly we'd also get into the whole authorization, permission and real-time collab topic)? And if not, how complicated an authentication system do we want for effectively just one account? Perhaps a simple username/password combo with temporary session tokens is sufficient?

zefhemel:
Btw, this is one of those tickets where I see it as important (security in general), but I have very personal little interest in implementing this myself. So if there's any takers, feel very free :)

tionis:
A OpenID connect login would be awesome to integrate the wiki with other services  
I currently don't have time and energy to implement the whole authentication.  
So, if someone would implement the groundwork for authentication using password.js, I could easily add OpenID connect support using https://www.passportjs.org/packages/passport-openidconnect/

tionis:
I've also got an idea regarding permission management:
In the simple but expandable spirit of the project, each user could belong to multiple groups.  
Each group has a JavaScript function taking the path name (and maybe some other data) as input and returning the permissions of the group.  
Each user gets the union of all group permissions. (Which means if they are in a group that gives them access to a page and one that does not, they will have access)
An ACL manager plug could later provide a UI that generates these functions automatically from regex matches etc.

zefhemel:
Until now SB has been single user, recently with the exception of [Collaboration on individual pages](https://silverbullet.md/%F0%9F%94%8C_Collab) should it remain that way or not? Adding authentication makes sense even for the single user use case, adding a permission system may not?

If SB should support teams, a permission system would make sense, but that's just the tip of the ice berg. What else should change? Likely all pages should be collaboratively editable, for instance. Version control support likely would need to be put in place. Any thoughts on that?

tionis:
Well, the optimal knowledge base setup I've been chasing and trying to build for quite some time is each person having their own wiki and being able to synchronize them with each other using an intermediate server.
[...]

PR #138, opened 2022-12-03
Allow the user to specify hostname to listen on

jkseppan:
I suggested this in https://github.com/silverbulletmd/silverbullet/issues/16#issuecomment-1336213572

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,8 +80,9 @@ terminal:
 silverbullet <pages-path>
 ```
 
-By default, SB will bind to port `3000`, to use a different port use the
-`--port` flag. By default SB doesn’t offer any sort of authentication, to add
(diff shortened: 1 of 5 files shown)


zefhemel:
Looks good! My only question is if hostname the right term. Maybe bind IP or simply an `--ip` flag instead?

zephraph:
Hostname matches what oak expects to be passed.

https://github.com/silverbulletmd/silverbullet/blob/0c966e5015cd5539735df1ec502af6f02b8a0430/server/http_server.ts#L131

I'd also say `--hostname` should be understood to be includes of explicit IPs whereas `--ip` would likely not. Probably isn't really a big deal either way so long as it's documented.

zefhemel:
Good point. Alright. Merging this, thanks a lot!

Issue #150, opened 2022-12-04
Update docker to Deno setup


zefhemel:
Currently still using the old npx deployment.

brtwrst:
I have 2 working versions based on the old one
A few things to note:
* the deno user has different IDs in the alpine and default (debian) image of deno (users might have to change the permissions of their /space folder)
* The folder /deno-dir/deps is not writable for the deno user in either images and has to be chown'd or the root user has to be used.
* I like to add [tini](https://github.com/krallin/tini) to containers so that shutting them down does not take several seconds

Dockerfile.debian
```Dockerfile
FROM denoland/deno:debian

RUN apt-get update; apt-get install tini
RUN deno install -f --name silverbullet -A --unstable https://get.silverbullet.md

RUN mkdir /space
WORKDIR /space
RUN chown deno:deno /space

# /deno-dir/deps needs write permissions for deno user or sb will not run
RUN chown deno:deno /deno-dir/deps

# deno user id is 1993 in debian image
USER deno

EXPOSE 3000

ENTRYPOINT ["/usr/bin/tini", "-g", "--"]
CMD ["silverbullet", "/space"]
```

Dockerfile.alpine
```Dockerfile
FROM denoland/deno:alpine

RUN apk add tini
RUN deno install -f --name silverbullet -A --unstable https://get.silverbullet.md

RUN mkdir /space
WORKDIR /space
RUN chown deno:deno /space

# /deno-dir/deps needs write permissions for deno user or sb will not run
RUN chown deno:deno /deno-dir/deps

# deno user id is 1000 in alpine image
USER deno

EXPOSE 3000

ENTRYPOINT ["/sbin/tini", "-g", "--"]
CMD ["silverbullet", "/space"]
```

MaverickDe:
Hi, Can I get assigned to this issue?

MaverickDe:
Please can you explain in depth what is to be achieved in this issue

zefhemel:
What we would like to have (from my perspective) is an updated `Dockerfile` (likely based on recent Deno-based, minimal) image, downloads the most recent version of SB into the image, allows volume mounting of the space folder into the image, and runs SB on boot. 

I think @brtwrst 's second solution (Alpine) is what I was expecting.

brtwrst:
Yes the second one does that and it's the one I'm using currently. Also has the upside of keeping the user ID the same compared to the previous (non deno based) image. 
[...]

PR #174, opened 2022-12-07
Deno docker update

zefhemel:
As discussed in #150 using the work done by @brtwrst @MaverickDe and @BaptisteBdn

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,26 @@
-FROM node:18.6.0-slim
+FROM denoland/deno:alpine-1.28.3
 
-RUN mkdir /space
-RUN chown node:node /space
-USER node 
-WORKDIR /space
+# The volume that will keep the space data
+# Create a volume first:
+#   docker volume create myspace
+# Then bind-mount it when running the container with the -v flag, e.g.:
+#   docker run -v myspace:/space -it zefhemel/silverbullet
+VOLUME /space
 
-RUN npx --yes @silverbulletmd/server || true 
+# Copy the bundled version of silverbullet into the container
+ADD ./dist/silverbullet.js /silverbullet.js
 
+# Make sure the deno user has access to the space volume
+RUN mkdir -p /space
+RUN chown -R deno:deno /space
+
+# deno user id is 1000 in alpine image
+USER deno
+
+# Expose port 3000
+# Port map this when running, e.g. with -p 3002:3000 (where 3002 is the host port)
 EXPOSE 3000
 
-CMD ["sh","-c","npx --yes @silverbulletmd/server --port 3000 /space"]
+# Run the server, allowing to pass in additional argument at run time, e.g.
+#   docker run -p 3002:3000 -v myspace:/space -it zefhemel/silverbullet --user me:letmein
+ENTRYPOINT ["deno", "run", "-A", "--unstable", "/silverbullet.js", "/space"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,4 +23,4 @@ EXPOSE 3000
 
 # Run the server, allowing to pass in additional argument at run time, e.g.
 #   docker run -p 3002:3000 -v myspace:/space -it zefhemel/silverbullet --user me:letmein
-ENTRYPOINT ["deno", "run", "-A", "--unstable", "/silverbullet.js", "/space"]
+ENTRYPOINT ["/tini", "--", "deno", "run", "-A", "--unstable", "/silverbullet.js", "/space"]
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,7 +5,7 @@ on:
     branches: ["main", "deno-docker"]
 
 jobs:
-  build-test-publish:
+  docker-build-push:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -1,12 +1,26 @@
...
+# The volume that will keep the space data
+# Create a volume first:
+#   docker volume create myspace
+# Then bind-mount it when running the container with the -v flag, e.g.:
+#   docker run -v myspace:/space -it zefhemel/silverbullet
+VOLUME /space
 
-RUN npx --yes @silverbulletmd/server || true 
+# Copy the bundled version of silverbullet into the container
+ADD ./dist/silverbullet.js /silverbullet.js
```
zefhemel, line 11: This relies on the `deno task bundle` to have run before.

BaptisteBdn: Where does `deno task bundle` run ? Inside CI ?
That would mean that this Dockerfile cannot be used as it is to build the image from source code

brtwrst:
While it is true that "tini is included in newer docker versions" it needs to be manually enabled by using `docker run --init [...]`.
Until "recently" this was not available in docker-compse and if I read the [docs](https://docs.docker.com/compose/compose-file/#init) right, you ahve to add init: true to the docker-compse file to make it work.

zefhemel:
Ok, now also added the @brtwrst added the `init` thing.

Dockerfile
BaptisteBdn: If this Dockerfile is not used to contribute and develop, indeed it does not really matter.
Maybe the same size can be achieved if we delete everything and just keep `silverbullet.js`.

zefhemel: Sadly deleting doesn't help (due to docker layers), not a big deal but hey... 

I've now also added the docker image build and publish to the PR. So on every merge to master a new docker image is built and pushed. Eventually we may want to do this only for tags, but we're on a streaming release for now ;)

brtwrst:
I tested it locally and when i specify init: true i get a warning: `[WARN tini (7)] Tini is not running as PID 1 and isn't registered as a child subreaper.`

Some research into the deno alpine image revealed that they include tiny already and do also provide an ENTRYPOINT
image

This means that init:true doubles up the signal handling and tini is not started as PID 1
So with this image, because it already includes an installation of tini and an ENTRYPOINT invoking it, the `init:true` is actually not needed or desired.

Sorry for the confusion, I added tini into the image when i was playing with the old npx version and never took it out when moving to the deno based image.

zefhemel:
Aaaaand removed `init: true` again 😆

BaptisteBdn (approved):

Dockerfile
```diff
@@ -1,12 +1,26 @@
...
+USER deno
+
+# Expose port 3000
+# Port map this when running, e.g. with -p 3002:3000 (where 3002 is the host port)
 EXPOSE 3000
 
-CMD ["sh","-c","npx --yes @silverbulletmd/server --port 3000 /space"]
+# Run the server, allowing to pass in additional argument at run time, e.g.
+#   docker run -p 3002:3000 -v myspace:/space -it zefhemel/silverbullet --user me:letmein
+ENTRYPOINT ["deno", "run", "-A", "--unstable", "/silverbullet.js", "/space"]
```
brtwrst, line 26: Any reason why this is ENTRYPOINT and not CMD?
by overriding the ENTRYPOINT that was set by the base image, tini is not called as PID 1 any more.
image

zefhemel: ok let me know if that works, and I'll re-add the init thing again

brtwrst: I am not able to build the image with the provided dockerfile. 
I don't have deno locally and I can't make the `deno task bundle` command work in a container to build it with a multi stage build in docker.

I played around a bit with my Dockerfile and found a few things 
* With the current entrypoint and no --init, signals do not get forwarded to the deno process and if the container is shut down, deno is killed -9 by the container after a few seconds of waiting 
* using `docker run --init` works
  * feels clunky and most of the time people are probably better off using docker compose to keep track of the options/parameters and to make it more portable and easier to back up
  * using `init: true` in docker-compose works 
  * the current docker-compose seems weird, the goal of a docker-compose file provided to "end users" should be an easy way to start it. The current docker-compose tries to build the project and then fails if the project was not "deno bundled" before.
  * I think the docker-compose should reference the "official" silverbullet image on dockerhub and just pull that and start it up
* setting the entrypoint to ` ["/tini", "--", stuff, to, make, it, run]` works
  * Signals get handles correctly by the executable `/tini` that is already provided by the base image and no `--init` or similar is needed
  * `docker run [...] --user='test:test'` also works

brtwrst: Looks good to me but although I can't test it locally until it is pullable from dockerhub

zefhemel: It's pullable now (auto builds and pushed from this branch)

brtwrst: Yes works for me.

I have my own compose file which includes this:
```yml
services:
  silverbullet:
    image: zefhemel/silverbullet
    container_name: silverbullet
    restart: unless-stopped
    volumes:
      - silverbullet:/space
    command: --user='user:pw'
```
So to use the password option with docker-compose the `command:` is how it is done - it might make sense to add this to the "example" compose file (maybe commented out)

Issue #127, opened 2022-11-24
Improve dark mode


zefhemel:
Recently (likely because I'm not a heavy user) dark mode has been regressing a bit in terms of looks. It would be great if somebody with interest in this (and some CSS skills) could improve things and make it look better. Various things can use improvement, but especially the code completion panel looks pretty bad right now.

simcard0000:
@zefhemel Hey Zef! Apart from the code completion panel, I'm curious to know what other aspects of the UI (dark mode or otherwise) need improvement 🙂

zefhemel:
I'll go through some of the ui and take some screenshots with pointers. Some may just come down to taste, but some are just objectively bad 🤡

I imagine if you operate in dark mode for some time you'll see stuff as well.

simcard0000:
@zefhemel That would be helpful! Yes, I noticed some things in dark mode can be hard to read (like when you're typing a page name to open in the "open page/navigate to page" panel). I was looking through the SCSS files and I was thinking it might be helpful to define palettes for the light and dark themes using SCSS `maps` and then using values from those maps across the rest of the SCSS files.

zefhemel:
In the CSS area I'm 100% open to any suggestions for improvement. This is definitely not my area of expertise.

zefhemel:
![CleanShot 2022-12-08 at 15 54 27](https://user-images.githubusercontent.com/59073/206478796-2ea351c8-d83c-4c4b-a8b5-746140853c95.png)

So if I knew how to improve this I'd do it, but I'm pointing out some things that don't feel great, per arrow

1. Contrast
2. Button color
3. Wiki link color
4. Header color

zefhemel:
<img width="380" alt="CleanShot 2022-12-08 at 15 57 22" src="https://user-images.githubusercontent.com/59073/206479216-bcbe13ec-6562-413c-bec8-163ee00c6198.png">

Code complete is pretty bad (also notice the bracket color)

zefhemel:
That would be awesome!

simcard0000:
Nice; could I get assigned to the issue btw? 🙂

simcard0000:
@zefhemel Created a draft PR at: https://github.com/silverbulletmd/silverbullet/pull/179 ✨

zefhemel:
Another thing is to increase the contrast between the background color and main text: simply white text rather than gray?

simcard0000:
Will do; I think that'll be easier on the eyes as well.

[...]

Issue #10, opened 2022-07-14
Theming support


zefhemel:
Currently all styles are hard-coded in [various SCSS](https://github.com/silverbulletmd/silverbullet/tree/main/packages/web/styles) files, we probably will want a theming system.

Plenty of design decisions to be made here:

* Should themes be defined in plain CSS or e.g. JSON? We’d want to be able to hot load and unload these and likely namespace them somehow.
* Should themes be plugs, part of plug YAML definitions maybe even, or be something completely different?
* [CodeMirror has its own theming mechanism](https://codemirror.net/examples/styling/) should we leverage this somehow?
* Custom syntax in plug YAML can define styling as well, such as colors. How would this interact with themes?

zefhemel:
Very related indeed

zefhemel:
See #58 for a discussion

kjk:
Roam Research has a nice and easy to use (from user point of view) theming system: you can create a special page `roam/css` page and put the css as a code block (or `@import` a `.css` file from url) and that CSS will-override the built-in CSS.

This is easy to use for users and easy for developers because updating the CSS in that page takes immediate effect in the app so you see changes quickly. It's a live update system.

They also have a bunch of themes created by people (and they are nice).

So one option would be to adopt Roam's CSS classes (to get instant access to their themes) and implement similar system of a special page.

See e.g.:
* https://nesslabs.com/the-best-css-themes-for-roam-research
* https://roamthemes.io/
* https://www.goedel.io/p/a-craft-like-theme-for-roam-research
* https://www.keepproductive.com/blog/best-roam-research-css-themes

zefhemel:
Having the CSS classes match is not likely to happen. I do like the idea of embedding styles into a page though... 🤔

rktic:
I have yet to find a theming solution based on utilizing CSS files, that's bearable to use for the average human. Including most of the developers I know, who all passionately hate CSS. :) Even though I love CSS.

That said, I _firmly believe_ there's an accessible, sustainable and scalable approach. Leveraging SCSS variables and a cleverly written CSS framework. For this specific use-case.

Based on having defined the **purpose** of **theming** in **Silver Bullet**. 

Mind you: Nobody ever changes a colour for the sake of changing a colour ;)

firasdib:
I would approach this through css variables. Try to create a minimum set of variables that handle different aspects of the website, and then use that throughout.

[...]

PR #179, opened 2022-12-09
WIP: Theming changes

simcard0000:
Right now, this is a very rough and incomplete draft, wherein the work will be split into two parts:

- [ ] refactoring SCSS code relating to theming
- [ ] overhauling the built-in dark theme and making some improvements to the light theme

Feedback is appreciated! Also, see [Theming support #10](https://github.com/silverbulletmd/silverbullet/issues/10), [Theming approach #58](https://github.com/silverbulletmd/silverbullet/discussions/58), and [Improve dark mode #127](https://github.com/silverbulletmd/silverbullet/issues/127).


zefhemel:
My current thinking (but I’m open to alternative ideas) is to distribute themes as plugs. If we do this, there’s a good argument to allow theming in pure CSS without having to preprocess them with e.g. SCSS, do you think that’s doable or would be too verbose? We can probably include a SCSS preprocessing step somewhere if necessary.

simcard0000:
@zefhemel It's a neat idea; and fits in with the overall ethos of "plugs all the way down". I'm not 100% sure if it'll be too verbose, but if themes are plugs then it makes it easy to enable/disable custom themes (which I guess is for now adding/removing plugs from the PLUGS page or through plug management). 

I think it's still the right track to have a `theme.scss` file where we can put any properties/parts that can be "safely changed"; and have it wrapped up in a CSS `@layer` rule, which means that any styles declared outside of this will override the styles declared in the layer, regardless of specificity. That way people can worry less about the structure of their CSS in their plugs.

zefhemel:
Alright, let me know when I should have a closer look. I'm no SCSS expert by any means, so I'll largely trust your judgment :)

simcard0000:
@zefhemel To be fair I'm no SCSS expert either 😅 but it's fun figuring out how to design a theming system 🙂

zefhemel:
Closing this due to staleness.

PR #186, opened 2022-12-11
Initial Admonition support (github format)

### LINKED ISSUE #67: Admonition support
It would be cool to support [github's admonitions](https://github.com/community/community/discussions/16925)

christian-schulze:
Resolves #67

Adds initial markdown support for Admonitions (using github syntax):
```
> **note** Note Admonition
> This is a Note Admonition
```
And:
```
> **warning** Warning Admonition
> This is a Warning Admonition
```

### DIFF
--- a/web/cm_plugins/admonition.ts
+++ b/web/cm_plugins/admonition.ts
@@ -11,6 +11,9 @@ import { decoratorStateField, isCursorInRange } from "./util.ts";
 
 type AdmonitionType = "note" | "warning";
 
+const ADMONITION_REGEX = /^>( *)\*{2}(Note|Warning)\*{2}( *)(.*)(?:\n([\s\S]*))?/im;
+const ADMONITION_LINE_SPLIT_REGEX = /\n>/gm;
+
 class AdmonitionIconWidget extends WidgetType {
   constructor(
     readonly pos: number,
@@ -46,6 +49,41 @@ class AdmonitionIconWidget extends WidgetType {
   }
 }
 
+type AdmonitionFields = {
+  preSpaces: string;
+  admonitionType: AdmonitionType;
+  postSpaces: string;
+  admonitionTitle: string;
+  admonitionContent: string;
+};
+
+// Given the raw text of an entire Blockquote, match an Admonition block.
+// If matched, extract relevant fields using regex capture groups and return them
+// as an object.
+//
+// If not matched return null.
+//
+// Example Admonition block (github formatted):
+//
+// > **note** I am an Admonition Title
+// > admonition text
+//
+function extractAdmonitionFields(rawText: string): AdmonitionFields | null {
+  const regexResults = rawText.match(ADMONITION_REGEX);
+
+  if (regexResults) {
+    const preSpaces = regexResults[1] || '';
+    const admonitionType = regexResults[2].toLowerCase() as AdmonitionType;
+    const postSpaces = regexResults[3] || '';
+    const admonitionTitle: string = regexResults[4] || '';
+    const admonitionContent: string = regexResults[5] || '';
+
+    return { preSpaces, admonitionType, postSpaces, admonitionTitle, admonitionContent };
+  }
+
+  return null;
+}
+
 export function admonitionPlugin(editor: Editor) {
   return decoratorStateField((state: EditorState) => {
     const widgets: any[] = [];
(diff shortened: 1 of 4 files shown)


web/cm_plugins/admonition.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export function admonitionPlugin(editor: Editor) {
+  return decoratorStateField((state: EditorState) => {
+    const widgets: any[] = [];
+
+    syntaxTree(state).iterate({
+      enter: (node: SyntaxNodeRef) => {
+        const { type, from, to } = node;
+
+        if (type.name !== "QuoteMark") {
```
zefhemel, line 57: This means: do the following for every single AST node that is NOT a quotemark, is this what you want? If you don’t want to match a quote mark (which is just the `>` symbol, its parent is probably something more useful), there’s probably a parent element (can’t remember from the top of my head and can’t check now, perhaps `Blockquote` or similar?) that is more restrictive? What you do here will likely work but is very expensive.

web/cm_plugins/admonition.ts
```diff
@@ -0,0 +1,127 @@
...
+          // extract raw text from admonition block
+          const text = state.sliceDoc(from, to);
+
+          // split text into type, title and content using regex capture groups
+          const regexResults = text.match(/^>( *)\*{2}(Note|Warning)\*{2}( *)(.*)(?:\n([\s\S]*))?/im);
```
zefhemel, line 62: Style thing: let’s lift this regex to a `const` at the top level of the file.

web/cm_plugins/admonition.ts
```diff
@@ -0,0 +1,127 @@
...
+          if (!regexResults) {
+            return;
+          }
+
+          const preSpaces = regexResults[1] || '';
+          const admonitionType = regexResults[2].toLowerCase() as AdmonitionType;
+          const postSpaces = regexResults[3] || '';
+          const admonitionTitle: string = regexResults[4] || '';
+          const admonitionContent: string = regexResults[5] || '';
```
zefhemel, line 71: Style suggestion: perhaps replace this with a construct like `const [preSpaces, admonintionType, postSpaces, …] = regexResults;` ? See other files in this folder for examples.

web/cm_plugins/admonition.ts
```diff
@@ -0,0 +1,127 @@
...
+              break;
+            default:
+              //
+          }
+
+          widgets.push(
+            Decoration.line({ class: "sb-admonition-title " + classes.join(" ") }).range(fromOffsets[0]),
+          );
+
+          if (!isCursorInRange(state, [iconRange.fromNoSpaces, iconRange.to - 1])) {
```
zefhemel, line 103: Once you make the `type` match (earlier comment) more specific, you probably want to back out of this whole logic if the cursor is in range earlier. That is: lift the `if(isCursorInRange(state, from, to)) { return; }` further to the top because likely you don’t want to show any widgets when your cursor is anywhere in a blockquote, right?

zefhemel (changes requested):
Nice start, have a look at my comments! Haven’t been able to actually test this yet (not behind a computer) but this should already be useful.

web/cm_plugins/admonition.ts
christian-schulze: Yep this is definitely a mistake on my part, my intention was definitely to match, not skip. Oddly enough it seems to be working...I will fix soon.

web/cm_plugins/admonition.ts
christian-schulze: So this plugin is dynamically applying styles to the existing editable area, its only changing colours and a few other minor things. This block is for the icon, if your cursor is within the `**note**` or `**warning**` text it will skip replacing it with the icon

christian-schulze: I have a larger regex further up which bails out if it doesn't match an Admonition formatted block

zefhemel: Alright, I haven’t been able to try how it works in practice, so I was just making assumptions based on the code.

web/cm_plugins/admonition.ts
christian-schulze: Yep have changed to `Blockquote`, works correctly. Will push updates in a few minutes.

zefhemel:
Looks reasonable enough from the code perspective, I’ll check it out (literally) later to see if actually works 😉 before merging (QA you know).

christian-schulze:
> Looks reasonable enough from the code perspective, I’ll check it out (literally) later to see if actually works wink before merging (QA you know). Or is there more you’d want to change (since it’s still marked as WIP).

I think there are many things that can be added, more types, different syntax support, custom types etc. But that can be for future PR's.
If you don't find any bugs am happy for you to merge. Will remove WIP.

If you don't find anything wrong and don't want to then maybe should merge it

zefhemel:
Overall this works ok. The only issue I have is when the live preview kicks in. In its current behavior, this works strangely when adding spaces. I'm exegarating it in this video, but adding a space (even one) after `**Note**` is completely invisible and feels buggy:

https://user-images.githubusercontent.com/59073/206925584-3e52f795-027f-4a56-9d71-f3b494925760.mp4

How I would solve this myself is to simply be more conservative when to actually show the live preview icon. I would simply not do any of that magic if the cursor is _on the first line_ (the line with the admonition). That way you can always see all the markup (including spaces), and when you move to the second (non-header) line, the header renders as a preview. What do you think about that approach?

Alternatively somehow the space thing needs to be handled differently, now it feels a bit broken.

web/styles/editor.scss
```diff
@@ -155,6 +155,46 @@
     min-height: 1em;
   }
 
+  .sb-admonition {
+    border-left-width: 4px !important;
+    border-left-style: solid;
+  }
+
+  .sb-admonition-title {
```
zefhemel, line 163: If it's empty, may as well remove?

web/styles/editor.scss
```diff
@@ -155,6 +155,46 @@
     min-height: 1em;
   }
 
+  .sb-admonition {
```
zefhemel, line 158: It's better to move most of these styles (maybe simply all)? to `theme.scss` this will make it easier to make this themable down the line (not yet there, but preparing for it).

web/cm_plugins/admonition.ts
```diff
@@ -0,0 +1,173 @@
...
+            Decoration.line({ class: "sb-admonition-title " + classes.join(" ") }).range(fromOffsets[0]),
+          );
+
+          // If cursor is not within the **note|warning** part of the first line,
+          // replace it with the correct icon
+          if (!isCursorInRange(state, [iconRange.fromNoSpaces, iconRange.to - 1])) {
+            widgets.push(
+              Decoration.replace({
+                widget: new AdmonitionIconWidget(
+                  iconRange.to,
```
zefhemel, line 151:
```suggestion
 iconRange.from + 1,
```

Seems to work better for me? Puts the cursor in either `Note` or `Warning`

christian-schulze:
@zefhemel have addressed your other feedback.

Regarding the replacement of icon, have changed it to your first suggestion (within first line or not). Its a nifty feature but not essential. Will take another look at that later on.

zefhemel:
Great, this is nice. I'll merge it, we can tweak things later. Thanks!

christian-schulze:
> Great, this is nice. I'll merge it, we can tweak things later. Thanks!

Awesome thanks. You've got an interesting and useful project here, I'm happy to contribute more in future.

PR #190, opened 2022-12-13
Fix "Rename page" command

chriszarate:
Renaming a page works when pressing <kbd>Enter</kbd> after editing the page name, but fails when using the "Rename page" command, either from the command palette for via hotkey (Cmd+Alt+R):

Screen Shot 2022-12-12 at 9 30 19 PM

The `renamePage` function expects a string but is called with a command definition (`cmdDef`). This probably deserves a refactor, but overloading the command definition with a `page` parameter allows it to be called successfully in both contexts.

### DIFF
--- a/plugs/core/core.plug.yaml
+++ b/plugs/core/core.plug.yaml
@@ -68,6 +68,7 @@ functions:
       name: "Page: Rename"
       mac: Cmd-Alt-r
       key: Ctrl-Alt-r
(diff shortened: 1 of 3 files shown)


zefhemel:
Yeah, this is not the cleanest of code, but I confirmed your fix works so I'll merge it. We can clean this later.

Thanks!

PR #192, opened 2022-12-13
added no-results text

### LINKED ISSUE #182: Show "No results" message rather than empty table for 0 result queries
What the title says: right now if you get an empty result set and use the default MD table renderer you get an empty table that doesn't even live preview. It would be better to replace this with a message saying "No results"

ItzNesbroDev:
Fixes #182

### DIFF
--- a/plugs/directive/query_directive.ts
+++ b/plugs/directive/query_directive.ts
@@ -10,7 +10,6 @@ export async function queryDirectiveRenderer(
   pageName: string,
   query: string
 ): Promise<string> {
-  console.log("gg");
   const parsedQuery = parseQuery(replaceTemplateVars(query, pageName));
 
   console.log("Parsed query", parsedQuery);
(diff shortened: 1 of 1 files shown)


plugs/directive/query_directive.ts
```diff
@@ -17,10 +17,10 @@ export async function queryDirectiveRenderer(
   const results = await events.dispatchEvent(
     `query:${parsedQuery.table}`,
     { query: parsedQuery, pageName: pageName },
-    30 * 1000,
+    30 * 1000
   );
   if (results.length === 0) {
-    return "";
+    return "No results";
```
zefhemel, line 23: Have you tried if this actually works? I haven’t but I’m pretty sure it doesn’t. The case of `results.length === 0` happens when you’re querying a query source that doesn’t exist (this should probably return an error message.

Where you’ll want to make changes is a few lines down (line 29, the one with `jsonToMdTable`, if that `results[0]` is an empty array, then you show `No results`.

zefhemel (changes requested):
Please have a look at my comments 🙂

zefhemel:
Looks good, thanks!

PR #197, opened 2022-12-15
Vim mode

### LINKED ISSUE #149: Add vim support
I'd love to use this tool being self-hosted and WYSIWYG and like Obsidian, but it doesn't yet have vim keybinding support. I see that it uses Codemirror, and I believe Obsidian supports its vim keybindings through a Codemirror repo and they don't maintain it directly (there is a large discussion here that I haven't looked at much: https://forum.obsidian.md/t/vim-support-for-the-new-editor-codemirror-6-neovim-libvim-or-something-else/21803/31).

I'd love it if Silverbullet could take advantage of an existing vim keybindings repo like this to support vim (and hopefully .vimrc files, too) in the editor. I know that Obsidian doesn't keep vim in its own plugin because it is too integrated with the core editor - not sure if a Plug would work for this or not either.

zefhemel:
Fixes #149

### DIFF
--- a/common/deps.ts
+++ b/common/deps.ts
@@ -44,7 +44,7 @@ export {
 
 export type { NodeType, SyntaxNode, SyntaxNodeRef, Tree } from "@lezer/common";
 
-export { searchKeymap } from "https://esm.sh/@codemirror/search@6.2.3?external=@codemirror/state,@codemirror/view";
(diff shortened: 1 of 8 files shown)


zefhemel:
There's some issues I have uncovered, I'll create upstream issues for this, e.g. https://github.com/replit/codemirror-vim/issues/68

PR #211, opened 2022-12-18
Minor change in docker compose to BasicAuth

dedioste:
I propose to change the "command" hint in docker compose to a list element.
This way:
- style is aligned to docker file
- users are allowed to just uncomment and insert desired user:pwd values to enable BasicAuth


zefhemel:
Works for me 😊

Issue #107, opened 2022-10-29
Builds don’t work on Windows


zefhemel:
It's generating absolute Windows path for the temporary file, which ESBuild apparently doesn't like, probably need to be translated to something more Unixy:

```
   1 │ import {bla} from "C:\Users\Zef\AppData\Local\Temp\efb0334f\file.ts";export default bla;
```

zefhemel:
Update: apparently building SB in general doesn't work on Windows, also due to `./build_plugs.sh` not being Windows compatible.

pm64:
For reference, here's the current build output on Windows for 8d3688b2fd6308c5f115ebfff4cac40cb3ffe08c:

```
PS D:\temp\silverbullet-main> deno.exe task build
Task build deno run -A --unstable --check build_plugs.ts && deno run -A --unstable --check build.ts
Check file:///D:/temp/silverbullet-main/build_plugs.ts
Building [
  "D:\\temp\\silverbullet-main\\plugs\\collab\\collab.plug.yaml",
  "D:\\temp\\silverbullet-main\\plugs\\core\\core.plug.yaml",
  "D:\\temp\\silverbullet-main\\plugs\\directive\\directive.plug.yaml",
  "D:\\temp\\silverbullet-main\\plugs\\emoji\\emoji.plug.yaml",
  "D:\\temp\\silverbullet-main\\plugs\\markdown\\markdown.plug.yaml",
  "D:\\temp\\silverbullet-main\\plugs\\plugmd\\plugmd.plug.yaml",
  "D:\\temp\\silverbullet-main\\plugs\\share\\share.plug.yaml",
  "D:\\temp\\silverbullet-main\\plugs\\tasks\\tasks.plug.yaml"
]
Caching plug https://get.silverbullet.md/global.plug.json to C:\Users\User\AppData\Local\plugos
Caching plug https://get.silverbullet.md/global.plug.json to C:\Users\User\AppData\Local\plugos
Caching plug https://get.silverbullet.md/global.plug.json to C:\Users\User\AppData\Local\plugos
Caching plug https://get.silverbullet.md/global.plug.json to C:\Users\User\AppData\Local\plugos
Caching plug https://get.silverbullet.md/global.plug.json to C:\Users\User\AppData\Local\plugos
Caching plug https://get.silverbullet.md/global.plug.json to C:\Users\User\AppData\Local\plugos
[...]

PR #229, opened 2022-12-22
deno task build fixes for windows

kjk:
Fix `deno task build` for Windows.

Windows doesn't support `.sh` files so replaced with direct usage in `deno.jsonc`.

Since we're no longer going via shell, the glob patterns are not expanded so modified `plugos-bundle.ts` to assume that if no files are given, we want to build everything and we glob ourselves.

Alternative: still pass glob pattern on cmd-line and glob internally (will be no-op if already globbed).

`build_plugs.sh` is no longer used but I left it in case you're using it from cmd-line.

Makes progress on #107

### DIFF
--- a/build_plugs.ts
+++ b/build_plugs.ts
@@ -1,0 +1,38 @@
+import { path, expandGlobSync, flags } from "./plugos/deps.ts";
+import { bundleRun } from "./plugos/bin/plugos-bundle.ts"
+import {
+  esbuild,
+} from "./plugos/compile.ts";
+
+if (import.meta.main) {
+  const args = flags.parse(Deno.args, {
+    boolean: ["debug", "watch", "reload", "info"],
+    string: ["dist", "importmap"],
+    alias: { w: "watch" },
+  });
+
+  if (!args.dist) {
+    args.dist = path.resolve(path.join("dist_bundle", "_plug"));
+  }
+
+  const manifests : string[] = [];
+  const pattern : string = path.join("plugs", "*", "*.plug.yaml");
+  for (const file of expandGlobSync(pattern)) {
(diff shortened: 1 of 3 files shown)


plugos/bin/plugos-bundle.ts
```diff
@@ -171,7 +171,15 @@ if (import.meta.main) {
     alias: { w: "watch" },
   });
 
-  if (args._.length === 0) {
+  // Windows shell doesn't expand glob patterns so if no files given
+  // assume we want to build everything and glob ourselves
+  const manifests = args._ as string[];
+  if (manifests.length === 0) {
+    for (const file of expandGlobSync("plugs/*/*.plug.yaml")) {
```
zefhemel, line 178: I understand this works, but I think it’s quite ugly to hardcode this in a general-purpose command like plugos-bundle. In that case I’d rather have a `build_plugs.ts` specifically to build built-in plugs in the root dir (next to `build.ts`). This wouldn’t need to take any arguments, but just always build this set of plugs.

zefhemel (changes requested):
See my comment. I understand this works, but it’s pretty hacky. Similar to the `deno task build` I’d then rather have a `build_plugs.ts` script that just builds the built-in plugs.

plugos/bin/plugos-bundle.ts
kjk: created build_plugs.ts

zefhemel:
@kjk this is looking good and seems to work (on Mac at least 🤡 ) ready to merge?

zefhemel (approved):

zefhemel:
I’ll take your silence as a HELL YES 😉

Issue #230, opened 2022-12-22
Command palette: continuous up/down list navigation


kjk:
In most command palette implementations, if you press arrow down / up key, you scroll through the list continuously.

In SB you have to press / depress / press / depress etc.

It's because it's hooked up to `keyUp` event in MiniEditor.

There's no support for hooking into `keyDown` event in MiniEdtor.

Also, it would probably be simpler for `FilterList` to hook global keydown / key up event and process them before they get to CodeMirror.

This is what I did when I implemented command palette for https://www.codeeval.dev/ (press "..." button to see in action).

Implementation is: https://github.com/kjk/codeeval.dev/blob/main/fe/SelectLangDialog.svelte#L87 (in Svelte but the idea would work the same in React).

PR #232, opened 2022-12-23
better up / down handling in FilterList

kjk:
As described in https://github.com/silverbulletmd/silverbullet/issues/230

I tried routing keydown via CodeMIrror / MiniEditor like keyup but couldn't figure it out.

Instead I hooked global keydown.

### DIFF
--- a/web/components/filter.tsx
+++ b/web/components/filter.tsx
@@ -164,9 +164,6 @@ export function FilterList({
       if (newSelected >= nOptions) {
         newSelected = nOptions - 1;
       }
-      if (newSelected === selectedOption) {
-        return true;
-      }
       setSelectionOption(newSelected);
       return true;
     }
--- a/web/components/filter.tsx
+++ b/web/components/filter.tsx
@@ -126,6 +126,59 @@ export function FilterList({
     };
   }, []);
 
+  useEffect(() => {
+    // TODO: why is this called after each setSelectionOption
+    // TODO: up / down also moves cursor inside MiniEditor to begin / end
+    function onKeyDown(e: KeyboardEvent) : boolean {
+      const nOptions = matchingOptions.length;
+      if (nOptions === 0) {
+        return false;
+      }
+      let newSelected : number | undefined;
+      switch (e.key) {
+        case "ArrowUp":
+          newSelected = selectedOption - 1;
+          break;
+        case "ArrowDown":
+          newSelected = selectedOption + 1;
+          break;
+        case "PageUp":
+          newSelected = selectedOption - 5;
+          break;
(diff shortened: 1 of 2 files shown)


zefhemel:
Thanks I'll have a closer look at this a bit later (holiday prep going on here). Key handling is a pain in the ass, so I need to do some proper testing to avoid any regressions.

web/components/filter.tsx
```diff
@@ -126,6 +126,56 @@ export function FilterList({
     };
   }, []);
 
+  useEffect(() => {
+    // TODO: why is this called after each setSelectionOption
```
zefhemel, line 130: Spoiler: because of line 177

web/components/filter.tsx
```diff
@@ -126,6 +126,56 @@ export function FilterList({
...
+      return true;
+    }
+
+    document.addEventListener("keydown", onKeyDown);
+
+    return () => {
+      document.removeEventListener("keydown", onKeyDown);
+    };
+
+  }, [selectedOption, matchingOptions]);
```
zefhemel, line 177:
```suggestion
  }, []);
```

to do this just once?

zefhemel:
I think this shouldn’t happen through a global event listener, instead the MiniEditor should be expanded with an `onKeyDown` event handler probably.

web/components/filter.tsx
kjk: I'm totally React newbie and yeah, I figured out that those are dependencies and when selectOption changes it causes that function to re-run.

The thing is: it shouldn't be necessary to re-bind event handler on every state change, only when the element mounts. I had no problem getting that in Svelte.

However, if I don't pass selectOption as dependency then it somehow gets memoized and only the first modification takes effect.

The docs hint that state changes must happen inside element but I don't know how to push that code there.

Like I said: React newbie. Naively there should be a way to only bind event handler on mount and still modify state from handler but didn't figure out how.

kjk:
I tried that as my first approach (i.e. added onkeydown to MiniEditor and hooked it up to MiniEditor and FilterList, same as onkeyup) but it straight up didn't work i.e. onkeydown handler in Code Mirror wasn't called. 

Also: a remaining mystery is why up / down also affects Code Mirror (moves cursor to start/end of field) even when I stop propagation after handling up/down keydown.

zefhemel:
@kjk sorry for butting into your PR 😆  Didn't realize I could just push commits here. What I did is basically move your code into the mini editor component so that other places can use a similar mechanism. Indeed, you were right, CM doesn't propagate the keydown event so you have to capture it at a higher level.

PR #239, opened 2022-12-27
Fixes #129: add "attachment" query source

### LINKED ISSUE #129: Add "attachment" query source
You can query pages, but not attachments right now. This is silly. Needs to be fixed.

ekaragodin:


zefhemel (approved):
Great, thank you!

PR #254, opened 2022-12-30
Prevent losing commands with context

ekaragodin:


zefhemel:
Sorry, I forgot about this one. Will check it tomorrow.

zefhemel:
I'll merge this because it's a bit cleaner, but I'm not sure it changes any behavior?

ekaragodin:
@zefhemel Sorry, I should have described it in the beginning. When a user opens the command palette for the first time the reducer gets all commands from the state, filters out all commands with contexts (e.g. Task context), and saves them back into the state. When the user moves the caret to the task and opens the command palette again there are no commands with Task context.

PR #277, opened 2023-01-04
Update Query.md documentation 

tristansokol:
clarify documentation that the data sources are space wide, not just for the page where the query lives.


zefhemel:
Thanks!

Issue #210, opened 2022-12-16
"Refactoring: Extract to Page" command


zefhemel:
Would cut the current selection (or if nothing is selected the current section based on header?), ask for a page name, replace it with a page reference and move the content to the new page.

yorrd:
also interesting, "merge with page"

zefhemel:
Yes, that makes sense. You could also decide to not merge it into core, but create a separate plug repo: https://github.com/silverbulletmd/silverbullet-plug-template

But honestly, I think this will be useful enough to have in core, so yes — I'd put this functionality in `plugs/core/refactor.ts` or similar.

PR #280, opened 2023-01-04
Update Query.md with information about "dynamic" query parameters. 

tristansokol:
Add in a description of hardcoded variables you can use in queries, seems pretty relevant for bullet journal style scripting of tasks. 

Since this is copied from core/templates, and likely to change(?) maybe it would be best to use an #include directive and break out into a separate file?

related: https://github.com/silverbulletmd/silverbullet/discussions/147


zefhemel:
We can refactor this later (when we get #210 😉 ), for now this is good to have!

PR #286, opened 2023-01-04
add in new refactor file and extract to page command; #210

tristansokol:
- `Refactor: Extract to new page` seems cumbersome, so I omitted the refactor. I'm not sure if the "category" is necessary for core plugins, but seems like a good idea for 3rd party. 
- mostly this is based on the existing plugs for renaming  (including the file overwrite protection) and text editing
- Happy to write some tests, but didn't see any infrastructure for it
- I don't know what `cmdDef.page` is doing, it seems like you can declare a argument in the yaml file, but I'm not sure how you would invoke it for page rename for example. Is that for use by other functions?

### DIFF
--- a/plugs/core/core.plug.yaml
+++ b/plugs/core/core.plug.yaml
@@ -336,7 +336,6 @@ functions:
     path: ./refactor.ts:extractToPage
     command:
       name: "Extract text to new page"
-      page: ""
 
   # Plug manager
   updatePlugsCommand:
--- a/plugs/core/core.plug.yaml
+++ b/plugs/core/core.plug.yaml
@@ -21,7 +21,7 @@ functions:
   setEditorMode:
     path: "./editor.ts:setEditorMode"
     events:
-    - editor:init
+      - editor:init
   toggleVimMode:
     path: "./editor.ts:toggleVimMode"
     command:
@@ -330,6 +330,13 @@ functions:
       name: "Text: Marker"
       key: "Alt-m"
       wrapper: "=="
+
+  # Refactoring Commands
+  extractToPageCommand:
+    path: ./refactor.ts:extractToPage
+    command:
+      name: "Extract text to new page"
+      page: ""
 
   # Plug manager
   updatePlugsCommand:
--- a/plugs/core/refactor.ts
+++ b/plugs/core/refactor.ts
@@ -37,5 +37,5 @@ export async function extractToPage(cmdDef: any) {
   console.log("Writing new page to space");
   const newPageMeta = await space.writePage(newName, text);
   console.log("Navigating to new page");
-  await editor.navigate(newName, cursor, true);
+  await editor.navigate(newName);
 }
--- a/plugs/core/refactor.ts
+++ b/plugs/core/refactor.ts
@@ -1,0 +1,41 @@
+import {
+  editor,
+  space,
+} from "$sb/silverbullet-syscall/mod.ts";
+
+export async function extractToPage(cmdDef: any) {
+  console.log("Got a target name", cmdDef.page);
+  const cursor = await editor.getCursor();
+
+  const newName = cmdDef.page ||
+    await editor.prompt(`New page title:`, 'new page');
+  if (!newName) {
+    return;
+  }
+
+  console.log("New name", newName);
+
+  try {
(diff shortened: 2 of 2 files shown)


tristansokol:
ahh, with the github check I notice that the test are side by side, but mostly for the plugos. Let me know what you think and I can dig into deno testing

plugs/core/refactor.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  editor,
+  space,
+} from "$sb/silverbullet-syscall/mod.ts";
+
+export async function extractToPage(cmdDef: any) {
```
zefhemel, line 6: Leave out this while cmdDef parameter, you don't need it. Also in the lines after. You'll want to ask for a page name anyway.

plugs/core/refactor.ts
```diff
@@ -0,0 +1,41 @@
...
+      // Expected not found error, so we can continue
+    } else {
+      await editor.flashNotification(e.message, "error");
+      throw e;
+    }
+  }
+  let text = await editor.getText();
+  const selection = await editor.getSelection();
+  text = text.slice(selection.from, selection.to);
+  await editor.replaceRange(selection.from, selection.to, "");
```
zefhemel, line 36: I'd imagine your wouldn't simply remove the text, but replace it with a `[[page reference]]` to the extracted page?

plugs/core/refactor.ts
```diff
@@ -0,0 +1,41 @@
...
+  console.log("Writing new page to space");
+  const newPageMeta = await space.writePage(newName, text);
+  console.log("Navigating to new page");
+  await editor.navigate(newName, cursor, true);
```
zefhemel, line 40: The cursor position would be off here. Navigating to the new page makes sense, but I'd simply call `editor.navigate(newName)` here. That's it.

zefhemel:
Looking good, added a few comments. Adding tests is still cumbersome, I'd have to think how to make this easier. Let's leave it for now.

zefhemel (changes requested):
Looking good, added a few comments. Adding tests is still cumbersome, I'd have to think how to make this easier. Let's leave it for now.

plugs/core/refactor.ts
tristansokol: What do you think the typical use case for this is going to be? I'm not sure this is a command I'd use myself. 

Are you imagining pages that eventually become too big and need to be broken out, so this would be a shortcut to that? (not opposing your suggestion in anyway, I just don't have enough of a mental picture of how this would be used to have an opinion on what a good option would be)

plugs/core/refactor.ts
tristansokol: Could you briefly describe what the cmdDef does? Why does the renaming use it?

plugs/core/refactor.ts
tristansokol:
```suggestion
  await editor.navigate(newName);
```

plugs/core/core.plug.yaml
```diff
@@ -331,6 +331,13 @@ functions:
       key: "Alt-m"
       wrapper: "=="
 
+  # Refactoring Commands
+  extractToPageCommand:
+    path: ./refactor.ts:extractToPage
+    command:
+      name: "Extract text to new page"
+      page: ""
```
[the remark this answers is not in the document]
tristansokol:
```suggestion
```

plugs/core/refactor.ts
zefhemel: Exactly. I'm imagining (hence the name) "extract method" type of functionality: text getting too long, let's move this to a separate page and replace it with a link.

plugs/core/refactor.ts
zefhemel: Generally `cmdDef` gets passed whatever is in the command definition in the plug.yaml file, so its name, and any additional attributes put in there. The page rename is a unique case in that I call it directly from the UI when you update the page name in the top bar. This happens nowhere else in the code base. It's a shame you happened to find this case and copied 😉

zefhemel:
Code looks good. Tested this locally and seems to work great. Thanks!

PR #290, opened 2023-01-05
update listify command to work with selection and add keyboard shortcut

tristansokol:
I originally wanted to add in just the keyboard shortcut (cmd+shift+8 I think is by far the most common, google/slack etc.) but also fixed a bug /expanded scope that allows you to listify the current line that the cursor is in, not having to select text first. 

I think ideally listifying would be more like a toggle (subsequent invocations would de-listify) instead of the current behavior where you just add bullets with each invocation. 

As always, open to whatever feedback and discussion!

### DIFF
--- a/plugs/core/core.plug.yaml
+++ b/plugs/core/core.plug.yaml
@@ -293,6 +293,8 @@ functions:
     path: ./text.ts:listifySelection
     command:
       name: "Text: Listify Selection"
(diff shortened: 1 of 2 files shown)


zefhemel:
Looks reasonable. Indeed, making it a toggle would be nicer, maybe a future change? Merging this one for now.

PR #303, opened 2023-01-11
WIP implementation of dead link searching

tristansokol:
Open points I'd like some feedback on before finalizing structure, cleaning up, properly attributing the included code, etc. 

- Is this desired to be a core plugin? 
- Is there a better UI that just writing a page? Is it bad to just write over the page?  
  - Is there a way to embed commands (with arguments) in the page? I was thinking it might be nice to have a "rename dead link to suggestion" type button. and embedding `{[rename(a,b)}] might be useful. 
- it might be great to have an idea around testing, I'm sure there are lots of edge cases my test space didn't show. Not sure what the performance would be with thousands+ files. 
- is it worth filtering to page names that have distances under a certain value? showing multiple suggestions?
Jan-11-2023 12-51-15

### DIFF
--- a/plugs/core/page.ts
+++ b/plugs/core/page.ts
@@ -278,3 +278,34 @@ export async function parseIndexTextRepublish({ name, text }: IndexEvent) {
     tree: await markdown.parseMarkdown(text),
   });
 }
+
+export async function findDeadLinks() {
+  //https://github.com/gustf/js-levenshtein/blob/master/index.js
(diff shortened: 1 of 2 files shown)


zefhemel:
Interesting questions:

1. Desired as a core plug-in? I think it can be part of it, but I don’t have a super preference, to be honest.
2. I think a page makes sense. I see two approaches, yours is probably the simplest but it “pollutes” the space a bit. Another would be to create a virtual page, this is possible and it’s how the full text search works: https://github.com/silverbulletmd/silverbullet/blob/sync/plugs/core/core.plug.yaml#L166 basically can “own” part of the page name space and implement it as functions. In this particular case it matches everything starting with a looking glass emoji, but you can also just put in “dead links” there, which would dynamically calculate the result upon request. Not sure if this is the right approach, since rendering this page could take very long so your approach may be better? A third alternative is to load it in a side panel, e.g. like the Backlinks plug: https://github.com/Willyfrog/silverbullet-backlinks maybe this is the better approach?

zefhemel:
For testing there’s no great to mock syscalls yet, however you can write unit tests. Have a look at any of the “.test.ts” files in the repo of how to do this. Tests are run via `deno task test` or `deno task test path/to/your/test.ts` to run just the one.

As to code style, one of the great things that Deno brings is no-discussion-around-formatting with `deno fmt` which you can run on this file (or the whole repo in one go). If you use VS Code, you can install the Deno extension to get the formatter integrated.

plugs/core/page.ts
```diff
@@ -278,3 +278,34 @@ export async function parseIndexTextRepublish({ name, text }: IndexEvent) {
...
+  let pages = await space.listPages()
+  pages = pages.map(p => p.name)
+  let content = await Promise.all(
+    pages.map(p =>
+      space.readPage(p).then(text => Array.from(text.matchAll(/\[\[([^\[\]\v]+)\]\]/gm), m => m[1]))
```
zefhemel, line 289: While this is concise and fast, it’s also not always correct. For instance, this would also match page-like links in contexsts where Markdown doesn’t accept them (e.g. in back-ticked code snippets, fenced code blocks etc.). The more proper way would be to actually parse the text as markdown, and find all page references through tree traversal, there’s an example of that here: https://github.com/silverbulletmd/silverbullet/blob/main/plugs/core/page.ts#L164

zefhemel: Actually I’d use the `collectNodesMatching` like here: https://github.com/silverbulletmd/silverbullet/blob/main/plugs/core/page.ts#L47

zefhemel:
Closing this as it's rather stale.

PR #382, opened 2023-03-29
Update Android build

joemccall86:
This PR allows the app to be built with the latest version of Android Studio.
It fixes a dependency issue, so all that needs to be done is `./gradlew
assemble` (or run it from android studio)

* Update the SDK compile version to 33 to fix dependency issues
* Update the Build plugin as a dependency


zefhemel:
Perfect, thanks!

PR #387, opened 2023-04-10
Add support for multiple order by

### LINKED ISSUE #385: Support multiple `order by` clauses
Currently when we use multiple order by clauses or use comma separated fields as an argument to order by clause, only first one is respected
```
<!-- #query task  where tags=["p0", "p1"] order by deadline order by tags desc where done=false render [[template/task]]-->
```
Will only order by deadline and not tags

siddhant3s:
Fixes #385 

* The grammar already supports multiple `order by` clauses
* Modified `ParsedQuery` to be able to contain multiple `order by` clauses by adding `ordering: QueryOrdering[]`
* Added the logic to do sorting by multiple keys
* Invariant: The behavior on single key sort is kept as before.

An Interesting bug that surfaced: When comparing tags, which are array, in earlier implementation the comparison always resulted in a "===" being false even when tags were equal. This automatically resulted in `a` being greater than `b`. For a single key it didn't matter. But kf we keep that implementation as is (which first compares the a and b by an `===`) then that would mean that it will always return false on comparison which would mean that we will never evaluate the latter keys.

### DIFF
--- a/plug-api/lib/query.ts
+++ b/plug-api/lib/query.ts
@@ -13,11 +13,25 @@ export type QueryFilter = {
   value: any;
 };
 
+export type QueryOrdering = {
+  orderBy: string;
+  orderDesc: boolean;
+};
+
 export type ParsedQuery = {
   table: string;
-  orderBy?: string;
-  orderDesc?: boolean;
   limit?: number;
+  ordering: QueryOrdering[];
+  /** @deprecated Please use ordering.
+   * Deprecated due to PR #387
+   * Currently holds ordering[0] if exists
+   */
+  orderBy?: string;
+  /** @deprecated Please use ordering.
+   * Deprecated due to PR #387
+   * Currently holds ordering[0] if exists
+   */
+  orderDesc?: boolean;
   filter: QueryFilter[];
   select?: string[];
   render?: string;
@@ -97,22 +111,22 @@ export function applyQuery<T>(parsedQuery: ParsedQuery, records: T[]): T[] {
       resultRecords.push(recordAny);
     }
   }
-  // Now the sorting
-  if (parsedQuery.orderBy) {
-    resultRecords = resultRecords.sort((a: any, b: any) => {
-      const orderBy = parsedQuery.orderBy!;
-      const orderDesc = parsedQuery.orderDesc!;
-      if (a[orderBy] === b[orderBy]) {
-        return 0;
-      }
 
-      if (a[orderBy] < b[orderBy]) {
-        return orderDesc ? 1 : -1;
-      } else {
-        return orderDesc ? -1 : 1;
+  if (parsedQuery.ordering.length > 0) {
+    resultRecords = resultRecords.sort((a: any, b: any) => {
+      for (const { orderBy, orderDesc } of parsedQuery.ordering) {
+        if (a[orderBy] < b[orderBy] || a[orderBy] === undefined) {
+          return orderDesc ? 1 : -1;
+        }
+        if (a[orderBy] > b[orderBy] || b[orderBy] === undefined) {
+          return orderDesc ? -1 : 1;
+        }
+        // Consider them equal. This way helps with comparing arrays (like tags)
       }
+      return 0;
     });
   }
+
   if (parsedQuery.limit) {
     resultRecords = resultRecords.slice(0, parsedQuery.limit);
   }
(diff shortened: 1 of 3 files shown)


siddhant3s:
Tested using:
```
* [ ] Task B 📅2023-04-09 #p2
* [x] Task D 📅2023-04-09 #p1
* [x] Task C 📅2023-04-10 #p2
* [ ] Task A 📅2023-04-10
* [ ] Task E  #p2

<!-- #query task order by tags  -->
|name       |done |deadline  |page |pos|tags|
|-----------|-----|----------|-----|---|--|
|Task A     |false|2023-04-10|index|160|  |
|Task D  #p1|true |2023-04-09|index|100|p1|
|Task C  #p2|true |2023-04-10|index|130|p2|
|Task E  #p2|false|          |index|186|p2|
|Task B  #p2|false|2023-04-09|index|70 |p2|
<!-- /query -->

<!-- #query task order by tags order by deadline -->
|name       |done |deadline  |page |pos|tags|
|-----------|-----|----------|-----|---|--|
|Task A     |false|2023-04-10|index|160|  |
|Task D  #p1|true |2023-04-09|index|100|p1|
|Task E  #p2|false|          |index|186|p2|
|Task B  #p2|false|2023-04-09|index|70 |p2|
|Task C  #p2|true |2023-04-10|index|130|p2|
<!-- /query -->

<!-- #query task order by tags order by done desc -->
|name       |done |deadline  |page |pos|tags|
|-----------|-----|----------|-----|---|--|
|Task A     |false|2023-04-10|index|160|  |
|Task D  #p1|true |2023-04-09|index|100|p1|
|Task C  #p2|true |2023-04-10|index|130|p2|
|Task E  #p2|false|          |index|186|p2|
|Task B  #p2|false|2023-04-09|index|70 |p2|
<!-- /query -->
```

zefhemel:
This is great, thanks for doing this. My concern is that the `applyQuery` implementation is "compiled" into many shipping plugs right now (they just import it directly, and via the ES Build compilation process include it verbatim in their code). Unless they're recompiled with the new applyQuery implementation (so update their dependencies), they'd get the new version of the `ParsedQuery` structure from the parser (because this part ships with SB itself), and then attempt to process it based on its old structure (with `orderBy` and `orderDesc` properties) which won't exist. 

In effect, this will probably break all existing plugs (not shipping with SB) that provide a query provider.

There are a few options:

1. We say "screw it": let's rip the bandaid and fix this (although likely adding `or` to queries down the line will break things again)
2. We _add_ the multiple ordering property as an extra field and leave the old one in place, and continue to support both for some time. I think this should work without breaking stuff.

To test if things are breaking, you could use this plug as a reference point: https://github.com/silverbulletmd/silverbullet-github

Even though it complicates the code I'd opt for option 2 right now.

zefhemel:
Ah, and regarding the failing "tests", be sure to run `deno task check` to run a TypeScript typecheck on the whole code base.

siddhant3s:
I didn't run the test. I'll do that an update the main README.md to include that step.

Thanks for raising the concern. It really depends on what kind of compatibility we have promised our users. 

For now, I'm happy to do 2. but let's talk about what the path forward should be.
Here are some lose thought I have (before injesting any caffein) 
1. We should add deprecated tags to the old fields so that we can remove them in the next major release and break all existing plugins
2. If plugs do depend on a core piece of code, then we're looking at some version resolution or package management system in long run.
3. We mark what data structures are considered the contract interface from core so that we're careful when changing it (say we explicitly use something like protobuff to make it super clear)

siddhant3s:
I didn't run the test. I'll do that an update the main README.md to include that step.

Thanks for raising the concern. It really depends on what kind of compatibility we have promised our users. 

For now, I'm happy to do 2. but let's talk about what the path forward should be.
Here are some lose thought I have (before injesting any caffein) 
1. We should add deprecated tags to the old fields so that we can remove them in the next major release and break all existing plugins
2. If plugs do depend on a core piece of code, then we're looking at some version resolution or package management system in long run.
3. We mark what data structures are considered the contract interface from core so that we're careful when changing it (say we explicitly use something like protobuff to make it super clear)

zefhemel:
Your suggestions makes sense to me and code looks good. Just want to test it a bit locally before merging it. Thanks for doing this!

siddhant3s:
Of course! Sounds good. I'm not sure why these test are failing here. On my system I made sure to run the tests. Let me know how it goes.

zefhemel:
Ok this seems to work, let's go! Thanks for this!

PR #388, opened 2023-04-11
Start SilverBullet on boot using systemctl

vektren:
# Overview 
The use case will be to have SilverBullet cloud hosted on Linode (or any other virtual server). Using systemctl enables SilverBullet to start when the host starts up. Normally simple to set up; however, as SilverBullet is a binary installed and run through Deno, the system is unaware of Deno at startup. 
## Setup background

1. Installed Debian Linux on Linode and followed instructions to [install Deno and Silverbullet](https://silverbullet.md/Server)
2. As it's running on Internet, I placed the server behind a Cloudflare Zero Trust tunnel and started Silverbullet using the `--user username:password` to allow me to log in.
4. Created a folder in my home directory called pages and moved all my `.md` here.
5. Then created a system service so that when I restart the host, SilverBullet starts up.

**NOTES**

* This is not the most secure deployment; the username and password, along with all your files, are stored on a server not owned by you.
* Ensure you have a firewall setup to only allow access to the server on the appropriate ports. `Example: port:80 and 443` when using Cloudflare or another reverse proxy.
* I do not allow root to log in over SSH and limit access to SSH from my IP. 
* Implement good access control and general security hygiene to prevent data loss.

## systemctl Configuration

First, create a shell script that will launch SilverBullet; this will include all the required syntax to run as a service. 

```bash
sudo vi /usr/bin/silverbullet.sh
```

The file's contents should look like this; change details such as  username to match your system. 

```bash
#!/bin/bash
# Script to start SilverBullet through Deno

/home/username/.deno/bin/deno run --allow-all --unstable --no-config https://get.silverbullet.md/ /home/username/yourmarkdownfiles --user webusername:passowrd
```

Make sure the script is executable: 

```bash
sudo +x /usr/bin/silverbullet.sh
```

Test the script, this command should run and show details on the SilverBullet URL etc..

```bash
/usr/bin/silverbullet.sh
```

Adding the script to systemctl you need to create a service file:

```bash
sudo vi /etc/systemd/system/silverbullet.service
```

Contents should look like this: 

* Description=**SilverBullet**
* Type=**simple**
* ExecStart=**silverbullet.sh** - *the script we created before*
* WantedBy=**multi-user.target** - *will start once Linux is at the correct run level*

```bash
[Unit]
Description=SilverBullet

[Service]
Type=simple
ExecStart=silverbullet.sh

[Install]
WantedBy=multi-user.target
```

Then enable the service (adding it so it starts automatically)

```bash
sudo systemctl enable silverbullet.service
```

And then start the service:

```bash
sudo systemctl start silverbullet.service
```

If all goes well, the server will be started on the default port, to troubleshoot, use `journalctl -xe`. This will output a log showing errors on failed services. 

**another note**

* Please let me know if there are things I could do better; I just googled to find the solution. I am sure there might be a better way.
* Repeating from before, do what you can to lock down the installation.


zefhemel:
I'd really like to have this, but your PR somehow modifies a TON of files. Could you figure out what's happening there? I'd expect a few files change only.

zefhemel:
Closing this because it's stale

Issue #111, opened 2022-11-04
ä-Umlaut shown as ‘


khooler:
If you type the umlaut ä on a german keyboard it is shown as ‘ in the editor. All other umlaut (ö,ß,ü) are correct

zefhemel:
Strange bug, but I think I fixed it, could you verify?

j6b72:
The bug seems fixed on my side, I can insert "ä" again without any issues. Thank you!

khooler:
Yes. Can confirm, it is working now. Thank you!

PR #396, opened 2023-05-04
feat: make darkmode more readable

jim-fx:
Hi @zefhemel :)

Silverbullet is an awesome app that I thoroughly enjoy. Thanks for creating it! In fact, I even tried to build [something similar](https://github.com/jim-fx/notarium) myself.

Among the many features of Silverbullet, I particularly enjoy using the vim-mode and dark theme. However, I believe there is still some room for improvement to the readable of the dark theme.

Changes:
- switched dark mode background from `#000` to `#111`
- Made the vim status bar the same width as the content window
- made the vim selection a bit transparent so text is readable when selected
- made the top-bar and the vim-status bar the same background color
- increased the contrast of the query header so it is readable
- make the "frontmatter" text readable 

Edit: I just saw that there are some open Issues #10 #58 #127  and a PR ( #179 ) regarding the same topic, but they seam to be stale

Bildschirmfoto am 2023-05-04 um 14 42 40

Bildschirmfoto am 2023-05-04 um 14 33 32

Bildschirmfoto am 2023-05-04 um 14 54 32

### DIFF
--- a/web/styles/theme.scss
+++ b/web/styles/theme.scss
@@ -129,6 +129,17 @@
     background-color: #d7e1f6 !important;
   }
 
+
+  .cm-panels-bottom {
+      background-color: #e1e1e1;
+      border-top: #cacaca 1px solid;
+      .cm-vim-panel {
+        padding: 0 20px;
+        max-width: var(--editor-width);
+        margin: auto;
+      }
+  }
+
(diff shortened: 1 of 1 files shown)


zefhemel:
Thanks a lot for this!

Issue #136, opened 2022-12-03
Does not run on raspberry 


ulrischa:
Because deno dies not run on raspberry pi silverbullet does not run too. Would like to use it on a raspberry pi. 
Please give another Option to install

zephraph:
I think your best shot is following along with this issue. https://github.com/denoland/deno/issues/2295. It sounds like you probably _could_ self compile deno and then run sb, but your milage may vary. This isn't an installation issue. You'd have to rewrite the tool to not depend on deno.

ulrischa:
For me a show Stopper for raspberry.

zefhemel:
If you run a 64 bit OS on a recent RPI this should work: https://github.com/LukeChannings/deno-arm64/releases

zefhemel:
Let me put this on my roadmap list. I have some Pis lying around, would be nice to use them for SB. Maybe I can provide an arm 64 docker image too.

zefhemel:
Maybe we can easily switch to this base image to get support for both intel and arm: https://hub.docker.com/r/lukechannings/deno

A shame it's based on Ubuntu.

ulrischa:
Oh cool. Could you tell me more Details if it works?

zefhemel:
Ok, so this works flawlessly once you setup Deno. Instructions:

Tested on Ubuntu server 64-bit on a Pi, but should work the same for any 64 (!! 32-bit won't work) Linux distribution on a RPI:

```shell
$ apt install unzip
$ cd ~
$ wget https://github.com/LukeChannings/deno-arm64/releases/download/v1.29.1/deno-linux-arm64.zip
$ mkdir -p ~/.deno/bin
$ cd ~/.deno/bin
$ unzip ~/deno-linux-arm64.zip
```

Add `~/.deno/bin` to PATH, e.g. in `~/.bashrc` add:

```
export PATH=$PATH:~/.deno/bin
```

Then to verify:

```shell
$ source ~/.bashrc
$ deno -V
```

Then, just follow standard SB install instructions:

```shell
$ deno install -f --name silverbullet -A --unstable https://get.silverbullet.md
$ mkdir ~/Notes
$ silverbullet ~/Notes --hostname 0.0.0.0
```
And access it on `http://ip-of-pi:3000`

zefhemel:
Properly documented here: https://silverbullet.md/Raspberry_Pi_Installation

ulrischa:
Thank you for that christmas present :)
[...]

Issue #198, opened 2022-12-15
Release tags as Github releases + properly tagged docker images


zefhemel:
Right now we're in streaming release mode, every merge to `main` results in a new release, both to `get.silverbullet.md` and the `zefhemel/silverbullet:latest` docker image. Likely this will have to change.

When tagging a release:
* We can release the bundled `silverbullet.js` as a proper github release (this is 100% self contained and can be run locally)
* We can push a docker image with a proper version number

sweisgerber:
Hi there,

there are surely dozens of work-flows out there, one of the better known is `git-flow`.
- https://jeffkreeftmeijer.com/git-flow/, which has been put into a nice git plugin:
- https://github.com/nvie/gitflow, that get's explained in a blog-post:
- https://nvie.com/posts/a-successful-git-branching-model/

That would surely help with doing proper releases and their tagging, IMHO.

The main aspect is, that one does only work on branches, which will get merged to `develop`.

Every release is then created by merging & tagging `develop->main` through release-branches.
One does never push to main or develop, only to branches (perhaps you push to develop as well if you are the only dev, although it's not allowed when applying yhe work-flow strict).
This leads not only to `main` being always ready to run, but to `develop` as well.
When using the linked git plugin the git-flow is pretty easy.

O.c. there are plenty of alternatives like [github-flow](https://githubflow.github.io/) and [more](https://www.jetbrains.com/help/space/space-git-flow.html).

sweisgerber:
Anyway, what I actually wanted to post :D :

I'm currently working on github action workflow that uses tags as proper release numbers.
It also adds these new features
- uses buildx to build arm64 & amd64 (#136)
- pushes builds to dockerhub & the github container registry as well.
- tags with `latest`, `nightly` & `releases`

https://github.com/sweisgerber/silverbullet/pull/1

If you want, I can redirect the PR to your repository as well while I test the semantics on my fork.
Final tests have to get done by you anyway, as it needs tags and you have to configure github container registry (ghcr) access.

PR #400, opened 2023-05-08
Feature: Docker multiarch image (amd64 & arm64) dynamic tags and ghcr repositoy

sweisgerber:
Hello,

this pull-request targets:

- Fixes: https://github.com/silverbulletmd/silverbullet/issues/136
- Extends: https://github.com/silverbulletmd/silverbullet/issues/198

Additionally it also pushes to the github container registry (GHCR.io)

An Example registry deployment can get found for the imaginary next release:

- https://github.com/sweisgerber/silverbullet/releases/tag/0.2.14
    - DockerHub: https://hub.docker.com/layers/sweisgerber/silverbullet/0.2.14/images/sha256-81ec535cb4b71a01623f15aa93587af6ef6639254478d91f794d986bc829fc3f?context=explore
    - ghcr.io: https://github.com/sweisgerber/silverbullet/pkgs/container/silverbullet/91495509?tag=0.2.14

If desired I can rollback pushing to ghcr or add more tags (e.g. for branches).

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,9 +2,17 @@ name: Docker
 
 on:
   push:
+    branches:
+      - "main"
     tags:
-      - '*'
-
+      - "**"
+env:
+  DENO_VERSION: v1.32.5
+  # Docker & Registries
+  ARCHITECTURES: linux/amd64,linux/arm64
+  IMAGE_NAME: silverbullet
+  NAMESPACE_GITHUB: silverbulletmd
+  NAMESPACE_DOCKER: zefhemel
 jobs:
   docker-build-push:
     runs-on: ubuntu-latest
@@ -13,10 +21,20 @@ jobs:
       - name: Setup repo
         uses: actions/checkout@v3
 
+      - name: Set up QEMU for multi-arch builds with buildx
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: ${{ env.ARCHITECTURES }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          platforms: ${{ env.ARCHITECTURES }}
+
       - name: Setup Deno
         uses: denoland/setup-deno@d4873ceeec10de6275fecd1f94b6985369d40231
         with:
-          deno-version: v1.32.5
+          deno-version: ${{ env.DENO_VERSION }}
 
       - name: Run bundle build
         run: |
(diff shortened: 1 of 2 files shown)


zefhemel:
Oh this is excellent, I will merge this to see if this all works.

zefhemel:
Confirmed: https://hub.docker.com/r/zefhemel/silverbullet/tags there are now ARM and Intel builds, thanks a lot!

zefhemel:
A tagged release will happen for the next release I suppose

sweisgerber:
@zefhemel yes, that's right.

If you would like to have different semantics, like, e.g.
- latest for most recent tagged version and
- edge for the most recent build on master 

I could help with that as well. I think it's a matter of preference. :)

ahbnr:
Is it possible that switching the base image for deno also caused the user id to change?
Because my instance started crashing on startup since the update from yesterday with
```
Permission denied (os error 13): open '/space/data.db'
```

The files in my silverbullet volume all have the permissions 1000:1000.
However, executing `id` inside the container now returns
```
uid=1993(deno) gid=1993(deno) groups=1993(deno)
```

So its uid 1993 now, not 1000 [as in the old base image](https://github.com/silverbulletmd/silverbullet/blob/c8d0b2df146742d7e07d04f6eb8e647ad939c301/Dockerfile#L33).

Backing up all files and destroying the volume fixes it, because the re-created volume automatically gets the new uid 1993.

zefhemel:
Oh that's not good. I think a fair number of people are using this image and will have uid 1000 as the owner of their files. Yeah I think the best would be to add a `silverbullet` user with uid 1000 and use that.

sweisgerber:
Allright, I'll fix it shortly

sweisgerber:
Created issue: https://github.com/silverbulletmd/silverbullet/issues/402 @ahbnr

zefhemel:
@ahbnr the fix has now been merged and a new image pushed, can you confirm this resolves the issue?

ahbnr:
Sorry for the late reply!  
Thank  you very much for handling this issue so quickly, @sweisgerber @zefhemel .

I can not test the new image on the original volume where the error occurred, because I already deleted that one and recreated it with the 1993 owner permissions.
However, I have repeated the process of deleting the volume and recreating the files under the ownership of 1000:1000.
With this setup, I switched to the new image and it is working.

In short, the container seems to be running with the user 1000:1000 again, and I do not encounter any issues when the volume and the files are owned by 1000:1000.

So I think the fix is working.

PR #403, opened 2023-05-16
SilverBullet pivot to become an offline-first PWA

zefhemel:
SilverBullet has rapidly increased in terms of complexity over the last months by adding a desktop and mobile app in addition to the server mode. The goal of these additional app distribution models was the following:

1. Simplicity of deployment installation: simply download the app and run it (no need to run a server from a command line)
2. Offline mode without the need for a connection to a server
3. Deeper integration into the OS (so far no such features have been added)

So now we have:

1. Desktop apps for Mac, Windows and Linux
2. Mobile app for iOS (and you can build it yourself for Android)

However, this has lead to a lot of added development, build and deployment overhead. Since this is an open source, volunteer project, my time is limited and I struggle to test and maintain all of this.

Therefore the proposed direction here is to pivot the project (back, in a sense) to a PWA, but adapted to be fully offline capable.

To support the offline-capable part, the architecture shifts a little bit:

* All content (pages and attachments) are now synced to the browser and kept in IndexedDB. A copy is kept on the server (as files in a folder, nothing changes there) and the previously developed sync engine is now used to keep the copy in the browser and server in sync whenever there is connectivity. The server effectively becomes a "dumb" file store.
* All plug code execution (such as indexing) now happens in the browser only (a lot happened on the server in the past).

From the user perspective not much changes:
* the UI remains the same (although there is now an indication of whether its is connected to a server or not).
* upon first launch, in the background all your content will be synced into your browser's database.

A few implications:

1. This branch removes the desktop and mobile apps. They will no longer be maintained.
2. To run SB in offline mode, it needs to run either on localhost or via HTTPS, because otherwise service workers (which serve the files offline) are not available — this is a browser security "feature".

Since this is a big shift, I took the liberty to introduce another big infrastructural change: plugs are now no longer compiled to (big) JSON files, but to (much smaller) JavaScript bundles. In principle, all that needs to happen is to recompile existing plugs using `silverbullet plug:compile` and list the `.plug.js` in your `PLUGS` file.

This `local` branch is continuously deployed to https://local.silverbullet.md

To test this branch:

```
deno run -A --reload https://local.silverbullet.md/silverbullet.js <your-pages-path>
```

You can delete the "data.db" file, because the server no longer keeps index data.

If you get into a weird state, you can flush all your caches and copies and stuff by visiting http://localhost:3000/.client/reset.html


jimt:
I think the correct command for testing is:
`deno run -A --reload https://local.silverbullet.md/silverbullet.js <your-pages-path>`
(deno expects options before the script argument.)

zefhemel:
@jimt thanks! Fixed in the description

PR #404, opened 2023-05-20
Accept SB_USER environment variable as a secondary source of creds

rjgibson:
Allow BasicAuth credentials to be specified with the `SB_USER` environment variable. This prevents the password from being revealed when listing processes with the ps command.


zefhemel:
Thanks!

PR #418, opened 2023-06-13
Add copy page to core plugin

JordanPaoletti:


plugs/core/page.ts
```diff
@@ -94,6 +94,39 @@ export async function deletePage() {
...
+  const oldName = await editor.getCurrentPage();
+  const newName = await editor.prompt(`New page title:`, `${oldName} (copy)`);
+
+  if (!newName) {
+    return;
+  }
+
+  try {
+    // This throws an error if the page does not exist, which we expect to be the case
+    await space.getPageMeta(newName);
```
JordanPaoletti, line 107: Was there a reason a `space.pageExists(name)` function hasn't be created? The heavy use of exceptions for control flow feels like an anti-pattern. I'd like to include refactoring that into this PR as well, but was unsure if there was an underlying reason.

zefhemel: No reason, adding it would be a good idea. I wouldn't want to add it to the `SpacePrimitives` interface, but having a syscall for it that wraps the `try-catch` could be nice.

zefhemel:
Do you want to attempt to add a `pageExists` to this PR or shall we merge this as in and do that separately?

JordanPaoletti:
Probably best to have it as a separate PR to keep the changes for this one localized to the specific copy functionality.

zefhemel:
Agreed

PR #437, opened 2023-06-22
WIP: Early work on federation

zefhemel:


zefhemel:
Update on this PR: I will likely revert the decision to move to relative links, but there's a lot of additional changes in this branch that are still worth reintegrating.

Issue #425, opened 2023-06-20
Query blocks take up a lot of visual realestate 


firasdib:
When the query blocks are not being edited, they still render the text inside of them, which makes them take up a lot of space. For more complicated queries, this can be quite a few lines. I propose that they are collapsed/hidden when not being edited. I have personally applied the following css rule to mitigate this issue on my end:

```css
  .sb-directive-start-outside {
    white-space: nowrap;
    overflow: hidden;
    pointer-events: none;
  }
```

zefhemel:
I'll play a bit with your suggestion, but I sense this is just the consequence of solving this issue: https://github.com/silverbulletmd/silverbullet/issues/208

I'm not sure we can have both.

zefhemel:
On mobile (or technically: narrow screens of 600px or less) this implemented: the height of the "directive bar" is always the same, on wider screen (likely desktops, tablets) this is not the case to avoid vertical jumping as you walk your cursor through the document or click places. It's a compromise, a trade-off. Closing this for now.

Issue #208, opened 2022-12-16
Eliminate vertical jumping


zefhemel:
When moving the cursor (by keyboard or mouse), a lot of vertical jumping happens around certain areas:

* [x] Frontmatter
* [x] Directives
* [x] Fenced code blocks

This is highly distracting, and sometimes leads to surprising results, like following links that jump “under” the cursor. While I think it’s visually clean to hide some of the markdown clutter, this vertical space should still be there in order to avoid this type of jumping.

zefhemel:
Here is an example for fenced code blocks:

https://user-images.githubusercontent.com/59073/208165010-00a0b0b0-fd84-498e-b575-4e153a0874d5.mov

zefhemel:
And directives


https://user-images.githubusercontent.com/59073/208165295-d420c8d3-2d4e-4ba0-8f8f-564b787b0ceb.mov

zefhemel:
Should be much better now, these element will always take up more vertical space though. Open to styling improvements.

jdiercks:
So far I'm willing to accept the jumpiness, we'll see if the click-misfire risks eventually convince me otherwise, but at least with STYLES we can choose. I even took it to the extreme with this, which visually I really like:

```css
/* collapse query/code block wrappers when hidden */
.sb-directive-start-outside, 
.sb-directive-end-outside, 
.sb-line-code-outside {
  white-space: nowrap;
  overflow: hidden;
  pointer-events: none;
  height: 1px;
}

```

PR #441, opened 2023-07-01
adjust directive header's height when cursor is outside

vuau:
The directive is taking up so much precious space in mobile view.

Screenshot from 2023-07-01 16-08-36

It should be smaller when the cursor is outside of the directive
Screenshot from 2023-07-01 16-27-11

And become bigger when the cursor is inside the directive
Screenshot from 2023-07-01 16-18-26


zefhemel:
I think this is the same as #425 see why I did this here: https://github.com/silverbulletmd/silverbullet/issues/208

vuau:
> I think this is the same as #425 see why I did this here: https://github.com/silverbulletmd/silverbullet/issues/208

Yes same. Have you merged those improvements to the main branch? I checkout it yesterday and the issue still remain

zefhemel:
The issue explains why I introduced this "problem" and why.

vuau:
So sorry I misread your point. My fix is just about reducing the vertical space and it doesn't cause the cursor jump as in my test

https://github.com/silverbulletmd/silverbullet/assets/259848/1617ed6d-235a-4c8c-8526-aa68e582ac7e

zefhemel:
This is indeed less of an issue on mobile, where you don't usually "walk" the cursor down (via the keyboard) but you just tap places. The vertical jump happens then (as you can also see in the video) but it's less of an issue, so maybe a compromise would be to do fold this vertically _just_ on mobile? An argument could also be the more limited screen real estate.

For this I/somebody would need to add a "on mobile" type CSS selector on the body tag maybe to make this distiction.

vuau:
Yeah. I think I can make that CSS if you don't have time to work on it. Do we have a JS function somewhere to check if SB is on mobile or can I just use CSS media-query?

zefhemel:
There's no function for this right now I don't think. A media query could indeed work, for certain size screens?

Issue #440, opened 2023-06-27
Failed to deploy to "Deno Deploy" due to missing module "hocuspocus/server@2.1.0"


kshehadeh:
Trying to deploy using deno.com and deployments fail with: 
```Error The deployment failed: Module not found "npm:@hocuspocus/server@2.1.0".```

![image](https://github.com/silverbulletmd/silverbullet/assets/107061/1b569cfb-964d-42c4-8bb3-c623bff1ff2d)

Just wondering if anyone else has tried running silverbullet on deno deploy and if they had any issues.

zefhemel:
This may be unblocked again soon with #449

zefhemel:
This should now be unblocked. Haven't tested, but these instructions should work again: https://silverbullet.md/Install/Deno%20Deploy

PR #444, opened 2023-07-02
Add option to remove hocuspocus from bundle 

ishehadeh:
"Fixes" #440

This PR replaces `CollabServer` with an interface `ICollabServer` and a concrete implementation, `HocuspocusCollabServer`. It also adds another implementation, `NoOpCollabServer`, that does nothing.

A plugin is added to esbuild in `build_bundle.ts` that replaces the import of `HocuspocusCollabServer` with `NoOpCollabServer` if the `SB_NO_COLLAB` environment variable is set.

Since the npm import of hocuspocus won't be included in the final bundle with `SB_NO_COLLAB` the bundle can be deployed.


zefhemel:
I may  want to take this one step further: #449

zefhemel:
Closing this, as real-time collab (and with that HocusPocus) has now been removed from SB entirely. Thanks for the effort though!

PR #452, opened 2023-07-05
Improvements for mobile view

vuau:
As discussed in https://github.com/silverbulletmd/silverbullet/pull/441, I came up with hidding the directive start and end blocks for small devices only (assuming the screen width is smaller than 600px). Also improve the display of tables to show a horizontal scroll bar instead of breaking words.

Screenshot from 2023-07-05 22-26-06

instead of

Screenshot from 2023-07-05 22-20-48

### DIFF
--- a/web/styles/editor.scss
+++ b/web/styles/editor.scss
@@ -259,3 +259,14 @@
 div:not(.cm-focused).cm-fat-cursor {
   outline: none !important;
 }
+
+@media only screen and (max-width: 600px) {
+  #sb-main .cm-editor {
+    .sb-table-widget {
+      overflow: auto;
+      th, td {
+        white-space: nowrap;
+      }
+    }
+  }
+}
--- a/web/styles/theme.scss
+++ b/web/styles/theme.scss
@@ -682,3 +682,11 @@ html[data-theme="dark"] {
   }
 
 }
+
+@media only screen and (max-width: 600px) {
+  #sb-editor {
+    .sb-directive-start-outside, .sb-directive-end-outside {
+      height: 22px;
+    }
+  }
+}


zefhemel:
Thanks for doing this! The table changes I'm on board with. What I'm not 100% convinced about is showing 0 indication that a certain block is rendered by a directive. What do you think about still having a gray border around it (like on wider screens) but perhaps with no or a fixed-height bar at the top and bottom? Perhaps just with this bit in there, but making the actual directive text display:none?

CleanShot 2023-07-06 at 10 45 18@2x

vuau:
How about this? If the goal is to distinguish a directive rendered block from others, I think we should keep the placeholder as well. I moved it out of the border for a better look.

Screenshot from 2023-07-07 10-50-52

zefhemel:
That's better, but I'd rather it mirror the look on "desktop" as much as possible in order not to confuse. Could it not just be made to look like this, just one line high?

CleanShot 2023-07-07 at 20 27 19@2x

vuau:
I think it's all good now

zefhemel:
This is what I'm seeing when inside the query (expected):
CleanShot 2023-07-10 at 08 39 52@2x

And this when I move outside:

CleanShot 2023-07-10 at 08 40 02@2x

There's still no indication that this is a directive block, nor the `#query` header. Is this what you intended?

vuau:
> This is what I'm seeing when inside the query (expected):
> 
> CleanShot 2023-07-10 at 08 39 52@2x
> 
> 
> 
> And this when I move outside:
> 
> 
> 
> CleanShot 2023-07-10 at 08 40 02@2x
> 
> 
> 
> There's still no indication that this is a directive block, nor the `#query` header. Is this what you intended?

No. My latest change is just to reduce the height. Maybe you can try resetting to my forked branch again. I force pushed several times that could be the problem.

zefhemel:
Merged here, thanks so much! https://github.com/silverbulletmd/silverbullet/commit/f7edfec6e04a68ec9e4e3d539a4e9d9131e2d5fc

PR #454, opened 2023-07-05
SilverBullet deployments examples

paletochen:
I created a deployments doc to provide a couple of examples on:
- How to deploy SB within docker and with caddy,redis and authelia
- How to deploy SB with Deno and set up a service to run it at boot time


zefhemel:
Thanks! I'll just merge this as is an make adjustments later.

PR #455, opened 2023-07-06
Refactor SCSS to use CSS variables

TheLD6978:
This pull request organizes and cleans up the scss files in preparation for css variable based theming (as proposed in https://github.com/silverbulletmd/silverbullet/discussions/58 and https://github.com/silverbulletmd/silverbullet/issues/127)

At the moment i just refactored the existing theme to use css variables and am quite sure not everything has a variable yet.

Also the customStyles should need less `!important` markers, as i moved the style tag below the main.css.

My idea going forward is the following:

1. create a css variable for almost everything. Make it quite verbose, so you can color your editor to look lika a rainbow if you wanted to.
2. Create a basic palette of colors which are assigned to all these verbose variables, but summarize common colors (like all subtle colors will use the --subtle-color var).

This should allow for fast creation of basic themes, but still leaves room for deep customization if desired. Also the verbose variables can change more frequently while basic themes may not break as fast if the variables names are chosen somewhat smart. If someone has any idea on what this basic theme should include (or if this is a good idea at all), please let me know.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -43,9 +43,8 @@
       };
     }
   </script>
-  <style id="custom-styles">
-  </style>
   <link rel="stylesheet" href="/.client/main.css" />
+  <style id="custom-styles"></style>
   <script type="module" src="/.client/client.js"></script>
   <link rel="manifest" href="/.client/manifest.json" />
   <link rel="icon" type="image/x-icon" href="/.client/favicon.png" />
--- a/web/styles/colors.scss
+++ b/web/styles/colors.scss
@@ -1,0 +1,387 @@
+#sb-root {
+  font-family: var(--ui-font);
+  background-color: var(--root-background-color);
+  color: var(--root-color);
+}
+
+.cm-cursor {
+  border-left: 1.2px solid var(--editor-caret-color) !important;
+}
+
+#sb-top {
+  color: var(--top-color);
+  background-color: var(--top-background-color);
+  border-bottom: var(--top-border-color) 1px solid;
+}
+
+#sb-top.sb-sync-error {
+  color: var(--top-sync-error-color);
+  background-color: var(--top-sync-error-background-color);
+}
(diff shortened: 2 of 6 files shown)


zefhemel:
Thank you so much for taking initiative on this. Do you consider this ready to go? I probably have to do some testing to see if there's no visual regressions, it's hard to judge from the massive diff. Any risky areas/changes I should pay attention to?

TheLD6978:
The current changes should be ready to go as a refactor of the existing themes. I tried to change as little as possible in the light theme apart from merging really similar colors like `rgb(72, 72, 72)` and `rgb(74, 74, 74)` into the same color. I included my dark mode fixes from my comment on https://github.com/silverbulletmd/silverbullet/issues/127.

Things that we should talk about before we can really call this a theming system are:
- I only considered color in this PR, do we want to allow other things as well? Like font style and border style/width for example.
- Variable naming. I mostly just went for the first thing that came to mind while naming everything. Maybe the scoping (`--editor-xyz` vs `--ui-xyz`) needs more thought?

But this can be implemented in a later PR in my opinion.

Considering Testing:
Most of the changes are just dividing the css into a color.scss which colors things, and the editor.scss which does all other stuff like paddings and layout. I tried to remove all `!important` rules I could find and merged a few selectors I deemed redundant. That mostly applies to modals and the autocomplete dropdown, but as these are fairly obvious if broken I dont think there are any risky areas i did not check.

I tested everything with the website space included in the repo and mostly looked at the SilverBullet and Directive/Query Pages. There I added basic things like blockquotes and admonitions.

Two selectors i didnt check as i could not figure out how to create a state where they would appear: `.sb-line-comment` and `.sb-fenced-code-hide`.

zefhemel:
Ok, played a bit with this. I can see no regressions. Awesome work. Merging this!

PR #463, opened 2023-07-11
Move Viewport Top to Cursor on page navigate.

ishehadeh:
By default, codemirror would only scroll enough to get the cursor on screen when navigating to a new page position.

This can be disorienting, especially when navigating to anchors like on the [directives](https://silverbullet.md/%F0%9F%94%8C%20Directive) page.


zefhemel:
Thanks!

PR #474, opened 2023-07-16
Add page names to search index

### LINKED ISSUE #473: Global search index does not include page names
How to reproduce:
1. Create a page called "vbhucijnxdmk"
2. Search for that page with "Search Space" command
3. See that the page is not included in search results
image

I expected the page to be included here in the search results like this:
image

I'll create a PR to fix it later.

v411e:
Closes #473

image


zefhemel:
Nice, thank you!

PR #503, opened 2023-08-19
Add inline manifest documentation

ishehadeh:
Just a little documentation for manifest type definition. Hoping to add this to copy most of this info to the website at some point. But I thought putting it in code would be a good place to start.


zefhemel:
Very nice, thanks!

PR #513, opened 2023-09-04
Add filename and directive content to directive error messages

ishehadeh:
Quick addition to add a little context to directive error messages. Helps find the problem on a page with a lot of them.


zefhemel:
Thanks

PR #528, opened 2023-10-10
Index plug: index paragraph nodes

ishehadeh:
Only considers top level paragraphs (right below Document node).
Includes attributes and tags in the paragraphs.

Just something I threw together real quick after seeing the discussion on discord. Definitely needs more tests.

### DIFF
--- a/plugs/index/paragraph.ts
+++ b/plugs/index/paragraph.ts
@@ -1,0 +1,56 @@
+import type { IndexTreeEvent } from "$sb/app_event.ts";
+import { removeQueries } from "$sb/lib/query.ts";
+import { indexObjects } from "./api.ts";
+import {
+  addParentPointers,
+  renderToText,
+  traverseTree,
+  traverseTreeAsync,
+} from "$sb/lib/tree.ts";
+import { extractAttributes } from "$sb/lib/attribute.ts";
+
+export type ParagraphObject = {
+  ref: string;
+  tags: string[];
+  text: string;
+  // TODO: maybe it would be useful to have a list of the the headings above the paragraph
+  page: string;
+  startPos: number;
+  endPos: number;
+} & Record<string, any>;
+
+export async function indexParagraphs({ name: page, tree }: IndexTreeEvent) {
+  const objects: ParagraphObject[] = [];
+  await traverseTreeAsync(tree, async (p) => {
+    // only search directly under document
+    if (p.type == "Document") return false;
+    if (p.type != "Paragraph") return true;
+
+    const tags = new Set<string>(["paragraph"]);
+    // tag the paragraph with any hashtags inside it
+    traverseTree(p, (e) => {
+      if (e.type == "Hashtag") {
+        tags.add(e.children![0].text!.substring(1));
+        return true;
+      }
+
+      return false;
+    });
+
+    const attrs = await extractAttributes(p, false);
+    objects.push({
+      ref: `${page}:${p.from}:${p.to}`,
+      text: renderToText(p),
+      tags: [...tags.values()],
+      page,
+      startPos: p.from!,
+      endPos: p.to!,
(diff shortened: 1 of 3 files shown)


plugs/index/paragraph.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+
+      return false;
+    });
+
+    const attrs = await extractAttributes(p, false);
+    objects.push({
+      ref: `${page}:${p.from}:${p.to}`,
+      text: renderToText(p),
+      tags: [...tags.values()],
```
zefhemel, line 44: I was thinking: we probably should only index paragraphs that contain a hash tag, but I think this implicitly does that. Nevertheless, I'd still prefer a check a little earlier that says "no tag found, let's skip" if that makes sense.

plugs/index/paragraph.ts
```diff
@@ -0,0 +1,56 @@
...
+      page,
+      startPos: p.from!,
```
zefhemel, line 46: Thus far most indexes have just stored the start pos (as `pos`), for consistency I'd prefer that here too unless you think it's useful to also keep the end position?

plugs/index/paragraph.ts
ishehadeh: should be fixed with [5aad876](https://github.com/silverbulletmd/silverbullet/pull/528/commits/5aad876c52d6cc95b260f077b2cec993e2337ca0)

plugs/index/paragraph.ts
ishehadeh: I'm not sure about keeping the end position. It seemed like it *could* be useful so I added it but I can't  come up with an example.

removed in [5aad876](https://github.com/silverbulletmd/silverbullet/pull/528/commits/5aad876c52d6cc95b260f077b2cec993e2337ca0)

plugs/index/paragraph.ts
```diff
@@ -0,0 +1,56 @@
...
+        tags.add(e.children![0].text!.substring(1));
+        return true;
+      }
+
+      return false;
+    });
+
+    const attrs = await extractAttributes(p, false);
+    objects.push({
+      ref: `${page}:${p.from}:${p.to}`,
```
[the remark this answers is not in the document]
ishehadeh: should be fixed with [5aad876](https://github.com/silverbulletmd/silverbullet/pull/528/commits/5aad876c52d6cc95b260f077b2cec993e2337ca0)

plugs/index/paragraph.ts
ishehadeh: Right now it returns all paragraphs, with at least the `paragraph` tag.

I think this behavior could be kind of useful. Queries like `paragraph where text =~ /SilverBullet/ig` may be nice.

Off the top of my head: say you were keeping a journal in SB. It might be fun and interesting to query all mentions of a specific thing from your journal entries. Something like `paragraph where page =~ /journal\// and text =~ /Work/ig`.

zefhemel: Yeah, the only cost would be size of the index, but this may be fine. Have you checked that paragraphs actually end up in the index without a tag now? Because I don't think they will, since the `indexObjects` iterates over the `tags` and inserts one copy of the object for each, which would mean that if your `tags` is empty no items will be added at all. A good way to circumvent that is to always add at least "paragraph" to the list of tags.

ishehadeh: It adds a paragraph tag (`tags = new Set([“paragraph”]), I’ll check later on the size of the index. Maybe whether or not all paragraphs are indexed could be configurable?

zefhemel: Ah right cool. Should it be configurable? Maybe, but let's go with this for now.

zefhemel (approved):

PR #532, opened 2023-10-15
style queries a bit nicer :)

jim-fx:
Love the new query stuff!

There were some slight issues with readability on dark mode, so I made this PR.

- Moves the button bar right into the corner
- Make the background of the button bar a bit more consistent with the rest of the ui and more readable in dark mode
- Adds some slight padding to the left side of the results (10px)

Bildschirmfoto am 2023-10-15 um 23 02 48(1)


zefhemel:
I like it!

PR #555, opened 2023-11-02
Add a {time} template helper, reuse timestamp code from Quick Note as a niceTime

johnlunney:
I was looking to create an "interstitial journal" timestamp snippet, and found there was no {time} helper. I centralised the code from the Quick Note alongside niceDate.

This should yield hh:mm:ss (although I'd personally prefer hh:mm), bear in mind this is my first time writing TypeScript :)


zefhemel:
Very nice!

PR #561, opened 2023-11-11
Add docs for admonitions

v411e:
Create Admonitions.md

This is already mentioned in the `CHANGELOG.md`, but I think it is easier to find when it has a seperate page.


zefhemel:
Thanks!

PR #571, opened 2023-11-22
Upload file command with copy/paste logic

prcrst:
It works, I'd like to get rid of some of the copy/pasted code

### DIFF
--- a/plug-api/types.ts
+++ b/plug-api/types.ts
@@ -143,5 +143,6 @@ export type LintDiagnostic = {
 
 export type UploadFile = {
   name: string;
+  type: string;
   content: Uint8Array;
 }
--- a/plug-api/types.ts
+++ b/plug-api/types.ts
@@ -143,6 +143,6 @@ export type LintDiagnostic = {
 
 export type UploadFile = {
   name: string;
-  type: string;
+  contentType: string;
   content: Uint8Array;
 }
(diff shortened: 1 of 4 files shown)


plugs/editor/upload.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (file.content.length > maximumAttachmentSize) {
+        editor.flashNotification(
+            `Attachment is too large, maximum is ${maximumAttachmentSize / 1024 / 1024
+            }MB`,
+            "error",
+        );
+        return;
+    }
+
+    const suggestedName = resolve(folderName(await editor.getCurrentPage()), file.name);
```
zefhemel, line 40: Honestly I think you can replace the `resolve` call with a simple `folderName(...) + "/" + file.name` here and get rid of the fancy implementation of that.

prcrst: ok

plug-api/types.ts
```diff
@@ -143,5 +143,6 @@ export type LintDiagnostic = {
 
 export type UploadFile = {
   name: string;
+  type: string;
```
zefhemel, line 146: and maybe rename this to `contentType` for consistency with other parts of the code

prcrst: ok, I kept it consistent with `File`... I'll change it

zefhemel:
Pushed one minor tweak (for when uploading to the root folder). Merging!

PR #573, opened 2023-11-24
Initial implementation of command link arguments

prcrst:


plugs/editor/navigate.ts
```diff
@@ -87,7 +87,11 @@ async function actionClickOrActionEnter(
     }
     case "CommandLink": {
       const commandName = mdTree.children![1]!.children![0].text!;
-      await system.invokeCommand(commandName);
+      const argsNode = findNodeOfType(mdTree, "CommandLinkArgs");
+      const argsText = argsNode?.children![0]?.text;
+      // Assume the arguments are can be parsed as the innards of a valid JSON list
+      const args = argsText ? JSON.parse(`[${argsText}]`) : [];
```
zefhemel, line 93: I think this is a decent place to start (parsing it as JSON for now), I'd put a try-catch around it so it wouldn't blow up too badly if this malformed.

zefhemel:
I added some docs and a custom flash message thing (for demo purposes, really). I'd be ok to merge this as it is and iterate as required later.

prcrst:
Sounds good to me

PR #604, opened 2023-12-18
docs: update deployment guide for cloudflared

lroolle:
The user inside the container (defined by PUID and PGID environment variables) needs to have write access to this directory. Otherwise, the silverbullet would run in an error state:

*attempt to write a readonly database: Error code 8: Attempt to write a readonly database*


zefhemel:
Thanks!

Issue #609, opened 2023-12-23
Authentication Error with New Docker Install


cblt2l:
I am new to silver bullet & trying to run via the following docker compose file

```
services:
  silverbullet:
    image: zefhemel/silverbullet:latest
    restart: unless-stopped
    environment:
    - SB_USER="admin:test1234"
    - PUID=1044
    - GUID=100
    volumes:
      - /volume2/docker-silverbullet/:/space
    ports:
      - 3007:3000
```
Also, /volume2/docker-silverbullet is indeed owned by user 1044 with gid of 100.

When I navigate to "http://serverurl:3007"  in a browser I get the notification "You are not authenticated, redirecting to login page... ". Clicking 'Ok' then redirects me to a login page. I put in the credentials from the compose file (user: admin, password: test1234)  but I get a "Invalid username or password" message.

I checked the logs from docker but only see the following:

```
Updating page list cache
Updating page list cache
Unauthorized access, redirecting to auth page
Unauthorized access, redirecting to auth page
Updating page list cache
Updating page list cache
Updating page list cache
.....
```

I tried rolling back the docker image to the previous version 0.5.8 but ran into issues before the server even started.

Do I have a configuration issue or is this a bug?

zefhemel:
Try removing the quotes:

```
- SB_USER=admin:test1234
```

cblt2l:
Yep, that was it.

Thank you

PR #620, opened 2024-01-08
Update docs Install/Docker.md

qaqland:
remove the quotes which cause "Invalid username or password" https://github.com/silverbulletmd/silverbullet/issues/609


qaqland:
I carefully read the document and install with Docker. Met the same issue. 

It would be more appropriate to remove the quotation marks in the document.

zefhemel:
Ah yes, thanks!

Issue #246, opened 2022-12-28
Emoji aliases


zefhemel:
I use specific emoji for specific use cases but cannot always remember their name. It would be nice if I can specifics aliases for existing emoji that I can use to find them more easily. 

These aliases can be configure in `SETTINGS`

yorrd:
maybe instead of going through the pain of re-inventing, we could use an existing alias list?

I'm aware of this one for example: https://emojibase.dev/emojis . While additional configurability is nice, I have to admit I have never missed custom aliases in my applications if they have some kind of alias / tags available by which to discover the emojis

Issue #226, opened 2022-12-21
Make plugs more explicitly define settings


zefhemel:
Right now various plugs implicitly read values from `SETTINGS` possibly documenting them, possibly not. It would be nicer if plugs can explicitly define their settings as well as their schema (at least their type). Same for `SECRETS` I suppose.

Not only is this better for documentation, it would also enable building a automatically generated proper settings UI listing all settings, changing them etc. Something more intuitive than YAML in a page.

Not sure how this should be specified yet, one way would be to extend the `.plug.yaml` format at the PlugOS level, something along the lines of:

```yaml
name: core
settings:
  initPage:
    type: string
    default: index
    description: Specifies the first page to load when loading your space
  weeklyNoteTemplate:
    type: string
    default: template/Weekly Note
functions:
    # …
```

zefhemel:
The `type` field may need to be something along the lines of JSON schema maybe to allow more complex settings. For instance, the [silverbullet-mattermost](https://github.com/silverbulletmd/silverbullet-mattermost) plug uses more complicated values for settings (using a single `mattermost` key with a mapping of server names to URLs etc.)

zefhemel:
Another idea here would be to “auto scope” settings by plug name. Perhaps `core` would be the exception (with settings listed at the top level), but e.g. the `markdown` plug could list all its settings under `markdown` in settings. For instance:

```yaml
markdown:
   renderer: default
```

Then we don’t have to worry about name clashes.

pihentagy:
Not digged into the code actually, so say if this won't work, but my thoughts are:

- require a zod scheme for all plugs
- parse the provided schema (not allowing extra fields), and provide the plug this (validated) object
- plugs cannot directly access the setting (there could be secrets anyway)

This way plugs declaratively state what settings they require.

You can combine this setup with autoscoping.

pihentagy:
There could be dependant plugs, so one may want to allow to dig in other plugs settings though...

MrMugame:
> * require a scheme for all plugs (use [zod](https://zod.dev/) or [valibot](https://github.com/fabian-hiller/valibot))

We would need some kind of serializable scheme, because we need to pass from a webworker. But I like the idea of silverbullet providing it, because every plug that uses settings is going to need some validation. No point in having 10 zod instances in 10 webworkers, that's just dead weight on the plug bundles.
[...]

PR #679, opened 2024-02-03
Plug: Emoji aliases (#246)

Maarrk:
Allows users to define emoji aliases in SETTINGS which are then used in hints after `:`. Some details we might want to discuss:

The settings are `emoji: aliases:`, instead of `emojiAliases:`, it feels a bit weird to require double indent, but it's putting everything under the plug's name (as hinted here https://github.com/silverbulletmd/silverbullet/issues/226#issuecomment-1362014066). And getting an existing emoji list would have its own key for URL 🤔 

If I understood correctly, the complete handler would be run on each keystroke, so the config is updated in the background, with an interval [like federation config](https://github.com/silverbulletmd/silverbullet/blob/04abc283b065fd5666bbe33b67206ef816303bfc/plugs/federation/config.ts#L12-L14)

### DIFF
--- a/web/types.ts
+++ b/web/types.ts
@@ -2,7 +2,6 @@ import { Manifest } from "../common/manifest.ts";
 import { PageMeta } from "$sb/types.ts";
 import { AppCommand } from "./hooks/command.ts";
 import { defaultSettings } from "../common/util.ts";
-import { EmojiConfig } from "../plugs/emoji/emoji.ts";
 
 // Used by FilterBox
 export type FilterOption = {
@@ -33,6 +32,10 @@ export type ActionButton = {
   description?: string;
   command: string;
   args?: any[];
+};
+
+export type EmojiConfig = {
+  aliases: string[][];
 };
 
 export type BuiltinSettings = {
--- a/web/types.ts
+++ b/web/types.ts
@@ -2,6 +2,7 @@ import { Manifest } from "../common/manifest.ts";
 import { PageMeta } from "$sb/types.ts";
 import { AppCommand } from "./hooks/command.ts";
 import { defaultSettings } from "../common/util.ts";
+import { EmojiConfig } from "../plugs/emoji/emoji.ts";
 
 // Used by FilterBox
 export type FilterOption = {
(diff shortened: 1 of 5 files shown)


zefhemel:
Nice! There's also this https://github.com/silverbulletmd/silverbullet/blob/f82d72e626a7b43ade8585db6ac968e18d01a5d3/web/types.ts#L37  Could you add it to `BuiltinSettings` as well?

Maarrk:
I updated the PR a bit more times than I wanted, but changes are cosmetic:

- Add the `emoji` key to `BuiltinSettings`
- Place the user-defined emojis before the default list in the picker
- Slightly change error message, since the newline seemed to be ignored

I had to change VS Code settings to actually use Deno for formatting files, even if the user has something else configured. Most changes in 7c28fee35c6f6277370bddc6b1b4f2480b567c9a were formatters disagreeing, and I couldn't just disable formatOnSave for a moment, since it's on in the workspace settings.

web/types.ts
```diff
@@ -2,6 +2,7 @@ import { Manifest } from "../common/manifest.ts";
 import { PageMeta } from "$sb/types.ts";
 import { AppCommand } from "./hooks/command.ts";
 import { defaultSettings } from "../common/util.ts";
+import { EmojiConfig } from "../plugs/emoji/emoji.ts";
```
zefhemel, line 5: Ok, this may seem tiny but the impact is big: turn this into a `import type` (rather than plain `import`) so that not the entire emoji JSON bundle gets imported into the client.js script in addition to the emoji plug.

zefhemel:
One small comment and then this is ready to merge IMO.

web/types.ts
zefhemel: Maybe ESBuild is smart enough to tree shake this out, but I prefer either an `import type` here or to move the EmojiConfig type into `web/types.ts` — which is also a reasonable solution.

zefhemel: Actually maybe move the type here so not to cross-import plug code into the client bundle

Maarrk: I was thinking where to put it, but didn't consider bundling at all. Thanks for the explanation, I'll try to change it later today.

zefhemel:
Ok just did that change for you, merging. Thank you!

Issue #639, opened 2024-01-20
Button to add file, image/camera


bilogic:
On mobiles, it would be wonderful if there was a button to directly attach file/image while typing markdown

zefhemel:
I can see that, my only concern is screen real-estate and minimalism. Right now there are 3 buttons (and soon there will be a 4th), is adding a 5th worth it? Not entirely convinced yet.

bilogic:
Thanks, how about this?

In mobile portrait, all buttons are hidden into a burger menu

In tablet/desktop, these are expanded for direct clicks

bilogic:
Hi,

I did some analysis:

1. `Mode` button, is there a use case where this needs to be toggled constantly? My view is no, so perhaps this can be hidden into a command
2. `Home` button, this kind of assumes we constantly need to edit the same page or jump back to home, to navigate elsewhere, is that the case?
3. `Open` button, definitely needed
4. `Run` button, definitely needed

From this, I think 1 and 2 can potentially be removed if there is no strong use case to support them.

zefhemel:
How about we make all top buttons configurable so you can pick icons and commands to run? I think that will be nice right?

bilogic:
Yes that would be great!

zefhemel:
#643

bilogic:
Just curious, the camera will still be added right? I tried to upload an mp4 file, but could not find it on my server.

zefhemel:
There's already an `Upload: File` command, which should give you the option to also select from the camera. That said, there is a 10MB file upload limit right now, which may not make this great for video...

bilogic:
ok, just tested and it works. Think I also found a bug, if a file upload fails, user should be informed. :)

jeroenev:
Is there an option to change the upload limit?

Issue #214, opened 2022-12-18
navigation controls for mobile PWA


strcrzy:
When using silverbullet as a Mobile PWA on iOS, users lose Safari navigation controls.

I believe the mobile PWA could benefit from back and forward buttons in the nav bar.
I have attached a crude design proposal (my suggestions in red).

![mobilepwa](https://user-images.githubusercontent.com/119740/208312070-7fde79f4-4086-42f6-8e95-d6ca0194e9e7.jpeg)

zefhemel:
Yep, we can do this. I should mention that the usual swipe from left or right border from the screen gestures to go backwards and forward work fine. That's what I use myself.

v411e:
It looks like on Android there is no such a swipe gesture.

Maarrk:
What if we make it possible to display `actionButtons` from SETTINGS conditionally based on mobile and/or PWA display?

Would also be helpful for adding files, or I'd hide my template picker on mobile.

Probably more feasible now, with Expression language

zefhemel:
Implemented by @Maarrk via custom action buttons.

PR #686, opened 2024-02-06
Improve mobile navigation through history

Maarrk:
This is my idea for solving #214 using `actionButtons` without cluttering the desktop client (also relevant for #639).

I was considering implementing some history internal to SilverBullet, but this seemed both overkill and wouldn't work with external links nicely. So I added a syscall for the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History). Conveniently, both `back()` and `forward()` are equivalent to `go(-1)` and `go(1)`.

I tested that mobile detection works on my Chrome and Firefox, Do we need to check browser compatibility more, since it's just an opt-in minor convenience?

### DIFF
--- a/plug-api/silverbullet-syscall/editor.ts
+++ b/plug-api/silverbullet-syscall/editor.ts
@@ -73,6 +73,15 @@ export function openUrl(url: string, existingWindow = false): Promise<void> {
   return syscall("editor.openUrl", url, existingWindow);
 }
 
+/**
+ * This is calling the `go()` method from the History Web API.
+ * @param delta Position in history to move to relative to the current page,
+ * where a negative value moves backwards, and positive forwards
+ */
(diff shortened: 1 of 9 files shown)


Maarrk:
Another way to approach this, that I gave up on, would be to have an Expression Language Function and make the SETTINGS object somehow templated on that? But that was too difficult for me to figure out, and would complicate the config, and mix space files with display, and would be a much larger design choice...

I think I'd just use it for the buttons anyway

zefhemel:
I like it. Tested in on the iOS simulator (works fine), Safari on Mac (works fine as in: no mobile buttons) and FF and Arc on mac. Let's merge this as is, we can iterate later. This is already pretty useful! I do need to fix the resizing of the top bar based on screen width though, will look at that in a bit.

Issue #675, opened 2024-02-03
Make fenced code blocks with arbitrary length backticks work


zefhemel:
Didn’t know this existed: https://fosstodon.org/@rauschma/111863880614883363

There are some things that break here though. Just creating this issue for myself as a reminder.

PR #694, opened 2024-02-12
Change fenced code regex to include tildes

### LINKED ISSUE #693: Issues with tilde-delimited code blocks
Expected result:
Fenced code block behaves the same way when delimited with backticks and tildes, since they are not distinguished in [CommonMark Spec](https://spec.commonmark.org/0.31.2/#fenced-code-blocks), which is identified in [Markdown page](https://silverbullet.md/Markdown) on the main instance.

I only found out about this syntax today, but it is really useful since we can share templates and space script using that.

Observed results:
1. A code block with tildes can be edited, and has correct syntax hints, but neither keyboard nor mouse can exit editing it
2. When refreshing the page contents stay empty

Reproduction rate:
5/5 on Chrome, including PWA, and Firefox

Appendix:

Error thrown for 1:
```
fenced_code.ts:24 Uncaught TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))
    at enter (fenced_code.ts:24:29)
    at Ub.iterate (common.mjs:2:3482)
    at fenced_code.ts:16:23
    at YR.update [as updateF] (util.ts:100:14)
    at Object.update (state.mjs:5:17052)
    at ts.computeSlot (state.mjs:5:25661)
    at jp (state.mjs:5:19840)
    at new ts (state.mjs:5:24855)
    at ts.applyTransaction (state.mjs:5:25631)
    at get state (state.mjs:5:21304)
```

Error thrown for 2:
~~~
TypeError: a.match(...) is null
    enter fenced_code.ts:24
    iterate common.mjs:2
    JV fenced_code.ts:16
    create util.ts:95
    create state.mjs:5
    create state.mjs:5
    create state.mjs:5
    jp state.mjs:5
    ts state.mjs:5
    create state.mjs:5
    pb editor_state.ts:62
    loadPage client.ts:1038
async.ts:99:12
    Ni async.ts:99

~~~

Maarrk:
This seems to fix #693

*Yes, it did take me longer to write the report than the PR, no it is not funny.*

### DIFF
--- a/plugs/query/complete.ts
+++ b/plugs/query/complete.ts
@@ -117,7 +117,7 @@ function attributeCompletionsToCMCompletion(
 }
 
 export async function languageComplete(completeEvent: CompleteEvent) {
-  const languagePrefix = /^```(\w*)$/.exec(
+  const languagePrefix = /^(?:```+|~~~+)(\w*)$/.exec(
     completeEvent.linePrefix,
   );
   if (!languagePrefix) {
--- a/web/client.ts
+++ b/web/client.ts
@@ -1077,12 +1077,13 @@ export class Client {
           const { text: stylesText } = await this.space.readPage(
             cleanPageRef(customStylesPage),
           );
-          const cssBlockRegex = /```css([^`]+)```/;
+          // Analogous to yamlSettingsRegex in settings.ts
+          const cssBlockRegex = /^(```+|~~~+)css\r?\n([\S\s]+)\1/m;
           const match = cssBlockRegex.exec(stylesText);
           if (!match) {
             return;
           }
-          accumulatedCSS.push(match[1]);
+          accumulatedCSS.push(match[2]);
         } catch (e: any) {
           console.error("Failed to load custom styles", e);
         }
--- a/common/settings.test.ts
+++ b/common/settings.test.ts
@@ -1,0 +1,61 @@
+import { assertEquals } from "$lib/test_deps.ts";
+import { parseYamlSettings } from "./settings.ts";
+
+Deno.test("Settings regex", () => {
+  const exampleSettings = {
+    foo: "bar",
+    żółć: "🟡", // make sure Unicode works
+  };
+
+  assertEquals(
+    parseYamlSettings(`
+The typical case would be like this
+
+\`\`\`yaml
+foo: bar
+żółć: 🟡
+\`\`\`
+`),
+    exampleSettings,
+  );
+
+  assertEquals(
+    parseYamlSettings(`
+Tilde delimiters or yml should also work
+
+~~~yml
+foo: bar
+żółć: 🟡
+~~~
+`),
+    exampleSettings,
+  );
+
+  assertEquals(
+    parseYamlSettings(`
+\`\`\`yaml-settings
+wrong info string
+\`\`\`
+
+~~~yaml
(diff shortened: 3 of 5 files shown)


zefhemel:
Nice. I can't check this now but will have a look later. Is this really the only place matching triple backticks specifically? I'd be surprised.

Maarrk:
Changing to draft, should also look at #675

Maarrk:
Luckily, Markdown parsing with codemirror already handled this correctly, it was only being stopped by the language name. I've tried multiple combinations and haven't been able to break it so far.

I went through all triple backticks in the repo, and everything other was either `.md` files, some as info for the user, or *writing* a yaml block.

A remaining limitation is that both SETTINGS and STYLES still need to use exactly three backticks, but I didn't find any reliable way to fix that with regex only.

Should we add explanation of all the possibilities to Markdown/Basics#Code? Maybe not worth complicating it? If options other than triple backticks aren't mentioned anywhere, I don't feel that SETTINGS and STYLES need this described, just show the working example. Commonmark says it really concisely (emphasis mine):

> sequence of **at least three** consecutive backtick characters (`) or tildes (~)

plugs/query/complete.ts
```diff
@@ -117,7 +117,7 @@ function attributeCompletionsToCMCompletion(
 }
 
 export async function languageComplete(completeEvent: CompleteEvent) {
-  const languagePrefix = /^```(\w*)$/.exec(
+  const languagePrefix = /^(?:```+|~~~+)(\w*)$/.exec(
```
terr-steak, line 120: https://regex101.com/r/QBVhmo/1 to help validate regex

web/client.ts
```diff
@@ -1077,6 +1077,7 @@ export class Client {
           const { text: stylesText } = await this.space.readPage(
             cleanPageRef(customStylesPage),
           );
+          // TODO: Doesn't support ~~~ or longer delimiters
           const cssBlockRegex = /```css([^`]+)```/;
```
terr-steak, line 1081: Why not use ``(?:```|~~~)css([^`\~]+)(?:```|~~~)`` as you did similarly elsewhere?

Maarrk: I couldn't get it up to spec in the following example:
`````
````yaml
this should get in
~~~
this should still get in
```yaml - this should not match since it's inside
```
this should also be in first match
````
Ok, now we're done
`````

The problem with what you wrote here is that you need the end marker to match specifically what you started with. And when I wrote it with capture group for the start and later reused it, still got fairly complicated failure modes, and extra matches inside.

I went for "If I can't make it 100% spec I'll keep it simple" but I just realised if we're only interested in **first** block that matches, we can ignore inner matches.

I want to be careful here, because I expect all kinds of weirdness in SETTINGS

Maarrk: Gaze upon the incantation to please the arcane Regis Exodium
`````
^(```+|~~~+)yaml([\S\s]+)\1
`````

It works better on regex101 than regexr, I think I should test it better, but won't have the PC for the rest of the weekend

Maarrk: Honestly, I'll just write proper tests for it at this point, already thought of an example that breaks the mighty one

zefhemel:
Thanks for this!

PR #709, opened 2024-02-14
Add go code block support

vkuznecovas:
Adds go language support for code blocks.

Great project but this is the thing that currently keeps me from using silverbullet 😃


zefhemel:
Oh my, how could we have come this far without this. Thanks!

PR #714, opened 2024-02-16
Add note about https requirements

PatrikStenmark:
I just spent time trying out SilverBullet by just running the Docker container without any reverse proxy or anything in front of it. I couldn't understand why the offline mode didn't work. Finally I found the output in the dev console explaining that the web worker didn't launch because I didn't use https. To avoid other people wasting time, I added a note about this in the docs for Synced mode and the Docker.md page.

### DIFF
--- a/website/Client Modes.md
+++ b/website/Client Modes.md
@@ -23,7 +23,7 @@ Disadvantages:
 
 # Synced mode
 $sync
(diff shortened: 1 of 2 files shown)


zefhemel:
I think this is listed elsewhere too, but clearly not clearly enough. Thanks!

PR #716, opened 2024-02-16
Add custom data field to template plug

gorootde:
This PR adds the possibility to optionally use additional custom data to instantiate templates. 

The additional parameter can be used to inject additional data, so plugs or space-scripts can render data that was collected from external datasources.

### DIFF
--- a/plugs/template/page.ts
+++ b/plugs/template/page.ts
@@ -90,6 +90,7 @@ async function instantiatePageTemplate(
   templateName: string,
   intoCurrentPage: string | undefined,
   askName: boolean,
+  customData: any = undefined,
 ) {
   const templateText = await space.readPage(templateName!);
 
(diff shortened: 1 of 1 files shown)


plugs/template/page.ts
```diff
@@ -90,6 +90,7 @@ async function instantiatePageTemplate(
   templateName: string,
   intoCurrentPage: string | undefined,
   askName: boolean,
+  customData: any = undefined,
```
terr-steak, line 93: Maybe this could be a KV[] or AttributeObject[] instead to leverage known interfaces that are in use in the application currently, I am new to the code so just a thought. 

We might also want to think about adding this to the log statement on ln 97 to help with debugging.

gorootde: Well neither of these would allow passing arbitrary data structures to the parameter. Thus it would reduce flexibility.
Custom data (e.g. when fetched from a remote datasource) would always need to be mapped to one of these structures. Since this mapping is already implicitly done in the template (by using the corresponding placeholders) anyway, I wouldn't force anybody to do this a second time.

Issue #706, opened 2024-02-14
Add Powershell Syntax Highlighting


CyberPenguin:
Please could Powershell syntax highlighting be added to codeblocks?

**Reproduce Issue**
1. Insert code block and after the first ``` enter either ps or powershell
2. No syntax highlighting provided

zefhemel:
Should be easy to do. There is a power shell highlighter in legacy modes: https://github.com/codemirror/legacy-modes/tree/main/mode

Which needs to imported here: https://github.com/silverbulletmd/silverbullet/blob/main/common/languages.ts

I can do this after my vacation, but happy to take a PR that adds this.

zefhemel:
On discord somebody asked for diff, may well add that one too while at it. And potentially more languages in that list.

zefhemel:
Great!

giovannipollo:
I think this issue can be closed, since it has been solved!

PR #718, opened 2024-02-17
Added support for new languages in the code environment (diff, powershell, perl, tcl, verilos, vhdl, docker, cmake)

giovannipollo:
Pull Request associated to #706


zefhemel:
Awesome, thanks!

Issue #102, opened 2022-10-23
File tree


zefhemel:
I've resisted this for a long time, but it may be useful to have a "traditional" file/folder tree on the left-hand side. Fully optional of course. 

It should support folding and collapsing of folders, and highlighting your currently open page. 

This is fully implementable as a plug, and should use the `showPanel("lhs", ...)` API.

vassudanagunta:
This would complicate how page links work. 

Given this Hierarchy:
```
- Project A
  - Principles
  - Parser
  - Renderer
- Project B
  - Principles
  - Schedule
  - Team
- Principles
```

How would *Project A/Parser* link to each of the three *Principles* pages? Perhaps:

- fully qualified path to link to any place in the tree, e.g. `[[/Project B/Principles]]`
- `[[Principles]]` would link to page in same folder only

Alternatively, `[[Principles]]` could resolve to the "nearest" match, e.g. first match in this order: in same folder, first one down the tree (if deep trees are supported), first one up the tree. Fully qualified paths must be used to jump across branches.

Alternatively, you could force page names to remain unique irrespective of path. Folders are merely for organization and do not define a hierarchical page namespace.

vassudanagunta:
Also it would be nice if, continuing my example above, a folder node could have its own content, much like index pages on a website.

zefhemel:
@vassudanagunta this already works perfectly fine now, and you can use folders right now simply by putting slashes in page names (it creates the hierarchy automatically). There's a discussion on how to handle page reference links here: #116 

Right now we always require "full" links, and there's just shortened to the last pieces by live preview. Try it :)

zefhemel:
> Also it would be nice if, continuing my example above, a folder node could have its own content, much like index pages on a website.

And this also works perfectly fine, just create a page named "People" and one named "People/Vas" and this is what you'll get.

Bonus hidden feature, when you're on the "People/Vas" page, go to the Page switcher and hit the spacebar, it will complete the page name to be "People/" so you can easily add another page in that folder.

orausch:
Made a basic plugin for this here: https://github.com/orausch/silverbullet-filetree

zefhemel:
@orausch just tried it out, nice start! I have been looking around for a solid file tree component, one where we could also support things like drag & drop for moving pages etc, did you custom build something here?

ffernn-dev:
+1 on this, I haven't looked at the codebase yet but I might take a crack at it when I have some time.

[...]

PR #720, opened 2024-02-17
Allow passing additional arguments to `renamePageCommand` and `renamePrefixCommand`

joekrill:
Adds the ability to send additional arguments to `renamePageCommand` and `renamePrefixCommand` of the `index` plug, as well as returning `true` or `false` from each command to indicate whether the action was successful.

The arguments are completely optional, and the existing behavior remains the default when the additional arguments aren't provided, so everything remains fully backward-compatible. 

## Why

I've created a [tree view component/plug](https://github.com/joekrill/silverbullet-treeview) and one of the desired features is to [allow dragging and dropping nodes to rename pages](https://github.com/silverbulletmd/silverbullet/issues/102#issuecomment-1696908355). That renaming functionality already exists via these two methods, but we need to be able to pass in some values that are currently hard-coded in those functions. We also need to know whether the operation succeeded or failed to determine if the tree should be updated, and - in some cases - whether the user should be redirected to the renamed page.

## `renamePageCommand` 

This command already allowed a `page` argument to be provided that specified the name to rename to. If this wasn't provided the user is prompted for a new name. The following additional arguments can now be passed:

- `oldPage` - This is the page to be renamed _from_. Previous behavior always used the current page (`editor.getCurrentPage()`), and this is still the case if `oldPage` is not provided. But in the case of the tree view, the user may be dragging  page other than the current one.
- `navigateThere` - Specifies whether, after the rename, the user should be navigated to the new page. Previously this was always `true`, because it was always the current page being renamed, so it made sense to navigate to the updated page. This still defaults to `true`, but can be override to prevent that behavior. If the user drags a node from the tree view that isn't the current page, we don't want to redirect them to that page. 

## `renamePrefixCommand`

This command is written in a way that always prompts the user for the prefixes to rename from and to, as well as showing a confirmation dialog before performing the rename. It did not have the ability to be given any arguments at all. This update allows the following arguments to be provided programmatically:

- `oldPrefix` - The prefix to rename from
- `newPrefix` - The prefix to rename to. 
- `disableConfirmation` - When true, disables showing the confirmation prompt before renaming.

### DIFF
--- a/plugs/index/refactor.ts
+++ b/plugs/index/refactor.ts
@@ -183,7 +183,7 @@ export async function renamePrefixCommand(cmdDef: any) {
   );
 
   if (
-    !(cmdDef.disableConfirmation ?? await editor.confirm(
+    cmdDef.disableConfirmation !== true && !(await editor.confirm(
       `This will affect ${allAffectedPages.length} pages. Are you sure?`,
     ))
   ) {
--- a/plugs/index/refactor.ts
+++ b/plugs/index/refactor.ts
@@ -150,14 +150,31 @@ async function renamePage(
   return updatedReferences;
 }
 
-export async function renamePrefixCommand() {
-  const oldPrefix = await editor.prompt("Prefix to rename:", "");
+/**
+ * Renames pages based on a prefix string.
+ * @param cmdDef Optional command arguments
+ * @param cmdDef.oldPrefix The prefix to rename from. If not provided the
+ *   user will be prompted to enter a prefix.
+ * @param cmdDef.newPrefix The prefix with which to replace the `oldPrefix`
+ *   value. If not provided the user will be prompted to enter a new prefix.
+ * @param cmdDef.disableConfirmation If false, the user will be prompted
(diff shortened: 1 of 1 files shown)


zefhemel:
This looks very nice, but you'll have to wait a bit until I merge it since I'm not near a laptop for the next week or so. Very excited about the tree plug!

joekrill:
> Edit: scratch that. I'll have a closer look now and get this in. I really like to unblock you. Your tree looks great.

Awesome, thanks!! 

I've actually already got drag-and-drop working, but realized I can't release that version yet because it's not going to work correctly with anyone running an older version of SilverBullet (it will actually do some very unexpected and unwanted things!). I thought I could just check the version, and enable drag/drop if > 0.7.1, and disable it otherwise. But I actually don't think there's currently a way for a plug to get the current system version, unless I'm missing something? So I just threw up another #721 that adds a `system.getVersion` syscall. Does that seem like the best way to deal with this sort of thing? Or is there a better way or some other functionality I'm missing?

zefhemel:
I think that's a sensible approach for the future yes

PR #726, opened 2024-02-19
Update README.md video to the new one from 2024/02/09

johnlunney:


zefhemel:
Thanks! The whole readme needs updating but I'll do that later.

Issue #722, opened 2024-02-17
Switch docker builds to official Deno image


zefhemel:
Since there are official arm64 builds at last: https://hub.docker.com/r/denoland/deno

PR #728, opened 2024-02-19
Use official deno image

AlotOfBlahaj:
#722


zefhemel:
Hah. That's all there is to it? Cool. I'll merge this once I'm able to test it.

zefhemel:
Got two equivalent PRs at the same time. Merged the earlier one already. Thanks for the effort though!

Issue #725, opened 2024-02-19
Make tables a queryable object


gorootde:
Since tables are somehow already a data structure, why not make them also queryable as a data object?

Proposal for the spec & docs in [`Objects.md`](https://github.com/silverbulletmd/silverbullet/blob/main/website/Objects.md):

> ## table
> Markdown table rows are indexed using the `table` tag, any additional tags can be added using [[Tags]] in any of its cells.
> | Title | Description Text |
> | --- | ----- |
> | This is some key | The value contains a #table-tag |
> | Another key | This time without a tag |
>
> ```query
> table-tag
> ```
This would render the result as

| ref | tag | page | pos | title | description_text | tags
| --- | --- | ---- | --- | --- | --- | ---- |
| someref@23 | table |  ExamplePage |  23 |  This is some key |  The value contains a #table-tag | table-tag |

zefhemel:
I love it. I was thinking of supporting this for CSV embedded in fenced code blocks already, but also doing this for markdown tables makes perfect sense.

gorootde:
Glad you like it. I am going to prepare a PR. Feel free to assign this issue to me.

zefhemel:
Implemented by... you!

PR #729, opened 2024-02-19
Add support for queryable markdown tables

gorootde:
Adds support to query tables via the `table` object.
I'll do a little bit more polish the code later. However, the functionality is ready and seems to be working for all edge cases I tested.

This implements #725

I'd love to have a better and more intuitive value for the `table` property. But markdown doesn't support naming tables. Any ideas how we may name a table?

### DIFF
--- a/plugs/index/index.plug.yaml
+++ b/plugs/index/index.plug.yaml
@@ -25,11 +25,11 @@ functions:
   objectSourceProvider:
     path: api.ts:objectSourceProvider
     events:
-    - query:*
+      - query:*
   discoverSources:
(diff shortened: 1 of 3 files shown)


gorootde:
I'll do a little bit more polish the code later. However, the functionality is ready and seems to be working for all edge cases I tested. What do you think?

zefhemel:
Cool stuff. Will review and test this later this week.

zefhemel:
Merging this, then will follow up with a quick change removing console.log and adding `table` as a builtin

PR #738, opened 2024-02-22
Add `fetch` to editor plug's `requiredPermissions`

joekrill:
The "Link: Unfurl" command always fails because it uses fetch to get the title of the patch, but the plug does not have `fetch` permission.


zefhemel:
Hah! That's why it stopped working.

PR #739, opened 2024-02-22
Correctly skip adding default shortcuts for overriden commands

### LINKED ISSUE #688: `Ctrl-Shift-t` for Template Picker active even after overwriting shortcut
I noticed both on Firefox and Chrome PWA (but not as a tab within Chrome browser) than even if i overwritten the Template Picker shortcut to `Ctrl-Alt-k` as seen below:

image

the default shortcut `Ctrl-Shift-t` keeps working and opens the Template Picker.

`Ctrl-Shift-t` is normally used in the browser to re-open the last closed tab.

**Expected behavior:**
- When inside SB and I press `Ctrl-shift-t` I would expect to get the last tab closed reopened

**Actual behavior:**
- SB opens the `Template Picker`

Expected behavior only happening when in Chrome browser, and inside SB as a tab (not a PWA)
Actual behavior happens in Firefox and Chrome PWA

### LINKED ISSUE #689: Keyboard shortcut query is not showing overwritten shortcuts (by the user)
Below you can see a screenshot of what the new keyboard shortcut query shows vs the ones that I have setup in my instance (Ex: `Outline` shortcuts)

image

joekrill:
A small bug was causing existing shortcuts to continue to work even when adding overrides to `SETTINGS` (the overrides worked as well).

Fixes #688

(Note this does _not_ fix #689 - looks like that's caused by something different)

@zefhemel I hope you don't mind me putting up these PRs! I thought I could help fix some of these small bugs.

### DIFF
--- a/web/editor_state.ts
+++ b/web/editor_state.ts
@@ -302,7 +302,7 @@ export function createCommandKeyBindings(client: Client): KeyBinding[] {
   for (const def of client.clientSystem.commandHook.editorCommands.values()) {
     if (def.command.key) {
       // If we've already overridden this command, skip it
-      if (overriddenCommands.has(def.command.key)) {
+      if (overriddenCommands.has(def.command.name)) {
         continue;
       }
       commandKeyBindings.push({


zefhemel:
Dude, I really appreciate the help!

PR #751, opened 2024-02-25
Fix mq.datastore test

Maks-s:
Hello, awesome software :+1: !

I couldn't get the test to work no matter what I tried. Commit 687d1f0 helped, it made the test pass 9 times out of 10

This PR makes it work 10/10, but github loves to blow up when someone changes an indent so here's a rundown of what this PR does
- Replace `sleep`  with [FakeTime](https://docs.deno.com/runtime/manual/basics/testing/mocking#faking-time)
- Use try/finally to remove tmpFile and close the db even when a test fails
- Wait for the queue to be empty before running the 'subscribe' assert (see line 54)

If I may also suggest, [BDD testing](https://docs.deno.com/runtime/manual/basics/testing/behavior_driven_development) could be nice to better track and fix failing tests


zefhemel:
Thanks for this!

Regarding BDD: I'm open to it. For me the current approach has been ok, but open to contributions!

PR #752, opened 2024-02-25
Relax hashtag pattern matching (Support Emoji)

sirlancelot:
Adjust the RegExp for matching tags in order to better support compatibility  with other PKM systems. Though not the exact matcher used in Obsidian (since  its source is closed), this matcher is modeled mostly after it and allows most  non-whitespace characters, including Emoji.

I use a task line like `- [ ] Gift Idea #🎁 #Person-Name` to denote gift ideas for friends and family. This patch allows SilverBullet to recognize the emoji.

### DIFF
--- a/common/markdown_parser/parser.ts
+++ b/common/markdown_parser/parser.ts
@@ -19,6 +19,8 @@ import { TaskList } from "./extended_task.ts";
 
 export const pageLinkRegex = /^\[\[([^\]\|]+)(\|([^\]]+))?\]\]/;
 
+export const tagRegex = /#[^\d\s!@#$%^&*(),.?":{}|<>\\][^\s!@#$%^&*(),.?":{}|<>\\]*/;
+
 const WikiLink: MarkdownConfig = {
   defineNodes: [
     { name: "WikiLink", style: ct.WikiLinkTag },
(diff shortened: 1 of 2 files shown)


sirlancelot:
The RegExp is two largely identical character classes `[]`, with the first one also excluding numbers. The effect is that tags _must_ start with a non-numerical character, followed by zero or more of any other character, including digits.

I wrote some tests here: https://regex101.com/r/6g27ZK/1

zefhemel:
Cool, but I'm pretty sure there are multiple other places, for sure in some auto complete functions where similar regexes are used. Could you do a quick codebase search for the likes of hashtagRegex or tagRegex and similar as well? Browny points if you manage to centralize it all in one place.

sirlancelot:
I found one other file that uses it in `web/`. I went ahead and consolidated the usage to `common/` and imported it within `web/`.

sirlancelot:
Hold up, it looks like the two RegExps are ever so slightly different. I can fix it in just a min.

sirlancelot:
Okay, it should be ready for you now! :)

zefhemel:
This seems to work as advertised, thanks! Merging :)

sirlancelot:
Thank you!

PR #754, opened 2024-02-25
Add type guards in query functions

Maks-s:
As the user can type whatever they want as arguments, this PR changes the query functions' parameters' types to `unknown`  and adds type guards to provide better exception handling.

Here's an example, using the following snippet
````
```template
{{replace(45, "4", "5")}}
```
````

**Without type guards**
uh

**With type guards**
nice


zefhemel:
Very nice, thank you!

PR #762, opened 2024-02-27
Change top bar to flexbox (take 2)

onespaceman:
With alignment regression fixed


zefhemel:
This seems to work. Thanks for the quick fix!

PR #770, opened 2024-03-01
improve dependency management

Maks-s:
This PR seeks to improve how dependencies are managed : 
- [x] use only import_map.json to list remote URLs
	- [x] remove `server/deps.ts`
	- [x] remove `common/deps.ts`
	- [x] remove `test_deps.ts`
	- [x] remove per-file imports
	- [x] remove `web/deps.ts`
- [x] `plug_api/` shouldn't contain imports using `import_map.json`, as it can be imported by remote plugs and should therefore use `deps.ts` (see [Import Maps are for Applications](https://docs.deno.com/runtime/manual/basics/import_maps#import-maps-are-for-applications))
- [x] `lib/` should also not contain `import_map.json` stuff, for the same reasons as the last entry
- [x] merge `lib/deps.ts` and `plug_api/deps.ts` to make only a single `deps.ts` file

If you feel this is unnecessary, feel free to decline this PR :+1:


zefhemel:
Yes, thank you for thinking about this and doing this work. I'll look at the details, but I like the general direction. Let me know when you think it's ready to merge.

zefhemel:
There was a merge conflict with main that I tried to resolve. 

When I run this in Firefox things seem to run fine, but for whatever in Chrome(like) browsers I see this:

CleanShot 2024-03-09 at 21 07 03@2x

I suspect that a weird import happens somewhere (in the service worker now?) that imports a Deno library that shouldn't be imported... Do you see this too?

Maks-s:
Thanks for the merge conflict fix

Shoots, I missed that! I'll fix this when I've got some time

zefhemel:
Ok, this seems to work now. Merging, thanks a lot for this effort!

PR #774, opened 2024-03-03
Clean replacement

onespaceman:
Most of clean.ts can be replaced with good ol' css.

web/active_lines.ts applies a style to active and selected lines, replacing all the isCursorInRange checks, and iterating over the syntax tree and css does the rest.
A change in behavior is that marks now are visible whenever its line is active, instead of within the element. I don't think I can easily change this behavior to the old way, so if it's unacceptable, feel free to reject the pr. I just need to know if it's worth continuing to work on this.

End goal:
- Command for toggling mark hiding on and off
- Marks can be set to turn translucent instead of hiding

Checklist
- [x] Text formatting marks
- [x] Links
- [x] Headings
- [x] Lists
- [x] Wikilinks
- [ ] Codeblocks
- [ ] Blockquotes
- [ ] Frontmatter
- [ ] Toggle command
- [ ] Figure out if configuring through styles is good enough or if it needs settings

### DIFF
--- a/web/active_lines.ts
+++ b/web/active_lines.ts
@@ -1,0 +1,60 @@
+import { 
+    Facet,
+    EditorView,
+    Extension,
+    Decoration,
+    RangeSetBuilder,
+    ViewPlugin,
+    DecorationSet,
+    ViewUpdate } from "./deps.ts";
+  
+    const baseTheme = EditorView.baseTheme({
+      "&light .cm-zebraStripe": {backgroundColor: "#d4fafa"},
+      "&dark .cm-zebraStripe": {backgroundColor: "#1a2727"}
+    })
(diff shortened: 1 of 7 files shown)


zefhemel:
This is very interesting. Let me play with this a little bit and see how it feels in practice.

zefhemel:
Let's have a discussion here: https://community.silverbullet.md/t/request-for-feedback-testing/288

Honestly after playing with this a few minutes I'm not sure I like the expansion of the entire paragraph. Generally I don't like that text jumps when moving text around, although this is inherent in how live preview works. This jumping happens more often when all it requires is putting a cursor _anywhere_ in the line, so... hmm.

Issue #783, opened 2024-03-04
Additional admonition types


SONDLecT:
Admonitions are a great way of calling out sections of notes and adding splashes of color to your notes. As great as **note** and **warning** are; it would be even _better_ if there were other types of admonitions available. e.g. tip, success, question, etc.

[Admonitions - Material for MkDocs 3](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) has a good list of types:
- note - General notes and additional information.
- abstract - Summary or overview of the content.
- info - Informational content or details.
- tip - Helpful tips or suggestions.
- success - Successful outcomes or achievements.
- question - Questions for consideration or discussion.
- warning - Potential hazards or critical warnings.
- failure - Failures or errors to avoid.
- danger - Dangerous situations or risks.
- bug - Presence of bugs or issues.
- example - Illustrative examples or demonstrations.
- quote - Notable quotations or excerpts.

zefhemel:
This should be fairly straightforward and should be easy for someone to pick up. Implementation here: https://github.com/silverbulletmd/silverbullet/blob/main/web/cm_plugins/admonition.ts

zefhemel:
@JosNun yeah I agree, but indeed this is why the solution in #786 is quite elegant.

paletochen:
It would be great if @onespaceman adds in the community examples of admonitions and how they would look like :)

SONDLecT:
Wow, this was quick! Awesome and thanks @onespaceman ! Echoing @paletochen , though, a few examples would be super helpful.

onespaceman:
I made a few examples and a How-to for now
https://gist.github.com/onespaceman/71c31edf4bed5d33d58eee1bacd127af
I'm working on #790 which should make it easy to share styles among each other

paletochen:
> I made a few examples and a How-to for now https://gist.github.com/onespaceman/71c31edf4bed5d33d58eee1bacd127af I'm working on #790 which should make it easy to share styles among each other

Thank you so much!
I went ahead and replicated all the admonitions mentioned by @SONDLecT in the original post (https://squidfunk.github.io/mkdocs-material/reference/admonitions/)

![image](https://github.com/silverbulletmd/silverbullet/assets/109062850/e8ec9d29-bb3c-47ea-a549-caab6eb62d2b)

The code for these are (Note and Warning are provided by the Core Plug)


```css

.sb-admonition[admonition="abstract"] {
[...]

PR #786, opened 2024-03-04
Custom Admonitions

onespaceman:
Allows making new admonition styles and adds support for the new github style `> [!note]`

See changes to website/Markdown/Admonitions.md for instructions

#783


zefhemel:
Thanks for this. This is cool. Makes me think we should introduce space styles in addition to space script so we can put these things (but also themes) into libraries. 🤔

zefhemel:
> I'll look into it 👍

Wrote up this issue: https://github.com/silverbulletmd/silverbullet/issues/790

If you're interested to look into it — great, otherwise I'll try to get to it at some later point.

PR #796, opened 2024-03-07
Implement space-style

onespaceman:
- add space-style blocks
- Also imports custom stylesheets from settings

For documentation, idk if this is now the official way to do user styles, with customStyles in settings remaining as a legacy option, or If you want to keep both official.


zefhemel:
This is very cool. I'll have a closer look at this later. Thanks for doing this!

web/client.ts
```diff
@@ -1095,36 +1096,32 @@ export class Client {
...
+      "space-style",
+      {},
+    );
+    if (!spaceStyles) {
+      return;
+    }
+
+    // Sort stylesheets (last declared styles take precedence)
+    // Order is 1: Imported styles, 2: Other styles, 3: customStyles from Settings
+    const sortOrder = ["library", "user", "settings"];
```
zefhemel, line 1109: Very clever that you thought about this, I like it.

plugs/index/style.ts
```diff
@@ -0,0 +1,63 @@
...
+}>;
+
+export async function indexSpaceStyle({ name, tree }: IndexTreeEvent) {
+  const allStyles: StyleObject[] = [];
+
+  // Also collect CSS from custom styles in settings
+  let customStylePages = await readSetting("customStyles", []);
+  if (!Array.isArray(customStylePages)) {
+    customStylePages = [customStylePages];
+  }
```
zefhemel, line 19: in `plug-api/lib/resolve.ts` there's a `cleanPageRef` function that removes the [[ and ]] if they're there (but also supports the format without [[ and ]]), it map be nice to `map` that function here, then the references lower down (e.g. line 43) get a bit cleaner too.

plugs/index/style.ts
```diff
@@ -0,0 +1,63 @@
...
+    const codeTextNode = findNodeOfType(t, "CodeText");
+    if (!codeTextNode) {
+      // Honestly, this shouldn't happen
+      return;
+    }
+    const codeText = codeTextNode.children![0].text!;
+    let codeOrigin = "";
+    if (customStylePages.includes("[[" + name + "]]")) {
+      codeOrigin = "settings";
+      console.log(codeOrigin);
```
zefhemel, line 45: This can probably go

plugs/index/style.ts
```diff
@@ -0,0 +1,63 @@
...
+    } else if (name.includes("Library")) {
+      codeOrigin = "library";
+      console.log(codeOrigin);
```
zefhemel, line 48: This can probably go

plugs/index/style.ts
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      codeOrigin = "user";
+      console.log(codeOrigin);
```
zefhemel, line 51: This can probably go

plugs/index/style.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (!codeTextNode) {
+      // Honestly, this shouldn't happen
+      return;
+    }
+    const codeText = codeTextNode.children![0].text!;
+    let codeOrigin = "";
+    if (customStylePages.includes("[[" + name + "]]")) {
+      codeOrigin = "settings";
+      console.log(codeOrigin);
+    } else if (name.includes("Library")) {
```
zefhemel, line 46: maybe `.startsWith("Library/")`?

plugs/index/style.ts
```diff
@@ -0,0 +1,63 @@
+import { IndexTreeEvent } from "../../plug-api/types.ts";
+import { collectNodesOfType, findNodeOfType } from "../../plug-api/lib/tree.ts";
+import { ObjectValue } from "../../plug-api/types.ts";
+import { indexObjects } from "./api.ts";
+import { readSetting } from "$sb/lib/settings_page.ts";
+
+export type StyleObject = ObjectValue<{
+  style: string;
+  origin: string;
+}>;
```
zefhemel, line 10: I just I realized I didn't do this for `space-script` yet either, but brownie points for adding this to the `plugs/index/builtins.ts` so you get nice auto complete for `space-style` and its attributes.

zefhemel (changes requested):

zefhemel:
As to documentation — feel free to include this part or not, happy to do it myself — but I think we can instantly consider `space-style` to be the successor to the customStyles thing, so I'd just remove all references to that option from the docs, and pretend we only have Space Styles now (with a `Space Style` page under `website/` documenting how to use it).

onespaceman:
All clear, captain.

zefhemel:
Tweaked a few things, but looking good. Thanks again for tackling this, merging!

PR #817, opened 2024-03-17
Fix space on end of page name creates two pages, one shadowed #615

### LINKED ISSUE #615: Space on end of page name creates two pages, one shadowed
**What happened**
1. Create a new page using the "Page: From Template" command with a space at the end of the name
image

2. See that two pages are created, but links always point to the page without space at the end
image

3. View on filesystem
image

**What I expected**
- Creating a new page using the "Page: From Template" command with a space at the end of the name should strip the space from the name before creating a page.
- Only the page without a space at the end should be created

fflorent:
## Context

Fixes #615.

## Proposed solution

1. Disallow PUT on filename with filenames surrounded with spaces;
2. If the filename provided on GET is surrounded by spaces, redirect to its trimmed version;
3. In case of required authentication, return a 401 (UNAUTHORIZED) + redirect;

## Other

If you want me to create tests, I would like to know if there already exists files places I can put them in.

### DIFF
--- a/common/spaces/http_space_primitives.ts
+++ b/common/spaces/http_space_primitives.ts
@@ -34,14 +34,18 @@ export class HttpSpacePrimitives implements SpacePrimitives {
         throw new Error("Offline");
       }
       if (result.redirected) {
-        // Got a redirect, we'll assume this is due to invalid credentials and redirecting to an auth page
-        console.log(
-          "Got a redirect via the API so will redirect to URL",
-          result.url,
-        );
-        alert("You are not authenticated, redirecting to login page...");
-        location.href = result.url;
-        throw new Error("Not authenticated");
+        if (result.status === 401) {
+          console.log(
+            "Received unauthorized status and got a redirect via the API so will redirect to URL",
+            result.url,
+          );
+          alert("You are not authenticated, redirecting to login page...");
+          location.href = result.url;
+          throw new Error("Not authenticated");
+        } else {
+          location.href = result.url;
+          throw new Error("Redirected");
+        }
       }
       if (result.status === 401) {
         location.reload();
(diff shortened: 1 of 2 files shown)


common/spaces/http_space_primitives.ts
```diff
@@ -34,14 +34,18 @@ export class HttpSpacePrimitives implements SpacePrimitives {
...
+        if (result.status === 401) {
+          console.log(
+            "Received unauthorized status and got a redirect via the API so will redirect to URL",
+            result.url,
+          );
+          alert("You are not authenticated, redirecting to login page...");
+          location.href = result.url;
+          throw new Error("Not authenticated");
+        } else {
+          location.href = result.url;
```
zefhemel, line 46: Previously it would show an alert in this case, any particular reason why it shouldn't anymore?

server/http_server.ts
```diff
@@ -469,6 +470,7 @@ export class HttpServer {
         const req = c.req;
         const name = req.param("path")!;
         const spaceServer = await this.ensureSpaceServer(req);
+        const mdExt = ".md";
```
zefhemel, line 473: If you introduce constants like this, I'd prefer them at the top level (top of the file, next to `authenticationExpirySeconds`) for instance.

zefhemel (changes requested):

common/spaces/http_space_primitives.ts
fflorent: The alert still appears when the reason of the redirection is a `401 - Authentication Required` (line 42) and the server code has been adapted below (in `server/http_server.ts:491`).

server/http_server.ts
fflorent: Done

common/spaces/http_space_primitives.ts
fflorent: @zefhemel Does this answer your question? Or you are still worried about that change?

zefhemel:
Sorry for the delay, this looks good. Thanks!

PR #828, opened 2024-03-23
Index scripts and styles in templates

onespaceman:
Index space-scripts and space-styles in templates, and removes them when rendering

### DIFF
--- a/plugs/index/index.plug.yaml
+++ b/plugs/index/index.plug.yaml
@@ -131,12 +131,14 @@ functions:
     path: script.ts:indexSpaceScript
     events:
       - page:index
+      - page:indexTemplate
 
     # Style
   indexSpaceStyle:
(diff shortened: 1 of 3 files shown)


zefhemel:
Right, I had something like this for a bit, but then reverted it. The reason is that I don't know what the desired behavior is. Does it make sense to remove space scripts from rendered templates? What one could expect to work is creating e.g. a page template that actually _generates_ a space script based on template variables. I don't immediately see use cases for this yet, but I'd expect that to work. What do you think?

onespaceman:
I was trying to make an widget, which needed a custom function to process the text to a table. It made sense to include the script in the template file instead of separately.
For the styles, it's for the future when we can add css classes to blocks. Makes sense to include it along with the widget it styles.
Put all these together and it's a mini plug system.

PR #831, opened 2024-03-26
New font and css fixes

onespaceman:
Fix tooltip colors in dark mode
Change font to [jetbrains mono](https://community.silverbullet.md/t/change-the-default-font/316)
Preview - styling fixes and follow editor font


zefhemel:
Thanks for doing this. Now I need some adjusting time. From the screenshots it looked good. Now that it's "real" and I can test it, I have to get used to this new font after looking at iA Writer's for 2 years 😆

zefhemel:
I've been using this for a few days or a week and... not in love to be honest. Maybe I'll revert this. Or somehow we make this an option. Maybe there's a way to serve font files from a space directly so it's easier for people to use their own non-default fonts 🤔

paletochen:
> I've been using this for a few days or a week and... not in love to be honest. Maybe I'll revert this. Or somehow we make this an option. Maybe there's a way to serve font files from a space directly so it's easier for people to use their own non-default fonts 🤔

Same here,
It is not the experience I was expecting. I have been using JetBrains Mono for years in my terminal but i cannot make it work (to my eyes) in SB.

I like the original one more

onespaceman:
I have no opinion on the change because I'm using my own font with space style
`@import url("https://github.com/iosevka-webfonts/iosevka-term/blob/main/iosevka-term.css");`
Google fonts can also be imported this way. Is that a good enough solution?

zefhemel:
I ended up reverting the font aspect of this in https://github.com/silverbulletmd/silverbullet/commit/82e431d35896c8c175ac12451212d315dd06b57b

Still appreciate the effort of doing this though! :)

PR #833, opened 2024-03-31
Support size for images

### LINKED ISSUE #493: Add "resize (scale) image" eg. ![](image.jpg | 100x100)

fflorent:
Fixes #493

### DIFF
--- a/common/markdown_parser/parser.ts
+++ b/common/markdown_parser/parser.ts
@@ -599,10 +599,10 @@ export const FrontMatter: MarkdownConfig = {
 
 const ImageWithSize: MarkdownConfig = {
   defineNodes: [
-    { name: "ImageWithSize" },
-    { name: "ImageWithSizeAlt" },
-    { name: "ImageWithSizeURL" },
-    { name: "ImageWithSizeSize" },
+    { name: "ImageWithSize", style: t.link, block: true },
+    { name: "ImageWithSizeAlt", style: t.link },
+    { name: "ImageWithSizeURL", style: t.url },
+    { name: "ImageWithSizeSize", style: t.meta },
   ],
   parseInline: [
     {
@@ -622,12 +622,14 @@ const ImageWithSize: MarkdownConfig = {
         const altPosEnd = altPosStart + altText.length;
         const urlPosStart = altPosEnd + 2;
         const urlPosEnd = urlPosStart + url.length;
-        const sizePosStart = fullMatch.indexOf(size, urlPosEnd);
+        const sizePosStart = fullMatch.indexOf(size, urlPosEnd - pos);
         const sizePosEnd = sizePosStart + size.length;
 
         return cx.addElement(
           cx.elt("ImageWithSize", pos, endPos, [
+            cx.elt("LinkMark", pos, altPosStart),
             cx.elt("ImageWithSizeAlt", altPosStart, altPosEnd),
+            cx.elt("LinkMark", altPosEnd, urlPosStart),
             cx.elt("ImageWithSizeURL", urlPosStart, urlPosEnd),
             cx.elt("ImageWithSizeSize", sizePosStart, sizePosEnd),
           ]),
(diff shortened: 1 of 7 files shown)


zefhemel:
Thanks for this, I’ll have a closer look soon. Bit busy, but you’re not forgotten :-)

zefhemel:
Is it ready to look at? It’s marked as draft.

fflorent:
> Is it ready to look at? It’s marked as draft.

Sure! Also it is a working version. I just have doubts:
  - as it is not standard markdown, I have not much opinion about the extension we should implement, though I wonder if there would not be some alternative we could implement and that could be a better fit (like a syntax that could be extended to allow more features, like [Kramdown does](https://kramdown.gettalong.org/syntax.html#images));
  - I am not sure when markdown_reader.ts is executed, is it related to the federation feature?

zefhemel:
Sorry for the delay on this. To answer your question on where markdown_render is used: this is used in a few places:

* Markdown preview (Ctrl-p)
* Markdown table rendering (when the cursor is not in a markdown table)
* Any markdown widget, this includes live templates and live queries

On the latter, an easy way to test this:

~~~
```template
Put your ![image](....) here
```
~~~

In your implementation, you don't currently inject any width and height during markdown_rendering, correct? So this would be something to add. Other than that this looks great!

plugs/markdown/markdown_render.ts
```diff
@@ -338,6 +338,22 @@ function render(
         },
         body: "",
       };
+    case "ImageWithSize": {
+      const alt = findNodeOfType(t, "ImageWithSizeAlt")!.children![0].text!;
+      const src = findNodeOfType(t, "ImageWithSizeURL")!.children![0].text!;
+      const dimensionsToParse = findNodeOfType(t, "ImageWithSizeSize")!.children![0].text!;
+      const [, width, widthUnit = "px", height, heightUnit = "px"] =
+        dimensionsToParse.match(/(\d*)(%)?x(\d*)(%)?/) ?? [];
```
fflorent, line 346: By chance, do you have any idea where this piece of code could be factorized? It's just a copy-paste from `InlineImageWidget#getDimensions()`

common/markdown_parser/parser.ts
```diff
@@ -20,6 +20,9 @@ export const pageLinkRegex = /^\[\[([^\]\|]+)(\|([^\]]+))?\]\]/;
 export const tagRegex =
   /#[^\d\s!@#$%^&*(),.?":{}|<>\\][^\s!@#$%^&*(),.?":{}|<>\\]*/;
 
+export const imageWithSizeRegex =
```
zefhemel, line 23: Since regexes are hard to read for regular people, it would be very helpful to add a comment with 1-2 examples that match this regex here I think

plugs/markdown/markdown_render.ts
```diff
@@ -33,7 +33,7 @@ function cleanTags(values: (Tag | null)[], cleanWhitespace = false): Tag[] {
   return result;
 }
 
-function preprocess(t: ParseTree, options: MarkdownRenderOptions = {}) {
+function preprocess(t: ParseTree) {
```
zefhemel, line 36: Thanks for removing this, I've been thinking to remove it every time I open this file but thought "nah, maybe we'll need it later". KILL ALL THE CODE :D

plugs/markdown/markdown_render.ts
zefhemel: Honestly in this case I'd stick with the duplication. I was thinking of moving it in with the parser code, but that way you pull in the entire markdown parser into a plug which isn't great.

zefhemel (changes requested):
One very minor request

common/markdown_parser/parser.ts
fflorent: Done!

zefhemel:
Awesome, thanks a lot!

PR #735, opened 2024-02-21
Add copy to clipboard button to code blocks

joekrill:
For any code blocks - except `template`, `query`, and `embed` - adds a "Copy" button to the upper-right hand corner (next to the language string), which copies the raw text:

Screenshot 2024-02-21 at 2 37 25 PM

Screenshot 2024-02-21 at 2 37 05 PM

Implements #345

PR #834, opened 2024-04-01
Remove dollar before commands in "shell" block

Maarrk:
With the blocks consistently having the `shell` specifier, the dollar sign is not needed to indicate these lines are terminal commands. There is also no need to differentiate user `$` and root `#` shell, since the lines contain `sudo` whenever root privileges are required.

This significantly improves UX considering #735

Credit goes to Russ Williams (rlw1138) for suggesting this https://community.silverbullet.md/t/copy-to-clipboard/351

---

Since the majority of command blocks used the `shell` info string, I changed the few `bash` ones to match (as a proud `zsh` user).


zefhemel:
Thanks 👍🏻

PR #837, opened 2024-04-02
Window controls overlay

s1gnate-sync:
Based on this article: https://learn.microsoft.com/en-us/microsoft-edge/progressive-web-apps-chromium/how-to/window-controls-overlay

Some minor eye-candy.

Screenshot 2024-04-02 15 39 16

### DIFF
--- a/web/manifest.json
+++ b/web/manifest.json
@@ -11,6 +11,7 @@
   "capture_links": "new-client",
   "start_url": "/#boot",
   "display": "standalone",
+  "display_override": ["window-controls-overlay"],
   "scope": "/",
   "theme_color": "#e1e1e1",
   "description": "Markdown as a platform"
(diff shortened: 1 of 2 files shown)


s1gnate-sync:
to make it even better it would be great if silverbullet can match os theme, it should be relatively easy to detect using

```
@media (prefers-color-scheme: dark) {
  :root {
    /* Dark mode variables go here */
  }
}
```

s1gnate-sync:
There is a room for improvement, but in require a bit more changes so I decided to split it in two. Since we have `titlebar-area-height` we can also make navigation bar and icons smaller so they would be the same height as native window decoration.

zefhemel:
Interesting. I’ll give this a try later. It looks a bit odd in your screenshot, but maybe I’m just not used to the Windows aesthetic.

s1gnate-sync:
@zefhemel it's chromeos :-)

zefhemel:
I don't have chromeos so I can't test it there, but this doesn't seem to break anything in Safari nor Chome. So I'll merge this. Thanks!

PR #840, opened 2024-04-05
Improve modals

daniel-michel:
I have noticed some small issues with the picker and other modals that I fixed / still try to fix:

- [x] The `MiniEditor` in the `Filter` component does not fill the full width
- [x] No styles for the buttons in modals (`Prompt` and `Confirm`) including CSS variables and dark theme
- [x] Buttons in modals can not be activated using enter
- [x] In Firefox when the results in shown in the `Filter` component is long enough to show a scroll bar, it can be focused using tab. I found this documented in the [Mozilla wiki](https://wiki.mozilla.org/XUL:Focus_Behaviour#:~:text=you%20can%20focus%20a%20scrollable%20HTML%20element%20%2D%2D%20one%20with%20overflow%20set%20to%20%27scroll%27%20or%20%27auto%27%20%2D%2D%20if%20you%20change%20the%20focus%20with%20the%20tab%20key%2C%20but%20not%20by%20using%20the%20mouse.)
- [x] When a `Prompt` or `Confirm` modal is shown, it can be ignored and the page can be edited (this could be considered fixed, the current behavior with this PR is that clicking outside the modal does not do anything
- [x] When for example using the Search Space command and the focus is on the `MiniEditor`, <kbd>tab</kbd> has to be pressed repeatedly to change focus to the buttons
- [ ] ~~When the focus is no longer on the `MiniEditor` key events are delegated to the main editor. While it is not possible to type into it, other things are possible (like moving the invisible cursor, indenting/outdenting, using formatting shortcuts, etc.) and <kbd>tab</kbd> cannot be used to move focus to the next button~~

Currently the modals also have a slightly darkened background, because that comes by default with the dialog element, but I can change that back if wanted.

Here is a comparison of how things look:
| Before | After |
| - | - |
| Screenshot_20240409_185009 | Screenshot_20240409_184709 |
| Screenshot_20240409_185047 | Screenshot_20240409_184747 |
| Screenshot_20240409_185126 | Screenshot_20240409_184246 |
| Screenshot_20240409_185205 | Screenshot_20240409_184326 |
| Screenshot_20240409_185226 | Screenshot_20240409_184542 |

I am open to changing things, if wanted. I also don't have any prior experience with react/preact.


daniel-michel:
I am almost done now.

I striked through one of the points in the original post because I think it would make more sense to take a look at it separately. The part with the delegation is no longer happening but keyboard shortcuts can still be used like <kbd>Ctrl</kbd>+<kbd>b</kbd> which is strange behavior, but some other shortcuts like <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>t</kbd> should maybe still work when, for example, the page picker is open?

And for the last other point that is still open in my original post I am not sure how to proceed: The issue that focus sticks to the text field when pressing tab. I figured out that it is caused [here](https://github.com/silverbulletmd/silverbullet/blob/main/web/components/mini_editor.tsx#L269) however I am not sure why this was done. I could just get rid of it if there are no other situations where it is required? I also noticed that while the line I linked to is run in both, firefox and chrome, it only causes the focus to actually stick to the text field in chrome, in firefox it does not appear to have any effect.

daniel-michel:
I noticed that my previous commit [Prevent key events from propagating outside of modals](https://github.com/silverbulletmd/silverbullet/pull/840/commits/f14373c15e3e1737b6208938c75d47392aa989ee) prevented navigating through the results with the arrow keys. I change the [mini editor to add the events directly to the div](https://github.com/silverbulletmd/silverbullet/pull/840/commits/c25b7923653481857b384823d649ee98ae2158ca) although I am again not sure why it was this way. I hope this is good?

zefhemel:
Thanks. I'll have a look at this over the next few days.

zefhemel:
Overall this looks good, but I found one issue (so far): when you use PgDown a few times to go down in a longer filtered list, the selection moves beyond the viewport and you can't seem to get out of it. Previously this worked. Not sure you can see this screen recording (you can't see the keyboard shortcuts unfortunately):

https://github.com/silverbulletmd/silverbullet/assets/59073/dd450b3a-2488-4cf1-9455-90e1845c55d7

zefhemel:
Tests failed because dependencies went out of date. I think this looks good, let me merge it and we'll fix any bugs from there. Thanks for this!

PR #841, opened 2024-04-06
Allow middle click to open link

daniel-michel:
Middle clicking on links just puts the cursor into the link. I set `contenteditable` to `false`, like on wiki-links to make middle click also work to open links.


zefhemel:
Nice!

PR #846, opened 2024-04-11
Configurable upload file size for drag and drop too

tmichela:
following #832 was not handling drag and drop of files.


zefhemel:
Nice!

PR #854, opened 2024-04-21
Provide a way to customize app name & desc

s1gnate-sync:
@zefhemel hey, here is my simplistic implementation of customizable `manifest.json`. What do you think? 

If you like it I will update documentation and cli usage message too.

### DIFF
--- a/cmd/server.ts
+++ b/cmd/server.ts
@@ -119,10 +119,25 @@ export async function serveCommand(
       new PrefixedKvPrimitives(baseKvPrimitives, ["*"]),
     );
   }
+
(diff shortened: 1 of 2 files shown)


zefhemel:
This looks reasonable. I think adding even more command line options doesn't make sense, I think configuration through environment variables is sufficient. 

Could you add the extra configuration options to the website/Install/Configuration.md file as well? Then I think we're ready to go!

Thanks!

s1gnate-sync:
@zefhemel done!

zefhemel:
Awesome, thank you!

PR #863, opened 2024-05-12
Fixes panel border espacially for dark mode

MrMugame:
So currently the border around panels is wrong in many cases
image
Especially in the `bhs` and `lhs` case
image
It's most noticeable in dark mode, because in white mode the border color is the same as the background (white)

This PR fixes both the position of the border and the color of the border in dark mode
image


zefhemel:
Perfect, thank you!

PR #876, opened 2024-06-04
Fixed greedy parsing of settings block

MrMugame:
This alters the regex for getting the settings block to not parse the content of the block greedily, because this could lead to problems where two consecutive blocks would be captured as one. [Example](https://regex101.com/r/4eekHd/1)


zefhemel:
Thanks!

PR #879, opened 2024-06-06
Add subscript and superscript

### LINKED ISSUE #872: Feature request: Markdown extension for superscript/subscript
When displaying a number with an exponent, it would be nice to be able to superscript the exponent without needing to pull in a latex renderer.

Syntax could be along the lines of `^superscript text here^` for superscript, unsure how subscript is typically exposed in markdown extensions.

MrMugame:
This fixes #872

The syntax would look like this, which seems to be the one used by e.g. pandoc. The alternative would be  Github<sup>TM</sup>s `H<sub>2</sub>O`, but I think pandoc style looks neater. 
image


zefhemel:
This is super awesome, thanks for doing this!

PR #882, opened 2024-06-07
Fixes the white flash for widget iframes & Make use of panel api

MrMugame:
So this does the same as the other PR just for the widget iframes now and I also noticed that the `panelHTML` already has an API for setting the theme of the message, so I made use of that (The widgets already do). This now allows plugs to directly to use the data-theme property for panels.


zefhemel:
Great :)

PR #887, opened 2024-06-12
Strip Markdown in TOC

### LINKED ISSUE #867: Table of Contents Link Rendering
Good morning,

I have set up several 'ToDo' headers on my Index page that reference various tagged tasks that I keep on a separate page. I have made the index headers link to the same header of the main Tasks page so I can quickly add or remove items to those categories but still have an at-a-glance view on the Index page. The Table of Contents at the top does not render the various header links in the 'pretty' way, but instead displays the Markdown code used to generate the internal link.
image
It'd be great if the Table of Contents could render this the same as the appearance of the header below.
image

MrMugame:
This strips the markdown in the TOC. So something like this `==FOO==` gets stripped to `FOO`. Most notably, this also strips aliased links (`[[index|Home]]` -> `Home`) and should thus fix #867.
The only problem right now is that a headline like this `# Hello ]` still results in ugly markdown for the same reasons as before and I frankly have no good ideas on how to fix this. (Btw. this whole "problem" also exists in [Obsidian](https://forum.obsidian.md/t/how-to-escape-square-bracket-in-link-custom-display-text/66826), but for them it's not that important as they don't have something like the toc plug, which programmatically renders to aliased links)
I tried to allow the escaping of the closing brackets in the regex, but this then results in issues there this `- [index| Foo \] Bar]` gets parsed as a task. Fixing that would temper with the syntax etc.
The only real other solution I can think of would be to just create the TOC in html, but that would destroy the whole look imo
Soooo ... just life with it for now?

### DIFF
--- a/plugs/index/toc.ts
+++ b/plugs/index/toc.ts
@@ -1,11 +1,7 @@
 import { editor, markdown, YAML } from "$sb/syscalls.ts";
-import { CodeWidgetContent } from "../../plug-api/types.ts";
-import {
-  findNodeOfType,
-  ParseTree,
-  renderToText,
-  traverseTree,
-} from "$sb/lib/tree.ts";
+import { CodeWidgetContent } from "$sb/types.ts";
+import { stripMarkdown } from "$sb/lib/markdown.ts";
+import { traverseTree } from "$sb/lib/tree.ts";
 
 type Header = {
   name: string;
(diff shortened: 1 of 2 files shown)


plugs/index/toc.ts
```diff
@@ -94,3 +103,98 @@ export async function widget(
     ],
   };
 }
+
+function stripMarkdown(
```
zefhemel, line 107: I think is a generally useful function. Could you do me a favor and move this to a new file `plug-api/lib/markdown.ts`? Then it's easier to import for other plugs in the future.

MrMugame: I extracted it and added all nodes (at least I think so, I haven't tested everything) (Realistically one could also use the HTML renderer from the preview plug and then convert that to `innerText`)

zefhemel:
I think this is good enough as is. Thanks!

PR #893, opened 2024-06-16
Redirects to initial url after authentication and cleans up auth code

### LINKED ISSUE #886: Login page doesn't redirect to requested page
Requesting a specific page URL redirects to `/.auth` if you aren't logged in, but after you log in it loads the index page instead of the requested page.
Not a big deal at all but might be a small improvement to redirect to the original page.

MrMugame:
Things this changes:
- This uses query parameters to redirect to the requested url after authentication. There are probably a 100 ways to implement this, hope this is ok. (Fixes #886)
- I also separated the logout into it's own route as I don't see a reason why it would be a query parameter.
- I also changed the `.all` and if else code into the proper honojs code and used a hono validator to validate the form data. in the auth route handler
- vscode also did some formatting, which changed the indenting of some blocks, which is why the diff looks kinda insane

### DIFF
--- a/plugs/editor/clean.ts
+++ b/plugs/editor/clean.ts
@@ -10,5 +10,5 @@ export async function cleanCommand() {
   }
   await editor.flashNotification("Now wiping all state and logging out...");
   await debug.cleanup();
-  await editor.openUrl("/.auth?logout", true);
+  await editor.openUrl("/.logout", true);
 }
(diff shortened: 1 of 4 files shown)


zefhemel:
Thanks for this. The reindent makes it a bit hard to check if no important logic changed, but I think we're good.

PR #897, opened 2024-06-22
Added automatic quote insertion

### LINKED ISSUE #896: Typing a " or ' when text is selected, should surround the text with these quotes
This is how it works with brackets, but not quotes yet.

MrMugame:
Fixes #896


zefhemel:
So the issue with this is that it doesn't follow the standard open quote and close quote logic, that is. When you select `sup` and hit `"` you get: `“sup“` instead of `“sup”`

zefhemel:
Nice!

PR #898, opened 2024-06-22
Makes the behaviour of logic expressions consistent in all cases

### LINKED ISSUE #894: null coealsce / "truthy or"
Usecase:
```template
{{#each {sometag}}}
* [[{{ displayName }}]]
{{/each}}
```

`displayName` is frequently `undefined`, so where it is undefined, i'd like to fall back to `name`.

As far as I can tell, the easiest way to achieve this today is `displayName: displayName : name` (if there's something else, let me know, and perhaps this is just an expression-language-page update). This mainly just an unfortunate pattern because it repeats `displayName` twice each time

What might be nice as an alternative is:
* `displayName or name` which i'm calling "truthy or". Today, this just returns `"true"`, but what i'm suggesting is essentially python/javascript style evalutation of an `or` by eagerly returning the first truthy value.
* `coalesce(displayName, name)` (or some syntax, like `|`) as a more explicit way of coalescing than the former option.

EDIT: yes i meant `displayName ? displayName : name` above

MrMugame:
This fixes #894.
I also removed the legacy call syntax handling, because it caused problems where an expression like this `a or b` was turned into `a (or b)`, which just doesn't make sense. Personally I'm for deprecating that call syntax as I don't really see a reason to keep it. If it should stay I would add it directly into the grammar to avoid any ambiguities like that.


zefhemel:
I'm ok with this. I added it to make the transition from handlebars easier, but that happened a while ago now so maybe it's time to drop it. 

I don't have access to a proper device for a while (vacation), so a few things are harder to check:

* Are there still any templates in one of the Libraries (core or journal) that rely on the legacy syntax? They'd need to be updated. 
* I'm pretty sure somewhere in the docs this legacy syntax is mentioned. That should be removed. 
* What happens when people have templates with the old syntax. How will it break? It would be nice if we can give some guidance on how to fix it. 
* Perhaps this will actually be a reason to move to 0.8.0 because it's a breaking change. Probably we should. It should be documented in the CHANGELOG properly too.

zefhemel:
So looks like you found some uses. Which means practically all people will have broken templates if they don't reimport the Core library. Really need a way a solution to library updates or rather lack thereof before merging this I fear.

MrMugame:
Actually we can partially keep the old syntax. I haven't found any occurences of the old call syntax in the library, where arguments were passed. Only without arguments and that is already not handled as a call.
```js
let attrVal = attributeVal[val[1]];
const func = functionMap[val[1]];
if (attrVal === undefined && func !== undefined) {
  // Fallback to function call, if one is defined with this name
  attrVal = func(variables);
}
return attrVal;
```
So we can keep that and only remove the syntax with arguments, which actually seems like a pretty good way

website/Library/Core.md
```diff
@@ -13,13 +13,13 @@ To import this library, run the {[Library: Import]} command in your SilverBullet
 # Included templates
 ```query
 template
-where name =~ /^{{escapeRegexp @page.name}}\//
+where name =~ /^{{escapeRegexp(@page.name)}}\//
```
zefhemel, line 16: But this is such a case where previously escapeRegex was called with an argument and no parentheses right?

MrMugame: But I think `Library/Core.md` and `Library/Journal.md` pages aren't imported. They are just part of the docs

zefhemel: Ah yes d'oh. You're right.

MrMugame:
Do we want to add a null coalesce operator (`??`) ?

zefhemel:
I don't have anything against it. Why not

PR #899, opened 2024-06-23
Fix double imports

MrMugame:
This removes any packages, which were imported twice, resulting in a bundle size decrease of about 28% (`2272KiB` to `1648KiB`)


MrMugame:
It really gets annoying when it's multiple levels deep, but now we are down to `1620KiB`. This also fixes `css` not working for code blocks, probably because of `instanceof` issues

zefhemel:
Awesome thank you!

PR #900, opened 2024-06-23
Fix `allKnownFiles` not being updated for `pageExists()`

MrMugame:
The mapping of the files to page names has to happen with every call to `pageExists()` as the underling array will change. Initially it's even empty and never updating it will cause the resulting mapped set to stay empty. This would mean `pageExists()` returnes `false` every time.

This issue was introduced with 1f94915 and is thus still pretty new, which is probably why it was unnoticed until now.


zefhemel:
Right. I think I noticed this once. Didn't have time to look into it immediately and then forgot. Nice catch!

PR #901, opened 2024-06-25
Hides the '\' from escaped characters

MrMugame:
Basically just hides the slash from escaped characters unless the text is selected
image


zefhemel:
Very nice

PR #903, opened 2024-06-26
Link Embeds

onespaceman:
right now page embeds work in editor view with support for sectional embeds with #headers @position or $anchor links.

Todo: embedding in Markdown preview

I'm not sure if this is the best way either, because I was also trying to add support for embedding other file types, like videos and pdfs. I was thinking to do that by reading mimetype of the file, but haven't figured out how to only fetch the file once, maybe merge the inline image widget and the embed codeblock into one all-purpose widget?


zefhemel:
I've forked this branch and will continue it. Thanks for the work so far, it's almost there!

PR #904, opened 2024-06-26
Change selection behaviour

MrMugame:
Ok, this is more of shot into the blind. It does a few things:
- It filters out all `select.pointer` events and fires an empty transaction on `mouseup`, which basically replicates obsidians behaviour for selections, meaning while you still hold down the mouse nothing happens and only when you let go, the view updates. This removes all the weird jitter/feed-back loops when selecting.
- You can now select part of a widgets code and part of the outside text and it will still render the widget code.
- Change all events for different kind of links to `click` as that is also the normal behaviour of `<a>`s. It just seems weird that sometimes stuff happens on `mouseup` and sometimes on `mousedown` (Maybe there is a reason, but I couldn't think of one). This also sometimes caused problems: When you for example select some text and end on a link with the cursor, that link triggered.
- Remove the `drawSelection()` extension. This maybe controversial and is definitely up for debate, but to me the default browser selection behaviour just seems more natural and we aren't using multi cursor/selection stuff anyways.

I'd love some feedback on this

### DIFF
--- a/web/cm_plugins/fenced_code.ts
+++ b/web/cm_plugins/fenced_code.ts
@@ -6,7 +6,6 @@ import {
   decoratorStateField,
   invisibleDecoration,
   isCursorInRange,
-  shouldRenderAsCode,
 } from "./util.ts";
 import { MarkdownWidget } from "./markdown_widget.ts";
 import { IFrameWidget } from "./iframe_widget.ts";
@@ -18,7 +17,7 @@ export function fencedCodePlugin(editor: Client) {
     syntaxTree(state).iterate({
       enter({ from, to, name, node }) {
         if (name === "FencedCode") {
-          if (shouldRenderAsCode(state, [from, to])) {
+          if (isCursorInRange(state, [from, to])) {
             // Don't render the widget if the cursor is inside the fenced code
             return;
           }
--- a/web/cm_plugins/util.ts
+++ b/web/cm_plugins/util.ts
@@ -24,10 +24,6 @@ export class LinkWidget extends WidgetType {
 
     // Mouse handling
     anchor.addEventListener("click", (e) => {
-      e.preventDefault();
-      e.stopPropagation();
-    });
-    anchor.addEventListener("mouseup", (e) => {
       if (e.button !== 0) {
         return;
       }
--- a/web/cm_plugins/util.ts
+++ b/web/cm_plugins/util.ts
@@ -89,10 +89,8 @@ export function decoratorStateField(
     },
 
     update(value: DecorationSet, tr: Transaction) {
-      // if (tr.docChanged || tr.selection) {
+      if (tr.isUserEvent("select.pointer")) return value;
       return stateToDecoratorMapper(tr.state);
-      // }
-      // return value;
     },
 
     provide: (f) => EditorView.decorations.from(f),
@@ -159,21 +157,6 @@ export function isCursorInRange(state: EditorState, range: [number, number]) {
   return state.selection.ranges.some((selection) =>
     checkRangeOverlap(range, [selection.from, selection.to])
   );
-}
-
-export function shouldRenderAsCode(
-  state: EditorState,
-  range: [number, number],
-) {
-  const mainSelection = state.selection.main;
-  // When the selection is empty, we need to check if the cursor is inside the fenced code
-  if (mainSelection.empty) {
-    return checkRangeOverlap(range, [mainSelection.from, mainSelection.to]);
-  } else {
-    // If the selection is encompassing the fenced code we render as code, or vice versa
-    return checkRangeSubset([mainSelection.from, mainSelection.to], range) ||
-      checkRangeSubset(range, [mainSelection.from, mainSelection.to]);
-  }
 }
 
 /**
--- a/web/cm_plugins/util.ts
+++ b/web/cm_plugins/util.ts
@@ -53,6 +53,13 @@ export class LinkWidget extends WidgetType {
     anchor.href = this.options.href || "#";
     return anchor;
   }
+
+  eq(other: WidgetType): boolean {
+    return other instanceof LinkWidget &&
+      this.options.text === other.options.text &&
+      this.options.href === other.options.href &&
+      this.options.title === other.options.title;
+  }
 }
 
 export class HtmlWidget extends WidgetType {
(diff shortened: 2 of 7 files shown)


zefhemel:
Oh this is very interesting! Will require thorough testing though, so will get into this when I'm back

zefhemel:
Starting to test this.

First thing I noticed broken: Alt-clicking on a link (e.g. wiki link) seems broken now, this should move the cursor there without navigating. It seems to defocus the editor now.

zefhemel:
Also: clicking a wiki link only seems to register the second click. You basically have to click the editor (or link) twice to navigate there (on macOS with Vivaldi, but I assume this is the same in any desktop browser). I remember having this issue, which is when I started to use a bunch of the "mouseup" chenanigans.

MrMugame:
This [issue](https://discuss.codemirror.net/t/link-widget-is-clicked-only-if-editor-is-out-of-focus/7433) is described here. Interestingly enough, this worked in firefox before, they must be handling clicks in a different order.

MrMugame:
The defocusing issue seems to be unrelated to my PR, as I'm also experiencing it in my instance, which is running `0.7.7`. It doesn't happen in firefox, so that's also weird. I cannot really pinpoint the origin (Does clicking on widgets just steal the focus, maybe?), but it seems to be fixed if we just forcefully give back the focus.

zefhemel:
Ok, this is much better. I'm liking this. I only see two more minor issues:

1. When selecting widget text (so e.g. when a live query is rendered) the selection goes away once I stop holding the mouse button. In principle I can still select & copy if I copy while holding the mouse, but ideally the selection would remain.
2. The blinking cursor disappears when text is selected (also with the keyboard). I suspect this may be a side effect of removing drawSelection perhaps?

MrMugame:
1. Ok, I'm able to reproduce this in a minimal example [here](https://codemirror.net/try/?c=aW1wb3J0IHsgbWluaW1hbFNldHVwLCBFZGl0b3JWaWV3IH0gZnJvbSAiY29kZW1pcnJvciIKaW1wb3J0IHsgV2lkZ2V0VHlwZSwgRGVjb3JhdGlvbiB9IGZyb20gIkBjb2RlbWlycm9yL3ZpZXciOwppbXBvcnQgeyBTdGF0ZUZpZWxkIH0gZnJvbSAiQGNvZGVtaXJyb3Ivc3RhdGUiOwppbXBvcnQgeyBoaWdobGlnaHRTcGVjaWFsQ2hhcnMsIGRyYXdTZWxlY3Rpb24sIGtleW1hcCB9IGZyb20gJ0Bjb2RlbWlycm9yL3ZpZXcnOwppbXBvcnQgeyBoaXN0b3J5LCBkZWZhdWx0S2V5bWFwLCBoaXN0b3J5S2V5bWFwIH0gZnJvbSAnQGNvZGVtaXJyb3IvY29tbWFuZHMnOwoKY29uc3QgdmlldyA9IG5ldyBFZGl0b3JWaWV3KHsKICBkb2M6ICJIZWxsbyBXb3JsZCIsCiAgZXh0ZW5zaW9uczogWwogICAgaGlnaGxpZ2h0U3BlY2lhbENoYXJzKCksCiAgICBoaXN0b3J5KCksCiAgICAvLyBkcmF3U2VsZWN0aW9uKCksCiAgICBrZXltYXAub2YoWwogICAgICAgIC4uLmRlZmF1bHRLZXltYXAsCiAgICAgICAgLi4uaGlzdG9yeUtleW1hcCwKICAgIF0pLAogICAgU3RhdGVGaWVsZC5kZWZpbmUoewogICAgICBjcmVhdGUoKSB7CiAgICAgICAgcmV0dXJuIERlY29yYXRpb24ubm9uZQogICAgICB9LAogICAgICB1cGRhdGUoXywgdHIpIHsgCiAgICAgICAgY29uc3QgZGVjbyA9IERlY29yYXRpb24ud2lkZ2V0KHsKICAgICAgICAgIHdpZGdldDogbmV3IFRleHRib3hXaWRnZXQoIlRoaXMgaXMgc29tZSB0ZXh0LCB5b3Ugc2hvdWxkIGhpZ2hsaWdodCB0byBzZWUgdGhlIHByb2JsZW0iKQogICAgICAgIH0pLnJhbmdlKHRyLnN0YXRlLmRvYy5sZW5ndGgpOwoKICAgICAgICByZXR1cm4gRGVjb3JhdGlvbi5zZXQoW2RlY29dKTsKICAgICAgfSwKICAgICAgcHJvdmlkZTogZiA9PiBFZGl0b3JWaWV3LmRlY29yYXRpb25zLmZyb20oZikKICAgIH0pLAogIF0sCiAgcGFyZW50OiBkb2N1bWVudC5ib2R5Cn0pCgp3aW5kb3cuYWRkRXZlbnRMaXN0ZW5lcigibW91c2V1cCIsICgpID0+IHsKICB2aWV3LmRpc3BhdGNoKHt9KTsKfSk7CgpjbGFzcyBUZXh0Ym94V2lkZ2V0IGV4dGVuZHMgV2lkZ2V0VHlwZSB7CiAgY29uc3RydWN0b3IodGV4dCkgewogICAgc3VwZXIoKTsKICAgIHRoaXMudGV4dCA9IHRleHQ7CiAgfQoKICBlcShvdGhlcikgewogICAgcmV0dXJuIG90aGVyLnRleHQgPT09IHRoaXMudGV4dDsKICB9CgogIHRvRE9NKCkgewogICAgbGV0IHAgPSBkb2N1bWVudC5jcmVhdGVFbGVtZW50KCJwIikKICAgIHAuaW5uZXJUZXh0ID0gdGhpcy50ZXh0OwogICAgcC5zdHlsZS5tYXJnaW4gPSAiMHB4IjsKICAgIHJldHVybiBwOwogIH0KfQo=). The culprit seems to be the transaction. Only after firing it, is the selection reset to it's "before state". This leads me to think that it's some kind of internal codemirror thing, I don't fully understand yet as my codemirror knowledge kind of ends here.
2. Yeah, it shouldn't be too hard to implement a custom cursor view plugin.

zefhemel:
For the record, I don't see (1) as a blocker for merging this, but it's sub optimal. The cursor thing is a bigger issue.

zefhemel:
I'll merge this as is also to allow me to test it on more devices and let other people find issues with it. Thanks a lot!

zefhemel:
@MrMugame sadly on iOS the cursor is now completely invisible. Do you have a way to look into this? Otherwise I'll revert.

MrMugame:
Ok I reverted back to `drawSelection()` (#917). I could fix it but the code would get more convoluted by the second and it's probably just not worth anymore.

PR #905, opened 2024-06-27
Initial attempt to fix #560

### LINKED ISSUE #560: Assign emoji to specific tags, or at a per-page basis
These emoji would then appear:
* In the page title (top bar)
* In page links

e.g. let’s say we assign 📕 to the `#book` tag. A page “Harry Potter” tagged with `#book` would then appear as e.g. “📕 Harry Potter”

How to configure?

* Perhaps a `icon` key in front matter for per-page configuration
* Per tag… TBD

deepkn:
Enables decorations for user specified tags - currently supporting only a single decoration - a 'prefix' to be added to the page name. Prefix is handled in the top bar title, page navigator as well as links.

Since visually identifying my defined object types was somewhat important for me, I took a stab at #560 - Quite unfamiliar with typescript and in general PWAs - so there could be a lot that can be improved and even better ways to do this. Open to suggestions.

TODO: Handle this in queries as well, at least when rendered via templates - but I'm wondering if that could be achieved better by simply extracting out the prefix as an attribute and using that (leaving it to the user)

### DIFF
--- a/server/http_server.ts
+++ b/server/http_server.ts
@@ -489,7 +489,7 @@ export class HttpServer {
       }
     });
 
-    const filePathRegex = "/:path{[^!].*\\.[a-zA-Z]+}";
+    const filePathRegex = "/:path{[^!].*\\.[a-zA-Z0-9]+}";
     const mdExt = ".md";
 
     this.app.get(filePathRegex, async (c) => {
--- a/web/service_worker.ts
+++ b/web/service_worker.ts
@@ -103,7 +103,7 @@ self.addEventListener("fetch", (event: any) => {
         pathname === "/index.json"
       ) {
         return fetch(request);
-      } else if (/\/.+\.[a-zA-Z]+$/.test(pathname)) {
+      } else if (/\/.+\.[a-zA-Z0-9]+$/.test(pathname)) {
         // If this is a /*.* request, this can either be a plug worker load or an attachment load
         return handleLocalFileRequest(request, pathname);
       } else {


zefhemel:
This looks promising. Definitely a good starting point to a more elaborate implementation. I'll have a closer look soon!

zefhemel:
A few issues I've found:

* I'm not sure the prefix should actually become part of the page name in the top bar. I think it should be put in front of the editor component somehow so you cannot edit it out. Right now, you can simply remove the prefix there, which is a bit weird.
* The prefix is only added in the top bar page title when you navigate from another page to it, not when the page is loaded refresh (reload the page and you'll see what I mean)
* The prefix is only rendered for wiki links when they're out of the selection scope. I think may be ok, although it would be slightly nicer if it would also be there in the code complete as well as when the cursor is inside the link:

CleanShot 2024-07-03 at 07 48 59

Some of these issues may be harder to get done. If you get stuck or you'd like me to take over let me know. This is a good start, and this is definitely something I'd like to get in soon, so happy to help!

deepkn:
Thanks Zef. Happy to take another shot at it - after all I'm loving SilverBullet a bit too much :)
1. That makes sense - I'm guessing that would have to go as another component for prefix before the MiniEditor; do correct me if I'm completely off here.
2. I did see this happening last day - I mostly never leave keyboard while navigating(one of the things I love about sb) so that missed my eye. I think I know what's happening there - infact I think the very first commit would not have had this problem. Let me think of a cleaner fix.
3. Sounds like a little bit more trickier to do - atleast based on what I have explored so far, so any pointers welcome. But I'll still take a shot at it in the coming days.

zefhemel:
Let me know when I can have another look!

deepkn:
Hey Zef, I think you can take another look. 1) and 2) from before must be fixed now and for 3) enabling the prefix in code complete turned out to be easier than I thought since codemirror autocomplete allows for a displayName override. What the latest patch doesn't have is the display of prefix when the cursor is inside the link.

zefhemel:
This is getting close! However I see some issues still especially on first load. See this GIF that I kick off with a reload. I hope that my mouse gesturing makes clear what I'm pointing at 😉 

CleanShot 2024-07-07 at 12 17 55

deepkn-ntnx:
Ah... dang - I should've seen that coming.!! Because reload clears everything and we need to wait for index to be built to get the tags to which the corresponding prefix was to be applied, I was updating that when the page list update is done, which happens async in the init sequence. It was not very apparent for me since I have a smaller number of pages currently - but I can guess what's happening on your end :) Since we do not want to block the app load on this, I guess this will always be async and there will be some delay - since the title is readily available - but the prefix is not until index is done atleast for the page in question. Few options that we can try from the top of my head:

1. await and query for the current page and update, instead of the entire pages to be indexed - but in the worst case, it'll look the same.
2. use browser's cache to cache current decorators for the page - but hard reload would see the same thing happening.
3. there is a neat little way around this which I'm obviously not seeing - I'm too new to this ;)

Would like your inputs on this one.

On displaying the prefix when cursor is inside a link - I have a way to do that, but I'm wondering if we'd end up with the same weirdness that you found in the topbar - since it's in the editor, that'll remain editable and the user could just remove it from the link if he chooses to - and it may come back to life if he moves the cursor out and back in again. I don't think we can make that non-editable within the editor component. What do you think?

deepkn:
Ah... dang - I should've seen that coming.!! Because reload clears everything and we need to wait for index to be built to get the tags to which the corresponding prefix was to be applied, I was updating that when the page list update is done, which happens async in the init sequence. It was not very apparent for me since I have a smaller number of pages currently - but I can guess what's happening on your end :) Since we do not want to block the app load on this, I guess this will always be async and there will be some delay - since the title is readily available - but the prefix is not until index is done atleast for the page in question. Few options that we can try from the top of my head:

1. await and query for the current page and update, instead of the entire pages to be indexed - but in the worst case, it'll look the same.
2. use browser's cache to cache current decorators for the page - but hard reload would see the same thing happening.
3. there is a neat little way around this which I'm obviously not seeing - I'm too new to this ;)

Would like your inputs on this one.

On displaying the prefix when cursor is inside a link - I have a way to do that, but I'm wondering if we'd end up with the same weirdness that you found in the topbar - since it's in the editor, that'll remain editable and the user could just remove it from the link if he chooses to - and it may come back to life if he moves the cursor out and back in again. I don't think we can make that non-editable within the editor component. What do you think?

plugs/editor/complete.ts
```diff
@@ -7,9 +7,15 @@ import {
...
 import { queryObjects } from "../index/plug_api.ts";
 import { folderName } from "$sb/lib/resolve.ts";
+import { readSetting } from "$sb/lib/settings_page.ts";
+import type { Decoration } from "$lib/web.ts";
+
+let decorations: Decoration[] = [];
 
 // Completion
 export async function pageComplete(completeEvent: CompleteEvent) {
+  updateDecoratorConfig();
```
zefhemel, line 17: Maybe `await` this so that you can be sure `decorations` will have a value from the first invocation (and you catch errors, which otherwise would go into the void).

plugs/editor/complete.ts
```diff
@@ -79,10 +85,19 @@ export async function pageComplete(completeEvent: CompleteEvent) {
     from: completeEvent.pos - match[1].length,
     options: allPages.map((pageMeta) => {
       const completions: any[] = [];
+      let namePrefix = "";
+      const decor = decorations?.filter(d => pageMeta.tags?.some((t: any) => d.tag === t));
```
zefhemel, line 89: `docorations` will always have a value, right? So no need for the `?` after it?

plugs/editor/complete.ts
```diff
@@ -79,10 +85,19 @@ export async function pageComplete(completeEvent: CompleteEvent) {
...
+      if (decor === undefined || decor.length == 0) {
```
zefhemel, line 90: And without the `?` from my previous comment, this can just become a `.length` check only probably.

plugs/editor/complete.ts
```diff
@@ -79,10 +85,19 @@ export async function pageComplete(completeEvent: CompleteEvent) {
...
+        namePrefix = "";
+      } else {
+        namePrefix = decor[0].prefix;
+      }
       if (isWikilink) {
         if (pageMeta.displayName) {
+          const decoratedName = namePrefix + pageMeta.displayName;
           completions.push({
             label: `${pageMeta.displayName}`,
+            displayLabel: `${decoratedName}`,
```
zefhemel, line 100: How about just `displayLabel: decoratedName,`

plugs/editor/complete.ts
```diff
@@ -93,8 +108,10 @@ export async function pageComplete(completeEvent: CompleteEvent) {
         }
         if (Array.isArray(pageMeta.aliases)) {
           for (const alias of pageMeta.aliases) {
+            const decoratedName = namePrefix + alias;
             completions.push({
               label: `${alias}`,
+              displayLabel: `${decoratedName}`,
```
zefhemel, line 114: Same as above.

plugs/editor/complete.ts
```diff
@@ -104,8 +121,10 @@ export async function pageComplete(completeEvent: CompleteEvent) {
             });
           }
         }
+        const decoratedName = namePrefix + pageMeta.name;
         completions.push({
           label: `${pageMeta.name}`,
+          displayLabel: `${decoratedName}`,
```
zefhemel, line 127: Same again

plugs/editor/complete.ts
zefhemel: And maybe use `.find` instead of `.filter`?

plugs/markdown/markdown_render.ts
```diff
@@ -554,25 +555,34 @@ function traverseTag(
...
         t.attrs!.src = options.translateUrls!(t.attrs!.src!, "image");
       }
 
       if (t.name === "a" && t.attrs!.href) {
-        t.attrs!.href = options.translateUrls!(t.attrs!.href, "link");
+        if (options.translateUrls) {
+          t.attrs!.href = options.translateUrls!(t.attrs!.href, "link");
+        }
+        if (t.attrs!["data-ref"]?.length) {
+          const pageMeta = allPages.filter(p => t.attrs!["data-ref"]!.startsWith(p.name));
```
zefhemel, line 576: Maybe `.find` instead of `.filter`?

web/cm_plugins/wiki_link.ts
```diff
@@ -67,9 +67,9 @@ export function cleanWikiLinkPlugin(client: Client) {
           }
           return;
         }
-
+        const pageMeta = client.ui.viewState.allPages.filter(p => p.name == url);
```
zefhemel, line 70: -> `.find` again

web/reducer.ts
```diff
@@ -20,6 +22,13 @@ export default function reducer(
       };
     case "page-loaded": {
       const mouseDetected = window.matchMedia("(any-pointer:fine)").matches;
+      const pageMeta = state.allPages.filter(p => p.name == action.meta.name);
```
zefhemel, line 25: `.find`

web/reducer.ts
```diff
@@ -20,6 +22,13 @@ export default function reducer(
...
+      const decor = state.settings.decorations?.filter(d => pageMeta[0]?.tags?.some(t => d.tag === t));
```
zefhemel, line 26: `.find`

web/reducer.ts
```diff
@@ -58,16 +67,31 @@ export default function reducer(
...
+      if (currPageMeta === undefined) {
+        currPageMeta = {} as PageMeta;
+      }
       for (const pageMeta of action.allPages) {
         const oldPageMetaItem = oldPageMeta.get(pageMeta.name);
         if (oldPageMetaItem && oldPageMetaItem.lastOpened) {
           pageMeta.lastOpened = oldPageMetaItem.lastOpened;
         }
+        let namePrefix = "";
+        const decor = state.settings.decorations?.filter(d => pageMeta.tags?.some((t: any) => d.tag === t));
```
zefhemel, line 80: `.find`

zefhemel (changes requested):

zefhemel:
In addition to the comments before, another ask:

In the near future I'd be interested in adding additional decorator things to pages, including e.g. colors or background colors (in addition to name prefixes). Therefore it seems to be useful to not specifically attach the `.namePrefix` to `PageMeta`, but just add `decoration` property (of type `Decoration`) to it, so that other properties (in addition to `prefix`) can be read from it in the future?

zefhemel:
The delayed showing of prefix in the top bar is acceptable I think, hard to fix that in a way that wouldn't slow a lot of things down. Not showing the emoji in the "cursor is in the link" state is actually desirable now that I think about it, it's consistent with "this shows what the ACTUAL underlying markdown is" philosophy, so that should stay that way.

However, not showing the prefix in the page wiki link until you move your cursor around is less idea. Basically what needs to happen is that whenever a `update-page-list` reducer is action is triggered, the editor updates (rerenders). I think you can do this with an empty `editorView.dispatch({})`. Have a look if you can figure that out. If not, address the issues in my comments and I'll merge it and try to fix that little issue myself.

Awesome work so far! Very eager to use this and take it further.

plugs/editor/complete.ts
zefhemel: And when switching to `.find` you can just check of not undefined I suppose.

deepkn:
> In addition to the comments before, another ask:
> 
> In the near future I'd be interested in adding additional decorator things to pages, including e.g. colors or background colors (in addition to name prefixes). Therefore it seems to be useful to not specifically attach the `.namePrefix` to `PageMeta`, but just add the full blown `decoration` property (of type `Decoration`) to it, so that other properties (in addition to `prefix`) can be read from it in the future. Would that be ok?

This makes sense - the only question I had here is the decorations could be, atleast in theory, a merge of different decorators defined for different tags applied to a page. Does it make sense to keep it merged per page? Maybe take a look at the latest update and see if that makes sense to you?

zefhemel:
I think it’s ok as is. If you can merge with current `main` and resolve the conflict I’ll give it another test and we can merge it and iterate from there. Thanks for this, looking forward to try and add to this!

PR #906, opened 2024-06-28
Support arm64 arch for Linux

fflorent:
Since deno version 1.1.41 has been released, [ARM64 architecture for Linux is supported](https://deno.com/blog/v1.41#linux-arm64-support).

This PR proposes to support and provide a binary for this architecture.


zefhemel:
Nice!

PR #912, opened 2024-07-02
Fix up validatePageName routine

nvkv:
File extensions can contain numbers too.
My main motivation was to enable myself hosting `woff2` files right from my space, like I do for `ttf` files.

Also added some tests.


zefhemel:
Thanks!

PR #917, opened 2024-07-03
Add `drawSelection()`

MrMugame:


MrMugame:
The extension being layered on ios is behaviour that was there before right?
image

zefhemel:
Ok. Things are back to normal on iOS with this. Thanks!

PR #921, opened 2024-07-06
Follow-up for parsing YAML dates

viktomas:
Small follow-up commit for https://github.com/silverbulletmd/silverbullet/commit/2eb7e25854000090bdb3a8729aed3c3c1f06e381#

- removes `pad` function in favour of `String.prototype.padStart`
- adds test cases for the date formatting

Thank you @zefhemel for not only taking the time to respond but also to fix a small bug in date parsing (discussion here https://community.silverbullet.md/t/struggling-with-yaml-frontmatter-date-attribute-and-date-comparison-query-in-silverbullet/454/13)

I thought I would also thank you for putting some of my time in. Let me know if this looks OK or if you want me to make any changes.

Since this is a pure refactoring, there doesn't have to be any CHANGELOG entry 👍

### DIFF
--- a/plug-api/lib/json.test.ts
+++ b/plug-api/lib/json.test.ts
@@ -18,4 +18,6 @@ Deno.test("utils", () => {
   assertEquals(cleanupJSON({ a: [{ "a.b": 1 }] }), {
     a: [{ a: { b: 1 } }],
   });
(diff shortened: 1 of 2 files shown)


zefhemel:
Thanks for cleaning this up 🙏

PR #932, opened 2024-07-11
Libraries

### LINKED ISSUE #926: Implement library updates
Currently there is no way to automatically update libraries. Upon initial onboarding, it is suggested to the user to import the Core library. However, after that this library is never updated again, even though new features may be added to it.

This issue is about adding update functionality to libraries.

# Configuration
In `SETTINGS` libraries can be configured as follows:

```yaml
libraries:
- source: "[[!silverbullet.md/Library/Core/]]"
  exclude:
  - "[[!silverbullet.md/Library/Core/Widget/Table of Contents]]"
  - "[[!silverbullet.md/Library/Core/Whatever/*]]"
```
Only `source` is required and will use page link with federation syntax. This is consistent with other page link settings and is convenient because it supports completion (also for federated links).

`exclude` can contain individual pages, and probably should support a `*` for wildcards. The purpose of this to be able to disable specific pages/templates if you like.

# Update behavior
A `Library: Update All` or similar command will do the following.

For each library:
1. Wipe out the prefix used by the library, e.g. `Library/Core/` in the current space to start with a clean slate
2. Redownload all pages again into the prefix

# What about local customization
In this initial version this will not be supported. While we keep the library paths editable, any changes will be wiped upon a library update. If you want to make changes to a template or other thing, the new recommendation is to make a copy of it, move it to your personal library and add it to `exclude` for the source library.

# Default SETTINGS
For new users, a SETTINGS will be generated that contains:

```yaml
libraries:
- source: "[[!silverbullet.md/Library/Core/]]"
```

Upon first boot, the `Library: Update All` command will automatically run to perform the initial import.

# What about personal libraries
We will document the _convention_ of using `Library/Personal` for all custom libraries.

zefhemel:
Fixes #926


bitraten:
@zefhemel Just one finding: This seems to break importing the Core Library via the Getting Starting template on new installations running a previous version (experienced with 0.7.7 on NixOS 24.04) since it pulls in the updated command from silverbullet.md which is not found in the local silverbullet instance though 😐 
Was able to import manually still, generally thank you so much for this nice software!

zefhemel:
Right. It would be a pain to keep supporting old versions. Can somebody update the version in nix perhaps?

PR #945, opened 2024-07-15
Per-tag page styling

onespaceman:


web/client.ts
```diff
@@ -1124,6 +1124,10 @@ export class Client {
         // Nothing in the store, revert to default
         enrichedMeta = doc.meta;
       }
+      if (enrichedMeta.itags) {
```
zefhemel, line 1127: I'm thinking in what scenario this is not set and if that's the case if the attribute shouldn't be cleared?

web/client.ts
```diff
@@ -1124,6 +1124,10 @@ export class Client {
...
+        const pageTags = enrichedMeta.itags.join(" ");
+        editorView.contentDOM.setAttribute("tags", pageTags);
```
zefhemel, line 1129: For form and attribute pollution reasons should this attribute be called `data-tags`?

zefhemel (changes requested):
Nice feature! I'd also be interested in something like this for https://silverbullet.md/Page%20Decorations

That's separate of course, or maybe not?

onespaceman:
Moved the attributes to the body tag, with data- prefix.
Now all metadata properties are added, but I'm not sure if the some like the contenttype, dates, or perms are useful. Should I filter them out?

> I'd also be interested in something like this for https://silverbullet.md/Page%20Decorations

I'm not sure what you're asking for here?

onespaceman:
Ok on a reread, I think that, instead of injecting the html for the prefix, you want to just attach a class to the links, and let people do whatever with it with space styles instead. Is that correct?

onespaceman:
Honestly, this is a pretty powerful feature. Allow any property instead of just a class, rebrand it as an autotagger, and now add page tags and whatever other metadata to page through a query.

Then attach the meta properties as attributes to the html and style as you please through space-styles

```
*[myProperty="value"] {
  color: #F00;
}
```

Could even do the same with the prefixes.

```
*[pagePrefix="tree"]::before {
  content: "🎄 ";
}

File: web/client.ts [hunk 3]
```diff
@@ -1155,6 +1157,25 @@ export class Client {
     }
   }
 
+  pageMetaToAttributes(meta: any) {
+    const bodyEl = this.parent.parentElement;
+    if (!bodyEl) return;
+    for (const p in meta) {
```
zefhemel, line 1163: Should this `p` not be somehow normalized before injecting it into the DOM? Potentially there could be weird characters in there, spaces, maybe other things?

onespaceman:
Check what I have so far if this is what you wanted.

an example snippet for page decorations based on itag:
```
/* add a prefix /*
body[data-itags~="test"] #sb-current-page .cm-line::before
a[data-itags~="test"]::before,
.sb-option[data-itags~="test"] .sb-name::before {
  content: "✨ ";
}

/* color it purple */
body[data-itags~="test"] #sb-current-page .cm-line,
a[data-itags~="test"],
.sb-option[data-itags~="test"] .sb-name {
  color: purple;
}
```

lib/web.ts
```diff
@@ -4,6 +4,7 @@ export type FilterOption = {
   description?: string;
   orderId?: number;
   hint?: string;
+  attributes?: { [key: string]: string };
```
zefhemel, line 7: Minor, but I tend to use `Record<string, string>` generally

web/cm_plugins/util.ts
```diff
@@ -4,12 +4,15 @@
...
 import { Decoration, EditorView, WidgetType } from "@codemirror/view";
+import { htmlEscape } from "../../plugs/markdown/html_render.ts";
+
 type LinkOptions = {
   text: string;
   href?: string;
   title: string;
   cssClass: string;
   from: number;
+  attributes?: { [key: string]: string };
```
zefhemel, line 15: Same as above

web/cm_plugins/util.ts
```diff
@@ -23,6 +26,10 @@ export class LinkWidget extends WidgetType {
     anchor.className = this.options.cssClass;
     anchor.textContent = this.options.text;
 
+    for (const a in this.options.attributes) {
```
zefhemel, line 29: I get nervous seeing this pattern in modern JS. I prefer the safer `for(const [attr, val] of Object.entries(this.options.attributes))`

web/cm_plugins/util.ts
```diff
@@ -168,3 +175,31 @@ export function isCursorInRange(state: EditorState, range: [number, number]) {
...
  */
 export const invisibleDecoration = Decoration.replace({});
+
+/**
+ * Flatten and sanitize metadata objects for injecting into HTML
+ */
+export function metaToAttributes(meta: any): { [key: string]: string } {
+  let attributes: any = {};
+
+  for (let m in meta) {
```
zefhemel, line 185: Same as above, I prefer Object.entries()

web/cm_plugins/util.ts
```diff
@@ -168,3 +175,31 @@ export function isCursorInRange(state: EditorState, range: [number, number]) {
  * Decoration to simply hide anything.
  */
 export const invisibleDecoration = Decoration.replace({});
+
+/**
+ * Flatten and sanitize metadata objects for injecting into HTML
+ */
+export function metaToAttributes(meta: any): { [key: string]: string } {
+  let attributes: any = {};
```
zefhemel, line 183: Why not `Record<string, string>`?

web/cm_plugins/util.ts
```diff
@@ -168,3 +175,31 @@ export function isCursorInRange(state: EditorState, range: [number, number]) {
...
+
+  for (let m in meta) {
+    // Skip non-useful attributes
+    if (["created", "lastmodified", "perm"].includes(m.toLowerCase())) {
```
zefhemel, line 187: I don't think you need to lowercase this? I think these will always be capitalized consistently? `created`, `lastModified`, `perm` no?

zefhemel: And maybe you want to style based on `perm` 🤷🏻

web/cm_plugins/wiki_link.ts
```diff
@@ -74,8 +75,10 @@ export function cleanWikiLinkPlugin(client: Client) {
           // Hide the ^ prefix
           cleanLinkText = cleanLinkText.slice(1);
         }
-        const linkText = alias ||
-          ((pageMeta?.pageDecoration?.prefix ?? "") + cleanLinkText);
```
zefhemel, line 78: This seems to break `prefix` decorations, why did you remove this?

web/editor_state.ts
```diff
@@ -105,6 +105,25 @@ export function createEditorState(
             client.clientSystem.slashCommandHook,
           ),
         ],
+        addToOptions: [{
+          render: (
+            completion: any,
+            _state: EditorState,
+            _view: EditorView,
+          ) => {
+            if (completion.metaData) {
```
zefhemel, line 114: So this would allow to style auto complete as well? If so, an example of this would be helpful in the docs.

web/cm_plugins/wiki_link.ts
```diff
@@ -74,8 +75,10 @@ export function cleanWikiLinkPlugin(client: Client) {
...
+        const linkText = alias || cleanLinkText;
+
+        const attributes = metaToAttributes(pageMeta);
+        console.log(attributes);
```
zefhemel, line 81: Probably want to remove this one

zefhemel (changes requested):

web/cm_plugins/wiki_link.ts
onespaceman: This replaces pagedecorations no? see comment below for reasoning.

onespaceman:
Just to go over what I have in mind:
1. Autotagger - 
Apply any arbitrary tag in settings instead of just limited to prefix.
Example:
```
autotag:
- where "tags= 'testTag'"
tags: "additionalTag"
iLikeKeys: "And also values"
hidden: true
```

2. Add a hidden property to pageMeta so that it can be hidden on a page-to-page basis or through autotagger instead of relying on prefix.

3. Attach pagemeta to wherever page names and page links appear, so that it can be styled with space-styles as shown [here](https://github.com/silverbulletmd/silverbullet/pull/945#issuecomment-2246960131). More flexible by allowing to style colors, background, font, prefix etc. at the cost of a somewhat ugly and harder to understand css snippet.
Or I could keep the prefix option, generate the css, and append it when building space-styles.

zefhemel:
> Just to go over what I have in mind:
> 
> 1. Autotagger -
>    Apply any arbitrary tag in settings instead of just limited to prefix.
>    Example:
> 
> ```
> autotag:
> - where "tags= 'testTag'"
> tags: "additionalTag"
> iLikeKeys: "And also values"
> hidden: true
> ```

So I was thinking along similar lines when I wrapped up the initial version of pageDecorations. My thinking was: what if we would add a generic mechanism to add attributes dynamically to _any_ object (selected based on a `where` clause like pageDecorations and other things work now). This would be a way to dynamically add pageDecorations onto pages, but could be used in other ways too. You could create dynamic attributes on tasks, you could use these attributes in queries etc.

For instance, you could then implement page decorations like this:

```yaml
decorators:
- where: 'tags = "person"'
  attributes:
     pageDecoration.prefix: "🥸 "
```

Or an even more out-there solution would be to allow Expressions as values so they can be dynamically computed:

```yaml
decorators:
- where: 'tags = "person"'
  attributes:
     pageDecoration.prefx: '"🥸 "' # Just a literal string
     fullName: "firstName + ' ' + lastName" # Compute based on other attributes, you could even call functions here
```

This would give all #person tagged objects a pageDecoration and fullName automatically. This would need to be implemented at the DataStore level.

In the end I didn't go for it because I feared that:

1. It would be too expensive, e.g. the way queries work is basically through a table scan where every item is evaluated against the `where` clause. This would add more complexity to that: as in, it would also have to create an "enriched" version of the item with all these dynamic attributes added and then run the filter against that. This seemed potentially expensive.
2. It adds complexity. Before you could just see: these are the attributes of this page, it's obvious because some are built in (like lastModified), others I defined in the front matter. That's it. No surprises. Adding dynamic attributes based on selectors could make this a whole lot more complicated (especially if people accidentally do things like applying them to _all_ objects, e.g. when they'd just say `where true` 🤯 )

But I'm open to be challenged on this. And as I'm writing this, I'm tempted to try it. I think this can be done, and it would be a very powerful mechanism that can be leveraged for various use cases, not just page decorations. But it's also one of those "with great power comes great responsibility" type of things.

> 2. Add a hidden property to pageMeta so that it can be hidden on a page-to-page basis or through autotagger instead of relying on prefix.

I'm not sure I follow this one.

> 3. Attach pagemeta to wherever page names and page links appear, so that it can be styled with space-styles as shown [here](https://github.com/silverbulletmd/silverbullet/pull/945#issuecomment-2246960131). More flexible by allowing to style colors, background, font, prefix etc. at the cost of a somewhat ugly and harder to understand css snippet.
>    Or I could keep the prefix option, generate the css, and append it when building space-styles. That way a simple prefix option is available in settings, with more advanced styling through space-styles.

I think the prefix use case is going to be the most common thing, so I'd prefer to keep that easy to do, as it is. I also still like my suggestion to add an explicit `pageClass` attribute to `pageDecoration` that would then be attached to the body to make targeting CSS styles easier. I'm ok with also attaching other page attributes to the body as `data` attributes, but for me that's less essential because you can achieve the same thing with `where` targeted decorations.

zefhemel:
FYI: I'm working on (1) as I proposed after all. Performance impact doesn't seem bad, hang tight :)

onespaceman:
> FYI: I'm working on (1) as I proposed after all. Performance impact doesn't seem bad, hang tight :)

I did this too, lol
check the last commit, I just attached the autotagger to the page indexer

zefhemel:
Have a look at: #995

zefhemel:
> > FYI: I'm working on (1) as I proposed after all. Performance impact doesn't seem bad, hang tight :)
> 
> 
> 
> I did this too, lol
> 
> check the last commit, I just attached the autotagger to the page indexer

Right, the issue there would be that you have to reindex your entire space every time you change these "auto tag" rules. My implementation does this on the fly. It's slightly more costly at read/query time though.

onespaceman:
I think that should be it.

New example:
```
objectDecorators:
- where: 'tags = "pink"'
  attributes:
    pageDecoration.cssClass: "['make-it-pink']"
```
Make page names pink
```
.make-it-pink.sb-decorated-object {
  &,
  .sb-name,
  .cm-completionLabel {
    color: hotpink;
  }
}
```
Add pink accent border to editor
```
body.make-it-pink {
  #sb-editor .cm-content {
    border-left: solid 0.5em hotpink;
  }
}
```

At this point we kinda need a dedicated page for snippets on the site.

onespaceman:
This one might be useful enough to add as a default
```
- where: 'tag = "attachment"'
  attributes:
    pageDecoration.prefix: "'🔗 '"
```

zefhemel:
Excellent, the only thing I changed here to rename the decoration to cssClassES (because it expects a list), and added some examples (nice Christmas decoration ones).

PR #951, opened 2024-07-17
chore: Update index.html to include no referrer

### LINKED ISSUE #950: URLs are not using noreferer in their links
Source: https://community.silverbullet.md/t/2-improvements-and-i-would-recommend/666

noreferer helps hide where a link was clicked/redirected from, which means that websites can find my notes server, not that big of a deal, but more a privacy concern.

You can use the https://play.silverbullet.md/ instance, and paste in https://duckduckgo.com/?t=h_&q=user-agent&ia=answer when you click the link you can see that https://play.silverbullet.md/ shows on the Referer: line.

roceb:
Fixes #950


zefhemel:
Thanks 😊

PR #960, opened 2024-07-19
Add setting for smart quotes

Maarrk:
As discussed in [Unwanted Unicode quotation marks - Troubleshooting - SilverBullet Community](https://community.silverbullet.md/t/unwanted-unicode-quotation-marks/669)

**Needs a refresh for the setting to take effect.** I noticed that syscalls don't work in the `web` folder, so I reworked the array into function like `createCommandKeyBindings`.

I don't know nice ways to save the reference to client and check the current setting when the shortcut is triggered ¿I would use `var` to save reference to client inside the function? I don't know...

If we accept that this needs a reload, I could make it a bit simpler.

### DIFF
--- a/type/web.ts
+++ b/type/web.ts
@@ -18,6 +18,7 @@ export type BuiltinSettings = {
   shortcuts?: Shortcut[];
   hideSyncButton?: boolean;
   hideEditButton?: boolean;
+  useSmartQuotes?: boolean;
   maximumAttachmentSize?: number;
   defaultLinkStyle?: string;
   actionButtons: ActionButton[];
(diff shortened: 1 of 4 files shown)


zefhemel:
I think it's perfectly fine to require a reload, want to simplify? I like simple (this is already pretty simple honestly)

Maarrk:
I have seen "try reloading the page" a few times on  the Community, so I changed the SETTINGS explanation a bit. Probably not worth tracking and documenting exactly which ones need it

zefhemel:
Ok. Good to go?

Maarrk:
I believe it's good *now*

PR #962, opened 2024-07-19
Add hidden pages

Maarrk:
Originally proposed in [Soft delete notes - Ideas - SilverBullet Community](https://community.silverbullet.md/t/soft-delete-notes/674) thread.

I love how PR is documented in the `website/` every time

### DIFF
--- a/plug-api/syscalls/editor.ts
+++ b/plug-api/syscalls/editor.ts
@@ -48,7 +48,7 @@ export function navigate(
 }
 
 export function openPageNavigator(
-  mode: "page" | "meta" = "page",
+  mode: "page" | "meta" | "all" = "page",
 ): Promise<void> {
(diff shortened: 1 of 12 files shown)


zefhemel:
I'm up for the all page picker. But for the #hidden thing I'd suggest a different solution now that we have page decorations. Namely... how about we make this a page decoration instead? 

```yaml
pageDecorations:
- where: "tags = 'hidden'"
  hide: true
```

Now naming is hard, so maybe something more verbose like `hideInPicker` would be better. 

This way we don't need to hard code another tag. 

Implementation wise this shouldn't be very hard. The pageDecorations attribute is available in the picker on any PageMeta.

Maarrk:
I like that it's opt in and has a configurable query, so the name isn't that important. Will update

zefhemel:
Nice

PR #963, opened 2024-07-19
Treat Chinese ideograms as separate words

Maarrk:
As requested on Community: [The Search function does not support Chinese characters - General - SilverBullet Community](https://community.silverbullet.md/t/the-search-function-does-not-support-chinese-characters/367/2), the discussion whether this makes sense or not continues there

Had to `export` the `tokenizer` to write tests in a separate file, is this how it's supposed to be done?


Maarrk:
Marked as draft to wait for feedback from some Chinese speakers

Maarrk:
It has to be a substring, random order is bad

PR #965, opened 2024-07-20
Added a timeout for fetch

### LINKED ISSUE #947: Sync mode takes too long to detect it's offline
I'm pretty sure I saw a discussion about this recently, but didn't see a gh issue.

I'm traveling and noticed that whenever my phone is not connected to my VPN, I basically can't use the silverbullet pwa right after I open it.

After opening it, it takes 2 or 3 minutes for the index note to load. Once it does, I get the red status bar and things seem to work fine until I close and reopen the pwa later.

My guess without being able to see the console is maybe it's timing out trying to connect to my server. The dns would still resolve, but it'd be to a private ip.

roceb:
Added a 5 second timeout on fetch request. Fixes #947

### DIFF
--- a/common/spaces/http_space_primitives.ts
+++ b/common/spaces/http_space_primitives.ts
@@ -29,8 +29,8 @@ export class HttpSpacePrimitives implements SpacePrimitives {
     }
 
     try {
-    const signal = AbortSignal.timeout(5000)
-      const result = await fetch(url, {signal,...options});
+      options.signal = AbortSignal.timeout(5000)
+      const result = await fetch(url, options);
       if (result.status === 503) {
         throw new Error("Offline");
       }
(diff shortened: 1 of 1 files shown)


common/spaces/http_space_primitives.ts
```diff
@@ -29,7 +29,8 @@ export class HttpSpacePrimitives implements SpacePrimitives {
     }
 
     try {
-      const result = await fetch(url, options);
+    const signal = AbortSignal.timeout(5000)
+      const result = await fetch(url, {signal,...options});
```
justyns, line 33: I haven't tested it, but the example on https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static looks like this:

```
  const res = await fetch(url, { signal: AbortSignal.timeout(5000) });
```

You could probably just change `options` directly, like `options.signal = AbortSignal.timeout(5000);`

zefhemel:
Looks good, although I'd extract the 5000 as a constant at the tip of the file.

zefhemel:
Added semicolons for sanity reasons. I'll merge this and let people report back if this is helping with #947

PR #974, opened 2024-07-24
 system.invokeFunctionOnServer syscall to force proxying syscalls to server

justyns:
Related to https://github.com/justyns/silverbullet-ai/issues/47

It's pretty much a copy of system.invokeFunction, but always uses `proxySyscall`.  I'm not committed to this approach if there's a better idea

### DIFF
--- a/common/syscalls/system.ts
+++ b/common/syscalls/system.ts
@@ -45,6 +45,34 @@ export function systemSyscalls(
       }
       return plug.invoke(functionName, args);
     },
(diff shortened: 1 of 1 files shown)


zefhemel:
Let's go with this to start and see how far it get you.

zefhemel:
Once you confirmed this helps, probably it's nice to add a wrapper like this one too: https://github.com/silverbulletmd/silverbullet/blob/main/plug-api/syscalls/system.ts#L6

PR #988, opened 2024-07-28
Support linking to and moving to line number in pages

Maarrk:
Added new link syntax as proposed in [Community post](https://community.silverbullet.md/t/grep-plug-for-alternate-search-implementation/709/5?u=maarrk) to link to a specific line.

Also added a syscall and command with the same logic for completion.

I think since commands are run by the user from the Picker or button, I made them show where the cursor was moved


Maarrk:
The force pushes were to handle this kind of references in Task Plug, and a small refactor into function to keep the logic in a single place.

zefhemel:
Tweaked two minor things. Thanks a lot!

PR #989, opened 2024-07-28
Use aliased preact import in Panel component.

### LINKED ISSUE #986: Everything breaks when using markdown preview
When trying to use the markdown preview command either in my own silverbullet instance or on silverbullet.md the title shifts, no markdown is shown and the following error pops up in the console:
image

Afterward the page picker and command palette do not seem to respond anymore either until reloading or deleting cookies.

joekrill:
This was causing multiple Preact versions to be imported, creating an error when opening panels. Fixes #986.


zefhemel:
Thanks!

PR #995, opened 2024-07-29
Object decorators

zefhemel:
To replace the `pageDecoration` with something more generic and powerful.

cc @onespaceman


onespaceman:
This looks good!

I'll rebase #945 to this branch.
Also I think you're right about just attaching a class through decorator, instead of data-attributes to the html. It should make the snippets much cleaner.

onespaceman:
A request: Could you make arrays be merged instead of overwritten when eriching metadata. useful for applying multiple classes/tags to a page

onespaceman:
Also, all attributes (dates, ref, name etc.) can be overwritten with the decorator.

zefhemel:
Ok, both should now be there.

PR #997, opened 2024-07-29
Improve Plug distribution through GitHub Releases (fix #975)

### LINKED ISSUE #975: Can't distribute Plug using ghr prefix
Looking at this line, it seems that I just need to name the `.plug.js` file same as repository
https://github.com/silverbulletmd/silverbullet/blob/1d8dfb964a25021fbe31430d0fcc09362c094617/plugs/plug-manager/plugmanager.ts#L158

But when I did just `ghr:Maarrk/silverbullet-grep/latest` it got ignored by this code:
https://github.com/silverbulletmd/silverbullet/blob/1d8dfb964a25021fbe31430d0fcc09362c094617/plugs/plug-manager/plugmanager.ts#L41-L48

Maarrk:
Summary:
- fixed the problem with `ghr` uri not ending with `.plug.js`
- the release manifest is always downloaded (instead of only doing that for latest) to give more descriptive errors
- since *every* repository in [Plugs#Third-party plugs](https://silverbullet.md/Plugs#Third-party%20plugs) is named with the `silverbullet-` prefix, this is handled and documented

### DIFF
--- a/plugs/plug-manager/plugmanager.ts
+++ b/plugs/plug-manager/plugmanager.ts
@@ -38,17 +38,28 @@ export async function updatePlugsCommand() {
     for (const plugUri of plugList) {
       const [protocol, ...rest] = plugUri.split(":");
 
-      const plugNameMatch = /\/([^\/]+)\.plug\.js$/.exec(plugUri);
-      if (!plugNameMatch) {
-        console.error(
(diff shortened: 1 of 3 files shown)


zefhemel:
Nice 👍🏻

PR #1009, opened 2024-08-01
Implement content merging + preservation of cursor

jcgurango:
This PR implements content merging and preservation of the editor cursor/selection. At the moment, when new updates are received from the client, the entire editor state is wiped out and replaced with what we receive from the server and the cursor goes back to the default position. With this PR, it instead does a 3 way text merge between:

1. The page text last time the editor state was loaded.
2. The server version of the file.
3. The current text in the editor.

Additionally, it'll try to preserve any cursor position or selection it finds. I've done some testing on this, and it seems to work pretty well. I can now generally be typing away, and when the page reloads I can continue to just keep typing.


zefhemel:
I like this (and it was on my list of things to do at some point). There's already an implementation (not currently used for this purpose) doing something similar: 

https://github.com/silverbulletmd/silverbullet/blob/357a50f8203dc7ea9dcdb0bbacc5f4d5319dac63/web/syscalls/editor.ts#L30

If you follow that, you'll find another library that I've used there. Could you perhaps rework this to share this logic somehow?

jcgurango:
I see what you're driving at. Just made the update and it is working swell.

jcgurango:
Hmm getting some strange behavior upon further testing.

jcgurango:
Okay, I've been fiddling with this for a bit and I think fundamentally this approach won't work. I left my code here so you can try it out yourself. The problem here is this: say there's two instances with the same page open.

1. Instance 1 adds a sentence and stops typing. This saves.
2. Instance 2 is mid-typing and the page is reloaded. All those changes from Instance 1 get added atomically. Instance 2 finishes typing, the page is saved.
3. Back in instance 1, the page is reloaded, and it got instances 2's changes *and its own changes*. But it doesn't know that those changes were persisted to the server, all it knows is it got some changes so what ends up happening is it adds those changes (since the last version it got from the server didn't have them), and you get some duplicates.
4. Repeat over and over again and you'll have even more duplicates.

The way the codemirror collab plugin solves this is by tagging each atomic update to a client ID, so if a client receives updates from its own client ID you're supposed to ignore them and then it has the functionality to rebase those changes against the latest version from the server (so last server version + latest server changes + local changes that haven't been persisted). We don't have this ability to identify clients and their changes, and adding it would result in a ton of complexity.

So, I still suggest we use the three-way merge strategy. It gets around this problem because if it receives an update from the server, and it happens to match what the client already has in the editor, then it just automatically considers it as an agreed upon change and keeps it without duplicating it. If the only data point we can look at is the editor content, the last page content retrieved from the server, and the incoming page content then it makes sense that we'd have to do a comparison of all three. Now, I spent some time trying to make my own implementation of this using that logic but there's a lot of cases to consider so I think it's a bit over my head to implement. Plus the `{ from, to, insert }` structure doesn't lend itself well to this sort of thing. A diff delta like what textdiff-create does makes more sense for this kind of comparison:

```js
const createPatch = require('textdiff-create');

const v1 = 'The sleepy brown fox';
const v2 = 'The quick brown fox jumps over the lazy dog';

const delta = createPatch(v1, v2);

console.log(JSON.stringify(delta));
// [[0,4],[-1,6],[1,"quick"],[0,10],[1," jumps over the lazy dog"]]
```

Because in this case you can iterate over both, see where both agree, and try to resolve any conflicts there.

If you think I should refactor that initial code I pushed in some way I could do those changes. I'm also open to using a different package, but I think if we don't consider all three - last retrieved content, current editor content, incoming content - then the implementation will be lacking.

jcgurango:
So I've reverted back to using 3-way merge but now I'm using diffAndPrepareChanges to apply them without overwriting the editor state. Also marking this as draft for now. Two things of note in this:

1. I've switched from using three-way-merge to node-diff3. I was finding three-way-merge a little lacking (I kept losing stuff because merge conflicts were just being dropped and I couldn't make sense of some of the merging it was doing) so I pulled in node-diff3 which seems to use a better algorithm and I have better control over how conflicts are handled. In this case, I'm just merging them both (equivalent of doing a git merge and just removing the markers). This occasionally results in duplicates but I think it works better. I'll roll with this for a while to get more of a feel for it.
2. diffAndPrepareChanges doesn't seem to work very well when there's many changes. I kept running into this error:
    image
    It seems if you apply many changes at once, the later changes are applied on the document *after* the previous changes have been applied. I haven't really checked if this has implications on `editor.setText()` but I have to guess that it does. I was still able to reuse it albeit in a roundabout way (in a while loop doing diff -> dispatch one change at a time).

Anyway, in actual practice, as a single user on multiple devices this seems to work pretty well. However, synthetically I can generate some pretty messed up merges by just constantly editing the same line in the same document on multiple devices. @zefhemel do you think a "Hey your version and the server version have diverged" type of window with some kind of conflict resolution would make sense?

zefhemel:
It can make sense, but the question is what the "some sort of conflict resolution" part would look like UX wise. It sounds like it could be quite complicated.

PR #1012, opened 2024-08-02
Add an in-memory cache of the file list for disk spaces

justyns:
Related to #947 

_(background note:  my primary sb instance is on a server with a network file system and can be kind of slow when accessing 100s or 1000s of small files)_

Basically this caches the file list in a variable.  

The cache is emptied every time a file is deleted or saved.

Every time fetchFileList is called, the cache is returned if it's less than 1 minute old.  A background sync of the cache is also triggered.  This could be on a timer instead, I'm not sure if it's needed or not?  The client triggers syncs pretty regularly, but if there are no clients connected, I'm not sure if there's a reason to constantly update the file list.

I haven't really tested this much with sync-mode and being offline for an extended period of time, but in my tests so far - /index.json is consistently returned much quicker now.

To test:

- create a new empty directory to use as a space
- create a bunch of files: `for x in `seq 1 5000`;do echo $x;date > testnote-$x.md; done`
- in getFileList(), uncomment `await new Promise((resolve) => setTimeout(resolve, 1));` to cause the file list to take longer

### DIFF
--- a/common/spaces/disk_space_primitives.ts
+++ b/common/spaces/disk_space_primitives.ts
@@ -13,11 +13,11 @@ function normalizeForwardSlashPath(path: string) {
 }
 
 const excludedFiles = ["data.db", "data.db-journal", "sync.json"];
-let fileListCache: FileMeta[];
-let fileListCacheTime: number;
 
 export class DiskSpacePrimitives implements SpacePrimitives {
   rootPath: string;
+  fileListCache: FileMeta[] = [];
+  fileListCacheTime: number = 0;
 
   constructor(rootPath: string) {
     this.rootPath = Deno.realPathSync(rootPath);
@@ -93,8 +93,11 @@ export class DiskSpacePrimitives implements SpacePrimitives {
         await file.utime(new Date(), new Date(meta.lastModified));
       }
       file.close();
-      fileListCache = [];
-      fileListCacheTime = 0;
+
+      // Invalidate cache and trigger an update
+      this.fileListCache = [];
+      this.fileListCacheTime = 0;
+      this.updateCacheInBackground();
 
       // Fetch new metadata
       return this.getFileMeta(name);
@@ -125,8 +128,11 @@ export class DiskSpacePrimitives implements SpacePrimitives {
   async deleteFile(name: string): Promise<void> {
     const localPath = this.filenameToPath(name);
     await Deno.remove(localPath);
-    fileListCache = [];
-    fileListCacheTime = 0;
+
+    // Invalidate cache and trigger an update
+    this.fileListCache = [];
+    this.fileListCacheTime = 0;
+    this.updateCacheInBackground();
   }
 
   async fetchFileList(): Promise<FileMeta[]> {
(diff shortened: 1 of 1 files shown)


zefhemel:
Adding another layer of caching is always scary (two hard problems in computer science: cache invalidation and naming things). The concern I see with a 1 minute cache is that new files put on disk (outside of SilverBullet's control) would only be detected after 1 minute, whereas this now happens in a matter of probably seconds. For instance, if you run the git plug to synchronize your space between two servers or whatever, it would likely take a minute or more for _new_ files to appear on the other end.

Now the question is: is that a problem? Should we shorten this to 20s or so, or... not do it at all. It's hard to judge to be honest.

Another approach would be to to use something like this: https://docs.deno.com/runtime/tutorials/file_system_events/ to trigger a refresh of the cache, but I have no idea how reliable this is, especially on remote file systems and other weird setups people may have.

zefhemel:
Another issue I foresee is indeed sync mode, which uses the file list to determine changes. If those cached values don't match eg what getFileMeta gives back, I'm not sure what will happen.

justyns:
To clarify, the 60 seconds is basically a way to guarantee that the cache will never be older than 60 seconds. In reality, each time you call `GET /index.json`, the cache will get updated in the background.  How old it is depends on how fast it takes to build that cache.

So if it takes ~7 seconds for your `getFileList()` to run normally, the behavior looks like this right now:

- GET /index.json - request aborted at 5 seconds
- sleep ~5 seconds
- GET /index.json - request aborted at 5 seconds
- sleep ~5 seconds
- GET /index.json - request succeeded at 3 seconds (maybe because of OS-level caching or something else that caused it to go faster)

With this change, the behavior looks like this:

- App starts up and calls getFileList, building the initial cache
- client GET /index.json - request succeeded at 100ms due to returning cached list
- cache update triggered in background
- sleep ~5 seconds
- GET /index.json - request succeeded at 100ms and potentially returned same list as first /index.json request
- another cache update triggered in background, but does nothing due to the last cache update still running
- first cache update finishes
- GET /index.json - request succeeded at 100ms, returning the new cached filelist

Whenever deleteFile or writeFile is called (by any client), the cache is also deleted and updated in the background. This _should_ let sync work as normal since no one will be receiving a cached response after a file is updated. With external changes like from the git plug, changes should still show up within 1 or 2 requests due to each request triggering a cache update.  We could also add an event like "cache:clear-file-list" or something that plugs could use.

> Another approach would be to to use something like this: https://docs.deno.com/runtime/tutorials/file_system_events/ to trigger a refresh of the cache, but I have no idea how reliable this is, especially on remote file systems and other weird setups people may have.

I'm not sure about the windows/mac implementation, but inotify on linux doesn't handle network file systems ime.  I wouldn't be against using it though since it'd work for the majority of use cases.  We'd just also need something that scans the fs on a timer too.

Also the current changes in the PR only updates the cached based on requests to /index.json which trigger getFileList.  I thought about _only_ serving cached requests and having something constantly updating the cache in the background, but it seemed more risky because of the issues you mentioned.

zefhemel:
Ok got it. Ok let me do some testing with this then.

Issue #942, opened 2024-07-14
Add edit/bake/refresh buttons to transcluded pages


zefhemel:
They're missing now, but would be useful. Also for consistency.

zefhemel:
Just mentioning @onespaceman in case he's bored and wants continue is excellent work in this area 🥸

PR #1013, opened 2024-08-03
Widget buttons for transclusions (#942)

onespaceman:


zefhemel:
Stream of consciousness ahead:

What I had in mind the Edit button would do is navigate to the transcluded page. However, now that I'm thinking about it this is somewhat inconsistent with the general Edit button meaning on widgets, which usually indeed puts the cursor inside the body to allow you to edit the source (query, template).

Now the question is: what is the "source" in this case, the link itself, or the page it links to? 

In practice what I would want to do when pushing this edit button is edit the underlying page (that is transcluded) so I'd still prefer the navigate option. Do you agree?

onespaceman:
Sounds reasonable. I could also change the icon to a link and hover text to "Open page" so there's no ambiguity.

zefhemel:
Sounds good

zefhemel:
Ready to look at again?

onespaceman:
yes

zefhemel:
Works great, thanks!

PR #1039, opened 2024-08-16
Add spacing to headlines

puresick:
Block direction margin has been added to headlines in this PR to improve the readability and visual structure, especially in large text documents.

I chose to use the `em` unit for the margin to be relative to each headline font size.

Hope it is okay that I added the spacing values to a separate partial. Did that so that there is one place of truth which needs to be edited if these should be changed in the future.

Also let's talk about the values if they feel off. I tried to get a good compromise without adding "handcrafted" margins to each headline separately.

### DIFF
--- a/web/styles/editor.scss
+++ b/web/styles/editor.scss
@@ -1,4 +1,4 @@
-@use "./partials/spacing";
+@use "./partials/spacing" as spacing;
 
 .cm-focused {
   outline: none !important;
--- a/web/styles/editor.scss
+++ b/web/styles/editor.scss
@@ -1,3 +1,5 @@
+@use "./partials/spacing";
+
 .cm-focused {
   outline: none !important;
 }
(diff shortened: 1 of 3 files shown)


zefhemel:
I think you may have misnamed a file here, this doesn't build

puresick:
For some reason the namespace was correctly retrieved by the partials filename. I explicitly declared one (see https://sass-lang.com/documentation/at-rules/use/#choosing-a-namespace) which solves the issue and does build successfully on my machine.

puresick:
Sorry for being silent for so long, got some busy weeks currently.
Looks like my naive way of adding styles does not work here. I will have to dig into CodeMirrors documentation on how to properly configure this.

I will let you know once I start working on a proper implementation.

puresick:
Hey, I am sorry for letting you wait that long.
Unfortunately I will not have any free time in the foreseeable future to work on this.

Therefore I will close this issue so it does not lie around and clutter things up.
If anybody else (or maybe I someday) has time and wants to work on this they might be able to continue here.

PR #1056, opened 2024-08-21
Use setText instead of replacing entire editor state

jcgurango:
Piggy-backing off of #1055, this PR uses `editor.setText()` instead of creating an entirely new `editorState` when new changes are received from the server. It's maybe not the most elegant solution but until there is a better one, this at least lets you keep typing away and keeps your cursor position when new changes are retrieved. I've also updated the editor.setText() to be an isolated transaction, this way when you lose your local changes you can undo to get them back then redo to get the server changes in.


zefhemel:
This is really great, thank you!

PR #1062, opened 2024-08-24
Docs fix

Jc-L:
In expressions, "/" was described as "addition" rather than "division"
Note that "arithmatic" was left unchanged but could be a typo of "arithmetic"


zefhemel:
Thanks!

Issue #635, opened 2024-01-19
Export (Share as) PDF, ePub, other formats


simone-viozzi:
Hi, thank you for the great tool.

If I need to share a note with a colleague, how can I print the silverbullet rendering of the note?

Doing print page to PDF from the browser (edge | firefox | chrome) prints only the first page.

Is there a command / plug I need to use to just save the rendered page as PDF?

Thank You!

zefhemel:
There’s no great way to do this right now. The only thing that comes to mind is to open the markdown preview (Cmd-p/Ctrl-p) and copy and paste from there into a word document or some other word processor and export to PDF from there.

I do see the value of this, though, but would have to investigate how to implement this.

gorootde:
In Safari on macOS all the pages are printed. In iOS / iPad OS as well. Seems to be browser related.

zefhemel:
One thing I'd like to investigate is how much weight it would be to add pandoc and latex to the docker image to achieve this and offer export to docx, epub and other formats while at it.

zefhemel:
Or this can be done as a separate pandoc plug which dynamically attempts to install all dependencies. That probably makes more sense. Will investigate at some point.

PR #1065, opened 2024-08-26
Export PDF through print.

ruibinx:
This resolves https://github.com/silverbulletmd/silverbullet/issues/635, sort of...

Usually when I export a page to PDF, I print the current page. This doesn't work in Silverbullet since panels are embedded in iframes. In Firefox, using default print will only print current screen(as shown below).

SCR-20240826-tzjt-2

This PR tries to solve this with a few lines of hacks.

SCR-20240826-ualn-2

However I'm not sure if this is the best way to solve the problem, so I set this PR to draft. Here are a few thoughts.

1. This PR will add a Printer button to all panels. Though it doesn't bother me , I don't think this behavior is expected by everyone. A better way could be adding a command "Markdown: Print" in markdown plug. But DOM is not accessible in web workers. Besides I need to find a way to pass around the ref to panel.
2. Another way is to create a plugin and use some third-party html2pdf libs like https://github.com/parallax/jsPDF. 
- Pros: we can have more control over what could be exported. For example, not exporting internal links.
- Cons: involves importing third-party dependencies, which I'm not sure how it can be done in plugs yet. 
3. I noticed that you mentioned using pandoc in the issue.
- Pros: pandoc could support more formats.
- Cons: 
   1. Have to be supported by server side. 
   2. I think there is no simple way to add native dependencies to silverbullet besides modifying Dockerfile?

Could you give me some suggestions? Thanks!

### DIFF
--- a/plugs/markdown/preview.ts
+++ b/plugs/markdown/preview.ts
@@ -53,18 +53,16 @@ export async function updateMarkdownPreview() {
 }
 
 function renderToolbar(): string {
-  const printerIcon =
-    `<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24"
-        viewBox="0 0 24 24" fill="none" stroke="currentColor"
-        stroke-width="2" stroke-linecap="round" stroke-linejoin="round"
-        class="feather feather-printer">
-      <polyline points="6 9 6 2 18 2 18 9"/>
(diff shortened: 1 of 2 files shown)


zefhemel:
This looks promising, but adding this to every panel doesn't make much sense. Can you not add this button just to the markdown preview panel inside the rendered HTML inside the iframe?

ruibinx:
> This looks promising, but adding this to every panel doesn't make much sense. Can you not add this button just to the markdown preview panel inside the rendered HTML inside the iframe?

Done. Please check.

zefhemel:
Nice!

Issue #1000, opened 2024-07-29
Feature request: index wikilinks in frontmatter


Maarrk:
Described in ["Live" Wiki Links in Frontmatter - General - SilverBullet Community](https://community.silverbullet.md/t/live-wiki-links-in-frontmatter/760) Community post.

Could be implemented by indexing text content of `FrontMatter` nodes, in a way similar to template code blocks:
https://github.com/silverbulletmd/silverbullet/blob/75471fa86b8eeac62685ed8e6355aa68737b2d4d/plugs/index/page_links.ts#L128-L140

I also run into it previously, I wanted to get book authors this way, but backlinks wouldn't work from frontmatter. I think it's expected to work, since there already is page link autocompletion inside frontmatter now.

Should the page ref get a `inFrontmatter` field like `asTemplate`?

zefhemel:
Sounds good. Question is how to detect wiki links. I'd suggest we follow the convention used in eg SETTINGS of expecting a `"[[link]]"` YAML value, you also get page completion that way. Note that values without those quotes are parsed by YAML as doubly nested links which is messy.

Maarrk:
I believe it's closed with #1066

Issue #736, opened 2024-02-22
Autocomplete on #tags and in frontmatter tags:


henrikx:
There should be autocompletion of tags while typing # for an inline tag or while writing tags in the frontmatter.

gorootde:
There actually is. Tags (once used anywhere in your space) are autocompleted after typing `#` + the first letter. In frontmatter you may use `tags: "#template"` to get auto competition. This only works when using double quotes.

zefhemel:
Would be nicer to get it without the # though. I agree.

henrikx:
Ah, I can see the autocomplete now. The reason why I thought it wasn't there was because I am running into some issues with it:

I like using tags with parents, e.g. `#test:testparent`, but this doesn't work with autocomplete. See:
![image](https://github.com/silverbulletmd/silverbullet/assets/10342989/8877a854-0512-4464-bb35-17a43d58bfd4)

Maarrk:
Another request from a friend who uses tags intensely would be to have a key to manually trigger tag completion like `Ctrl+Space` that shows suggestions after just # to avoid trying #a #b #c...

I'd say that it shouldn't always autocomplete after first # because it would pop out when typing headers.

---
Maybe triggering the hint with a key would make it simpler to invoke inside frontmatter. This could also be used anywhere in text, like let's say I have a word and I want to turn it into a similar tag that already exists, but I don't want to retype it all

PR #1066, opened 2024-08-27
Index wikilinks in frontmatter strings

Maarrk:
This implements #1000

The link is indexed only if it is the full content of a string node (allowing for quotes and whitespace) to discourage writing essays in the frontmatter :smile:

I added an explanation why the quotes are needed on YAML page, I can't think of a page to document that the feature exists at all, I think it's discoverable by looking at `indexPage` in SETTINGS

### DIFF
--- a/plugs/index/page_links.ts
+++ b/plugs/index/page_links.ts
@@ -1,4 +1,5 @@
 import {
+  collectNodesOfType,
   findNodeOfType,
   renderToText,
   traverseTree,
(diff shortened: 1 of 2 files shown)


Maarrk:
Short demo of what it allows, credit to [Raphwriter](https://community.silverbullet.md/t/live-wiki-links-in-frontmatter/760)

Zrzut ekranu z 2024-08-27 13-31-38

Zrzut ekranu z 2024-08-27 13-33-07

I'm starting to think they should be clickable after all. For that the cleanest approach would be to also recognise `WikiLink` nodes inside the frontmatter parse tree right? Should it include `space-config` as well?

Maarrk:
Would be nice to make them clickable, but I think that becomes a much larger feature, since they shouldn't hide the `[[` `]]` like links in normal text do when cursor is away. Same for tags in frontmatter.

Considering #736 this would be almost like another editor mode when inside frontmatter: different widgets, different autocomplete etc., I don't know how I would approach that. Hope this bit here helps

Maarrk:
Funnily enough, it did work in a small test space, but in my main one, I can't get the frontmatter mention to show up at all. I will look into it more these days, thanks for the heads up 👍

Maarrk:
Sorry for the delay, but the error was actually quite inconsistent to replicate. At the end I found it's a problem when reusing a Regex object with a global flag for multiple `.exec()` calls.

References:
https://stackoverflow.com/a/4724920/8531075
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex

---

There are multiple places in the codebase with `someImportedRegex.exec(text)`. When the const object is actually created is not obvious (at least for me). It's possible that every call site like this should have `someImportedRegex.lastIndex = 0` before, for general reliability improvement. Recreating the Regex everytime makes more sense, but is probably slower.

zefhemel:
Seems to work as advertised now. Thanks!

zefhemel:
And interesting that these regexes are indeed stateful, I don't like that. It may indeed be better not to define them as globals in that case... Labeling them as `const` makes this even more confusing.

PR #1099, opened 2024-09-28
Convert bullet to task when Task: Cycle State is called

### LINKED ISSUE #1074: Create new task if none exists at cursor during "Task: Cycle State"
In Roam Research, the "cycle task" action converts the current outline item into a task if none currently exists.

This is rather convenient behavior, and it would be nice to have it in SB.

meain:
Fixes #1074


zefhemel:
Thank you!

PR #1102, opened 2024-09-29
Fixes #1101

### LINKED ISSUE #1101: `{{notes}}` seems to be parsed as `{{not es}}`
In the following example `{{notes}}` outputs `false` instead of expected `Hello`. If `es: true` is changed to `es: false` then output is `true` which shows that `{{notes}}` seems to be parsed as `{{not es}}`.

````
---
es: true
notes: Hello
---


```template
{{notes}}
```

````

MrMugame:
Now uses the `kw<...>` rule everywhere, which should have been the case anyway. Probably legacy code.

(Fixes  #1101)


zefhemel:
Looks good, and I cannot see any regressions. Thanks!

PR #1110, opened 2024-10-08
Add settings to configure auto close brackets

fserb:
This PR creates a `autoCloseBrackets` setting that allows to configure which brackets to automatic close (or to set it empty to disable auto-closing)


zefhemel:
Nice!

PR #1116, opened 2024-10-14
Parse tags starting with digits and containing spaces (close #1072)

### LINKED ISSUE #1072: Allow tags to start with a numeral
Starting tags with a numeral kinda works, but the tag isn't highlighted in blue showing its recognised as a tag, and autocomplete doesn't work. Queries based on the tag still work though.
My use-case is that I'm following the filing system 'Johnny Decimal' which allows creating consistent filing across areas like notes, email, electronic folders, and physical folders. But this requires that the name starts with a numeric id.

Maarrk:
I'd suggest to continue discussion in #1072

### DIFF
--- a/plug-api/lib/tags.ts
+++ b/plug-api/lib/tags.ts
@@ -4,6 +4,7 @@ import {
   type ParseTree,
   traverseTree,
 } from "@silverbulletmd/silverbullet/lib/tree";
+import { tagRegex } from "$common/markdown_parser/constants.ts";
 
 export function updateITags<T>(obj: ObjectValue<T>, frontmatter: FrontMatter) {
   const itags = [obj.tag, ...frontmatter.tags || []];
@@ -26,7 +27,7 @@ export function extractHashTags(n: ParseTree): string[] {
   const tags = new Set<string>();
   traverseTree(n, (n) => {
     if (n.type === "Hashtag") {
-      tags.add(n.children![0].text!.substring(1));
+      tags.add(extractHashtag(n.children![0].text!));
       return true;
     } else if (n.type === "OrderedList" || n.type === "BulletList") {
       // Don't traverse into sub-lists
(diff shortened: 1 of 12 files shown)


plug-api/lib/tags.ts
```diff
@@ -4,6 +4,7 @@ import {
   type ParseTree,
   traverseTree,
 } from "@silverbulletmd/silverbullet/lib/tree";
+import { hashtagName } from "$common/markdown_parser/parser.ts";
```
zefhemel, line 7: The idea of `plug-api/*` is that these can be liberally imported from plugs without pulling in too much random code. In this case you're importing basically the entire markdown parser + all lezer parser code, not a good idea. Maybe you can move the `hashtagName` and related functions even to this very file (tags.ts) and import them into the markdown parser instead?

common/markdown_parser/parser.ts
```diff
@@ -594,6 +594,32 @@ const Hashtag = regexParser(
...
+    } else { // this is correct #<name>
+      return text.slice(2, -1);
+    }
+  } else { // this is just #name
+    return text.slice(1);
+  }
+}
+
+/** Get markup for a hashtag name with # prefix and angle brackets if necessary */
+export function hashtagText(name: string): string {
```
zefhemel, line 614: What about `renderHashtag`?

common/markdown_parser/parser.ts
```diff
@@ -594,6 +594,32 @@ const Hashtag = regexParser(
   },
 );
 
+/** Extract the name from hashtag text, removing # prefix and <angle brackets> if necessary */
+export function hashtagName(text: string): string {
```
zefhemel, line 598: what about `extractHashtag`?

common/markdown_parser/parser.ts
```diff
@@ -594,6 +594,32 @@ const Hashtag = regexParser(
...
+    }
+  } else { // this is just #name
+    return text.slice(1);
+  }
+}
+
+/** Get markup for a hashtag name with # prefix and angle brackets if necessary */
+export function hashtagText(name: string): string {
+  // detect with the same regex as the parser
+  const simple: string = "#".concat(name);
```
zefhemel, line 616: What about simply `const simple = "#" + name"`?

zefhemel (changes requested):
Added a few comments, but looking forward to get this in!

common/markdown_parser/parser.ts
Maarrk: Changed. I had had issues with `+` for strings in typescript, still don't know where I can and can't use it.

plug-api/lib/tags.ts
Maarrk: That's an area in which I have very little experience, and I'm glad to learn. Now everything is importing `markdown_parser/constants.ts` instead. Suddenly it makes sense why this module exists :smile:

plug-api/lib/tags.ts
```diff
@@ -4,6 +4,7 @@ import {
   type ParseTree,
   traverseTree,
 } from "@silverbulletmd/silverbullet/lib/tree";
+import { tagRegex } from "$common/markdown_parser/constants.ts";
```
zefhemel, line 7: The problem we have now is that this means that `$common` needs to be defined in any plug relying on this file, which will be the case for all the builtins, but not for external plugs. Two options: either move the constant into plug-api/lib (again: I think this very file tags.ts would be fine) or use a relative path: `../../common/markdown_parser/...` instead.

zefhemel:
Ok you know what, let me just merge this as it and fix it on main.

Maarrk:
OK, thanks! I'll write down the changes I needed to do here for future PRs

zefhemel:
I hope you're taking a note in SilverBullet.

PR #1121, opened 2024-10-17
Space config for smart quotes (close #1114)

### LINKED ISSUE #1114: Add configuration for smart quotes
First of all, thank you for this wonderful program! It's been a joy to use. On to the issue:

It's common for non-English languages to use quotation marks different from the English ones. For example, in German, you would usually quote like this:

„Hast“

In French, like this:

« mot »

And so on. It would be useful if there was a setting to configure custom left and right quotation marks when using smart quotes.

Maarrk:
Also deprecate `useSmartQuotes` in favour of `smartQuotes.enabled` for more logical hierarchy

### DIFF
--- a/type/client.ts
+++ b/type/client.ts
@@ -39,3 +39,9 @@ export type ActionButton = {
 export type EmojiConfig = {
   aliases: string[][];
 };
+
+export type SmartQuotesConfig = {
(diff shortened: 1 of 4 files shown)


zefhemel:
Awesome, thanks!

Issue #1042, opened 2024-08-18
Migrate PLUGS to space config


zefhemel:
Having a separate PLUGS file for maintaining what plugs to activate made sense at the time. But by now we have a pretty robust alternative with space config and should just move to that.

This can be done in a backwards compatible way by supporting both, with space config becoming the new way. 

To figure out: how to implement Plugs: Add, which currently updates your PLUGS file automatically.

nodu:
Perhaps drop the Plugs: Add command... ?

Maarrk:
Proposal:
1. Keep the plugs list in Space Config as:
~~~
```space-config
plugs:
  - github:silverbulletmd/silverbullet-git/git.plug.js
  - ghr:Maarrk/silverbullet-grep
```
~~~
2. `Plugs: Add` appends to this list *on the current page*
  - if there is no `space-config` block, add it at the end of the page
  - if the `space-config` block doesn't have `plugs:` at the top level, append it at the end of the block
  - append the URI into the list
 3. If the command is ran on the PLUGS page, it changes the `yaml` block with a list into the `space-config` block with the list under `plugs:` - this way the update happens automatically for the user, but in front of their eyes

---
  
Potential problem: there is more ways to write a list in YAML
  
~~~
```space-config
plugs: [ github:silverbulletmd/silverbullet-git/git.plug.js, ghr:Maarrk/silverbullet-grep ]
```
~~~
  
But I think this can also be handled reasonably if the inserting is done on the YAML as a parsetree instead of just adding a line

zefhemel:
I think this makes sense. The only nasty thing is you probably want to update YAML without rewriting it (e.g. so that comments get lost, reformat everything). For the PLUGS file the whole plug list was always rewritten when using `Plugs: Add` I think, but in the new system I wouldn't want to happen. It should somehow attempt a minimal change.

Maarrk:
New idea: the `Plugs: Add` command accepts a string argument that fills the `editor.prompt`, there can be a button `{[Plugs: Add|Install now](https://example.com/cool.plug.js)}`. The prompt would still show the user the address before adding it, so I think it's OK

Because of that, I'm back to the idea that the command should edit the PLUGS page, even if they can be loaded from anywhere, updated the https://github.com/silverbulletmd/silverbullet/issues/1042#issuecomment-2413192106

Maarrk:
Finally got around to finishing this in #1122

[...]

PR #1122, opened 2024-10-17
Move Plugs to Space Config and rework "Plug: ..." commands

Maarrk:
This is implementing #1042 with some minor improvements on the side.


zefhemel:
Ok this looks good, some minor issues but I'll fix those after merge. Thanks!

PR #1127, opened 2024-10-23
Enabler to make scripts loaded from CDN more secure

gorootde:
This adds the possibility to set an integrity hash for the script tag that is e.g. used by the mermaid plug. If properly used, this will prevent manipulated CDN sources from being loaded into silverbullet.


zefhemel:
Nice!

Issue #163, opened 2022-12-05
Support hosting at a URL path


activeshadow:
I'm wanting to put an instance of `silverbullet` behind an Nginx proxy at a particular path. For example:

```
location ~ /silverbullet(.*) {
  proxy_pass http://127.0.0.1:3000$1;
}
```

However, this does not work beyond the main index page since it attempts to download assets without prepending the `/silverbullet` path to the requests.

Is there a way to configure `silverbullet` or Deno to use a URL path (or recompile `silverbullet` with a URL path)?

I know I could just host `silverbullet` at a subdomain but for this particular situation using a subdomain is not an option.

zefhemel:
Not possible right now, but should not be very hard to add.

activeshadow:
Okay @zefhemel thanks for the reply. I'll see if I can't poke around and make something work then submit a PR.

zefhemel:
I forgot about the server side of course. All routes are defined in http_server.ts so that should be easy to find.

ryanwwest:
Following, I was having troubles getting this to work as well and thought it was just nginx, but it's probably this.

billbrod:
Wanted to check in on this -- `silverbullet` looks really cool and I'd be interested in putting it behind Nginx as well.

zefhemel:
The way I work around this myself is to simply put things on a separate (sub) domain. You probably considered this already, but just saying.

gardiol:
While i indeed too use silverbullet on a subdomain, i think having the option to host on a sub-path is a really needed feature. Not everyone want/can spin subdomains. For example, i would want to host silverbullet at work but obtaining subdomains trough corporate is impossibile so i would need to host on the one we already have, not cannot do that on root as well (no write access to root).

Shihira:
Hello everyone, I've been waiting too long for this feature and therefore I tried to work on this url prefix support on my branch, and it's starting to work on my own builds. Check out the commit in my fork https://github.com/silverbulletmd/silverbullet/commit/b1de4761169f0344e565086c2dbad5c3d9fd5717 if you please. If it's good enough for you to accept it into the upstream branch, I would be pleased to submit a PR. It's nearly a certain thing that there are edge cases that I didn't test out, or there might be lint problems, but as a normal user I've been doing my best.

gardiol:
Thanks @Shihira this is greatly appreciated and i hope that @zefhemel will seriously consider merging your PR

zefhemel:
[...]

PR #1131, opened 2024-10-27
Feature: URL base/URL subpath support

Shihira:
As discussed in issue https://github.com/silverbulletmd/silverbullet/issues/163 , the base URL feature is required to host silverbullet through reverse proxy without the need of subdomain, enabling the usage of some port proxy such as ngrok.

In this implementation, the url prefix is read from environment variable `SB_URL_PREFIX`, and passed to the client by text substitutions in `index.html`.

I did many coarse batch substitution in like `http_sever.ts` without fine tests, so it might need to be carefully looked into.


Shihira:
@zefhemel lint problems should be fixed in the appended commit. Restart the check will you please?

zefhemel:
Good start, testing this right now let me post my findings:

1. While clicking page links works, the URL they _appear_ to link to is still on the root (note the URL when you hover over them), likely you'll have to adapt `cm_plugins/wiki_link.ts` too.

zefhemel:
I noticed that I _have_ to prefix my prefix with a `/` otherwise it doesn't work (I tried `SB_URL_PREFIX` with `website` but then found only `/website` works). This ok, although it may be nice to sanitize this (just add a `/` if it's not there already)

zefhemel:
When I transclude an image or PDF, it renders nicely inline, but when I click the transclusion link it links me to a 404. For instance:

CleanShot 2024-10-28 at 14 37 40@2x

but clicking that link:

CleanShot 2024-10-28 at 14 37 50@2x

zefhemel:
Sync mode doesn't seem to work at all (try switching to it by using the Sync button at the top right)

Shihira:
It seems there is still a long way to go before completely usable. Do I need to work on this alone or the project members will also provide assistance?

Shihira:
I think I am doing it wrong. In most cases urls need to remain relative instead of getting a prefix injection everywhere. I will look into it by weekend.

Shihira:
@zefhemel  the latest commit should have fixed most of the url jumping problems you metioned above. However I still don't have a clue to the sync mode problem, it's a little bit too complicated to me. For some reason the file syncing didn't started on sync mode, and none of the plugs were loaded either, but it just failed silently without giving me a single hint as to where I can get started. I need some help from the silverbullet members.

Shihira:
@zefhemel the latest commit should have fixed most of the url jumping problems you metioned earlier. And the sync mode appeared to function out of box after I applied the latest upstream update (without any modifications on my end).

Shihira:
Yes please take over

flowkclav:
Since I also need the sub-path feature, I decided to take a crack at it.

I looked at the changes from this @Shihira PR for inspiration, but I changed the approach a bit. Most notably on the client side, the prefix is no longer passed from server to client in the config, but rather `document.baseURI` is used (since it factors in the `<base .../>` element) instead of location.host|origin.

I think I addressed most of the issues raised in this PR (assuming the "Navigate: To URL" command is no longer a thing in v2) and a few other I found while experimenting (like setting the path of the auth cookies according to prefix to avoid polluting the namespace of "root" ). I did end up not touching `common/spaces/http_space_primitives.ts` which is touched by this PR, but didn't experience any errors...

I did my work from the current `main` branch, i.e. on v2.  @zefhemel You are probably busy with v2 transition stuff, but I can submit a PR for this if you want.   If you want to peek my changes: https://github.com/flowkclav/silverbullet/commit/bed4fb6f47df02daf45df763f8d03d9fa9aa39ee

PR #1138, opened 2024-11-01
Fix build on windows

p-weston:
This fixes the build on windows, where `new URL("deno.json", import.meta.url).pathname` starts with `/C:/` which isn't a valid path.

First time with deno, so not sure this is idomatic.

### DIFF
--- a/build_plugs.ts
+++ b/build_plugs.ts
@@ -27,7 +27,7 @@ if (import.meta.main) {
       debug: args.debug,
       reload: args.reload,
       info: args.info,
-      configPath: new URL("deno.json", import.meta.url).pathname,
+      configPath: path.resolve("./deno.json"),
     },
   );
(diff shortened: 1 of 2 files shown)


p-weston:
Updated to use fileURLToPath which looks like a better fix.

zefhemel:
Ok indeed, your previous suggestion actually changed behavior but this looks good. Not a huge fan of starting to rely on node.js APIs, but for builds I think it's ok. Let me quickly check if this doesn't break anything on other platforms.

zefhemel:
Looks good. Thanks!

v-shenoy:
Still new to Deno, but I think this could have been done with [toFileUrl](https://jsr.io/@std/path/doc/~/toFileUrl) to avoid dependency on the Node API.

PR #1139, opened 2024-11-02
recognise tel: uri scheme as a non-local path

### LINKED ISSUE #1095: The `tel` URI scheme for telephone numbers not recognized in hyperlinks
The [RFC 3966](https://datatracker.ietf.org/doc/html/rfc3966) [`tel`](https://datatracker.ietf.org/doc/html/rfc3966#section-3) URI scheme is not recognized in hyperlinks. Desktop browsers usually do not know to handle it but that is not the case for mobile browsers where it usually fires an appropriate action.

```markdown
[Call](tel:+1234567890123)
```

When clicked then creating of page called `tel:+1234567890123` is requested instead of default browser action. I am not sure of the real cause of the behaviour, but could that be fixed, please? *Sidenote:* I haven't tried other URI schemes not I know what Markdown spec tells about this...

p-weston:
Fixes #1095 The tel URI scheme for telephone numbers not recognized in hyperlinks

This could cover all possible protocols like this:
```
return URL.parse(path)===null;
```
but that would block **lots** of page names containing a colon.

### DIFF
--- a/plug-api/lib/resolve.ts
+++ b/plug-api/lib/resolve.ts
@@ -52,7 +52,9 @@ export function isFederationPath(path: string): boolean {
 }
 
 export function isLocalPath(path: string): boolean {
-  return !path.includes("://") && !path.startsWith("mailto:");
+  return !path.includes("://") &&
+    !path.startsWith("mailto:") &&
+    !path.startsWith("tel:");
 }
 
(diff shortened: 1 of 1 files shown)


zefhemel:
Thanks!

Issue #976, opened 2024-07-24
Bug report: vim search mode would leak delete key


LumenYoung:
As suggested in the title, I find this potential bug, and it is easy to reproduce on my side

1. enable vim mode
2. type `/` to enter search mode
3. enter random characters and then backspace to simulate a deletion

Expected behavior: the character in search should be delete

Current behavior: the searched content is not deleted, rather the character under the cursor is deleted.

SB Version: 0.8.1

erinsummer:
I'm able to reproduce this in 0.9.0 as well.

I'm seeing the issue any time the cursor is supposed to be focused in the vim command line. I thought perhaps it was related to one of my (many) browser extensions, but after disabling them all the backspace focus shift persists. (And then I was unable to continue testing as my installation seemed to break fairly badly, leaving some pages inaccessible/unwritable, though I don't know that it's related?)

Specifics:

- SB: 0.9.0
- Debian GNU/Linux 11 (bullseye) 5.10.221-1
- via SSL/nginx (no proxy)

1. enable vim mode
2. use a vim motion (select some text, start a search, anything to open the command editor). I was trying to do a find/replace using `:s%`, and kept making typing errors. Arrow keys, backspace, and maybe some other keys seem to be sent back to the editor rather than the vim command line
4. Escaping out of the vim command leaves one with whatever changes were made by the keystrokes being passed through to their document instead of their vim command prompt

Expected: send the keystrokes to the (focused) vim command prompt rather than the main editor itself

meain:
Probably related. I've had cases where when I press G to go to the last line in normal mode, it would just enter that into the editor.

lutzky:
Silverbullet is based on codemirror, but this demo doesn't have this bug: https://codemirror.net/5/demo/vim.html. That demo is for an older version though.

LumenYoung:
Thanks a lot @lutzky  for the long waited fix! I've tested the latest commit and it works on my instance.

But one thing I still find not really reasonable for a long term vim user is **the del key in the normal mode** still deletes in the original text instead of just moving backward. I think it is pretty annoying and not intended for vim usage. 

Do you see a way to make this behavior consistent with vim's behavior? Technically it doesn't belongs to this bug but I would assume these are relevant and could be fixed with similar manner.

lutzky:
[...]

PR #1145, opened 2024-11-09
Do not use standardKeymap in vim mode

lutzky:
The standard keymap conflicts with the vim mode, e.g. not allowing arrow keys to move through the vim commandline for `/` and `:` commands. The keys described in
https://codemirror.net/docs/ref/#commands.standardKeymap seem to work just fine without it in vim mode.

Fixes https://github.com/silverbulletmd/silverbullet/issues/976

### DIFF
--- a/web/editor_state.ts
+++ b/web/editor_state.ts
@@ -367,7 +367,7 @@ export function createKeyBindings(client: Client): Extension {
     ...createCommandKeyBindings(client),
     ...createSmartQuoteKeyBindings(client),
     ...closeBracketsKeymap,
-    ...standardKeymap,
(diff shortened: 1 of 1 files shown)


zefhemel:
Alright, let's try, people will scream if this breaks stuff :)

lutzky:
Uhh... screamer here. I don't know how I missed this, but this breaks the Return key for ordinary paragraphs (it does nothing in Insert mode now). Super weird. Workaround is hitting `o` in Command mode instead; seeing as the previous version was destructive, this seems less bad, but still worth fixing. I'll take a look.

lutzky:
I'll move discussion back to #976.

PR #1146, opened 2024-11-11
Fix Enter for vim mode

lutzky:
This is a workaround for https://github.com/replit/codemirror-vim/issues/182

Fixes https://github.com/silverbulletmd/silverbullet/pull/1145 (harder)


LumenYoung:
Marvelous! When would zef push another edge image including this commit?

zefhemel:
It's already on edge within minutes of a merge. I'll do a proper release in the next few days.

PR #1150, opened 2024-11-12
Allow tab press to autocomplete

v-shenoy:
Accept auto-complete suggestions on tab press similar to code editors. Falls back to indent if no autocompletion context.


zefhemel:
Nice. Thanks!

Issue #1133, opened 2024-10-27
Protection against brute force attacks


Archie-2021:
Add a limit to number of failed login attempt in a shor period of time.

Thanks

zefhemel:
This has now been implemented.

PR #1152, opened 2024-11-12
Simple protection against brute force attacks

p-weston:
A **very** minimal solution for #1133 Protection against brute force attacks. Lots of little decisions, so let me know if they don't suit.

This kind of thing makes it easy to DoS a server, so it defaults off.  Configured with two new environment variables:
SB_LOCKOUT_LIMIT - Maximum number of login fails before lockout
SB_LOCKOUT_TIME_MS - Time to count fails over in ms

If this is OK, where should it be documented, and should there be a command line flag for it?

### DIFF
--- a/server/http_server.ts
+++ b/server/http_server.ts
@@ -20,6 +20,7 @@ import {
   parsePageRef,
 } from "@silverbulletmd/silverbullet/lib/page_ref";
 import { base64Encode } from "$lib/crypto.ts";
+import { LockoutTimer } from "./lockout.ts";
 
 const authenticationExpirySeconds = 60 * 60 * 24 * 7; // 1 week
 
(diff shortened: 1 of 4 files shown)


zefhemel:
This looks reasonable to me, but what do you think about enabling it by default with sensible defaults? I think I prefer to be on the safe side by default and let people disable or alter if they like. 

I think a command line arg version isn't necessary, I think we can slowly migrate to primarily using env variables for this. 

If you could add these env variables to the website/ docs that'd be great and I'd be happy to merge. 

Thanks for the effort!

zefhemel:
You know what, I'm quite eager to get this in so let me make some final adjustments myself.

zefhemel:
Merging this, and adding my changes on top in a bit.

zefhemel:
Made more changes here: 51fc5952bcc4e36533fc7f5523bd09e043f3ffc6

PR #1158, opened 2024-11-17
Implement native sharing

### LINKED ISSUE #1090: Feature request: Share target
As an Android user, I sometimes encounter something I'd like to send directly to silverbullet; a URL of a page I'm seeing, comment contents somewhere, etc. Android has a Share button that allows sending this to various apps - e.g. Obsidian supports adding this to the Daily note.

[This appears to be possible for PWA](https://developer.chrome.com/docs/android/trusted-web-activity/web-share-target).

As an immediate use case, I'm thinking of sending things into a new "Quick Note". Future customization might be interesting as well.

nightscape:
Closes #1090 

- [x] Call `handleShareTarget` method when native sharing button is used
- [x] Return HTML to the client
- [x] Store shared data in a target page
- [x] Redirect user to target page
- [x] Make target page for shared data configurable
- [x] Allow using template variables in page name
- [ ] Make it work in offline mode

### DIFF
--- a/plugs/share/share.plug.yaml
+++ b/plugs/share/share.plug.yaml
@@ -12,6 +12,12 @@ functions:
     events:
       - share:options
 
+  handleShareTarget:
+    path: share.ts:handleShareTarget
+    env: client
+    events:
+      - http:request:/share_target
+
   clipboardMarkdownShare:
     path: share.ts:clipboardMarkdownShare
     events:
@@ -31,3 +37,10 @@ functions:
     path: publish.ts:publishShare
     events:
       - share:publish
+
+config:
+  # Built-in configuration schemas
+  schema.config.properties:
+    shareTargetPage:
+      type: string
+      format: page-ref
--- a/plugs/share/share.ts
+++ b/plugs/share/share.ts
@@ -2,6 +2,7 @@ import {
   editor,
   events,
   markdown,
+  space,
   system,
 } from "@silverbulletmd/silverbullet/syscalls";
 import { findNodeOfType, renderToText } from "../../plug-api/lib/tree.ts";
@@ -11,6 +12,11 @@ import {
   encodePageURI,
   parsePageRef,
 } from "@silverbulletmd/silverbullet/lib/page_ref";
+import type { EndpointRequest } from "@silverbulletmd/silverbullet/types";
+import { localDateString } from "$lib/dates.ts";
+import { cleanPageRef } from "@silverbulletmd/silverbullet/lib/resolve";
+import { builtinFunctions } from "$lib/builtin_query_functions.ts";
+import { renderTheTemplate } from "$common/syscalls/template.ts";
 
 type ShareOption = {
   id: string;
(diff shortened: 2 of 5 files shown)


nightscape:
Oh, the way it is currently implemented does not work in offline mode, which makes it 80% less useful...

plugs/share/share.plug.yaml
```diff
@@ -12,6 +12,12 @@ functions:
     events:
       - share:options
 
+  handleShareTarget:
+    path: share.ts:handleShareTarget
+    env: client
+    events:
+      - http:request:/share_target
```
nightscape, line 19: @zefhemel I guess just specifying `env: client` here is not enough to trigger `handleShareTarget` through an incoming http request, right?
Is there already logic to delegate `http:request` events to plugs on the client side?
It looks like this would need to be done in `service_worker.ts` which currently only handles read requests.
Do you think it's a good idea to enhance `service_worker.ts` with `dispatchEvent` logic similar to `http_server.ts`?

nightscape: @zefhemel ping 😉

zefhemel: RIght. Right now handling HTTP requests through a service worker (so fully locally and therefore offline) is not possible. It's technically possible to do, but it's not there at this moment.

zefhemel:
Sorry for the huge delay of this.

I like this feature, but I'd implement it slightly differently, if this would work (I don't have an environment to test this on, because I think this is an Android feature only?) I'd implement the endpoint handling "hardcoded" in the service worker. That is: in the `fetch` event handler, decide on some URL (like `/.share` or something), and put the logic in the service worker directly rather than in the share plug. I feel this is functionality that deserves a place directly in the SB core, not so much in a plug (it needs to be hardcoded in the manifest anyway), and implemented in the service worker it can also be done (I think) in a way that works both in online and offline mode.

zefhemel (changes requested):
Sorry for the huge delay of this.

I like this feature, but I'd implement it slightly differently, if this would work (I don't have an environment to test this on, because I think this is an Android feature only?) I'd implement the endpoint handling "hardcoded" in the service worker. That is: in the `fetch` event handler, decide on some URL (like `/.share` or something), and put the logic in the service worker directly rather than in the share plug. I feel this is functionality that deserves a place directly in the SB core, not so much in a plug (it needs to be hardcoded in the manifest anyway), and implemented in the service worker it can also be done (I think) in a way that works both in online and offline mode.

zefhemel:
A head's up: working with service worker is a bit of a pain in terms of reload behavior. In Chrome, you can have the service worker update on every reload:

CleanShot 2024-12-14 at 09 52 48@2x

Still you probably need to reload your page a few times after every change to make sure you got the new service worker running before deciding things don't work ;)

plugs/share/share.plug.yaml
nightscape: Would `service_worker.ts` be the right place? Do you see any issues with using a similar approach as on the server-side, i.e. checking the plugs if one of them wants to handle the URL?

PR #1163, opened 2024-11-21
Add "Open with Codeanywhere" badge to README.md

nkkko:
## Why?

With Codeanywhere, devs and contributors can:

- Instantly launch a cloud or local IDE with remote dev environment.
- Automatically set up and configure a remote dev environments if a `devcontainer.json` file is present.

## Benefits:

1. Faster onboarding for new contributors.
2. Consistent and ready-to-use development environments.
3. Flexibility to work in the cloud or locally.


zefhemel (changes requested):
Just tried this out and it launches into the editor, but the environment is not setup. There's no deno for instance. As you suggest this would probably be set up with a devcontainer.json, but we don't currently have one, could you add that too? For the gitpod setup everything works in one go.

Issue #1160, opened 2024-11-20
Naked URL links shouldn't include final period/full-stop


lutzky:
Input:

    I like https://silverbullet.md/SilverBullet. It's great.

Desired output - period outside link

    I like <a href="I like https://silverbullet.md/SilverBullet">https://silverbullet.md/SilverBullet</a>. It's great.

Actual output - period inside link.

    I like <a href="https://silverbullet.md/SilverBullet.">https://silverbullet.md/SilverBullet.</a> It's great.

This occurs both in the live editor and in markdown preview. Notably, that's a [broken link](https://silverbullet.md/SilverBullet.).

Github gets this right:

> I like https://silverbullet.md/SilverBullet. It's great.

So does https://markdownlivepreview.com/ and https://dillinger.io/. SilverBullet actually gets this right for commas, but not for periods.

lutzky:
Issue is likely here:

https://github.com/silverbulletmd/silverbullet/blob/51fc5952bcc4e36533fc7f5523bd09e043f3ffc6/common/markdown_parser/parser.ts#L576-L585

lutzky:
Workaround: Use the proper commonmark style, `<https://silverbullet.md/SilverBullet>.` does the right think (in github too. Notably, commonmark specifies that "truly naked" URLs (without `<>`) [don't count as links](https://spec.commonmark.org/0.31.2/#autolinks).

stapelberg:
> SilverBullet actually gets this right for commas

No, I don’t think it does, actually: URLs are cut off at comma or semicolon characters, for example the following URL does not work in SilverBullet:

https://cs.opensource.google/go/go/+/master:src/archive/zip/reader.go;l=25;drc=9cfe3a86d34f7f4a401dae9a22389b12f7e8bb2e

Renders as:

<img width="808" alt="Screenshot 2024-11-20 at 23 20 55" src="https://github.com/user-attachments/assets/ce602790-73da-46c6-9d6d-49512771b0fe">

There might of course be more trigger characters I have not yet encountered in my usage :) We should probably locate the relevant RFC.

One heuristic I thought about is for SilverBullet to recognize period/comma **followed by whitespace** as not part of the URL. Though of course that breaks URLs which require a trailing comma. I am not sure how common these are, but the safest solution of course is to not use any heuristic and require users to be explicit when the behavior isn’t what they expect.

lutzky:
(Yo @stapelberg! Fancy meeting you on a bugreport)

[...]

PR #1175, opened 2024-12-04
Fix trailing periods for naked URLs

lutzky:
Fixes https://github.com/silverbulletmd/silverbullet/issues/1160


zefhemel:
Thanks for doing this (with tests)

PR #1177, opened 2024-12-06
doc: replace Oak mentions by Hono

Delapouite:
Hi

Oak was replaced by Hono in commit 291280b

Thanks


zefhemel:
Thanks. Meant to do this, but kept forgetting.

Issue #938, opened 2024-07-12
Render #tag in pages with HTML anchor, not just span, for accessibility


jamescooke:
## Current behaviour

When I write a `#meta` tag in a Page body, it's rendered as a nice "blobby" hashtag link. That link directs me to `http://localhost:3000/%F0%9F%93%8C%20meta` when I click on it with my pointer.

However, I'm not a mouse / pointer user - I interact with my browser via the keyboard. When that hashtag is rendered it appears to just be a `<span>` (I assume affected by some magic Javascript to make it clickable).

Here's a screenshot of a `#meta` tag in the page and the inspector showing it as a `<span>` below:

![Screenshot from 2024-07-12 14-02-57](https://github.com/user-attachments/assets/06e0866d-3d81-439f-8486-5d9710287c7f)

Other links on the Page, such as those using `[ ]()` and `[[ ]]` notation _are_ rendered with `<a ...>` HTML anchors and _are_ easily followable with my keyboard interface.

## Expected (requested) behaviour

Please can `#meta` style tags in Pages be rendered in an HTML anchor `<a ...>` so that they're easily discoverable and followable for non-pointer users. 🙏🏻

terr-steak:
@jamescooke changes were merged into main ([this is the PR if interested](https://github.com/silverbulletmd/silverbullet/pull/1179)) that I think should address this. If you want to give it a spin in the [nightly builds](https://community.silverbullet.md/t/living-on-the-edge-builds/27) or after the next release and let me know how you feel about them I can update as needed or mark this as done. Furthermore if you see any other issues or places that accessability can be improved further keep creating tickets and feel free to tag me as I'd like to make sure that SB works well for screen readers too!

jamescooke:
Hi @terr-steak I've tried out the `:edge` tagged Docker image with index digest `sha256:7d1482ae132...`

The good news is I _can_ now focus on `#tag` elements 👍🏻 .

The problem is it can't follow the link itself - or copy its URL to the clipboard. It looks to me like the `#tag` links are still different to "normal" page links - the main difference is when I hover them, nothing appears in the browser info bar and no tooltip pops up. However, hovering a "normal" page link makes the URL show in the info bar and a tooltip of "Navigate to [PAGE]" appears.

It's my understanding that that additional info (the URL into the info bar) is what's used when doing any accessibility.

terr-steak:
[...]

PR #1179, opened 2024-12-10
Update how tags are rendered to allow for custom styling 

terr-steak:
* Wrapped the existing span for a tag with an anchor element <a> to
improve ux for screen readers (#938) 
* Added data-tag-name attribute to tags so that given a tag `#my-tag` it
will have the attribute `data-tag-name="my-tag"` added to the new `<a>`
element wrapper allowing for end users to target tags for styling with a
css selector such as `.sb-hashtag[data-tag-name="my-tag"]{...}`
* Updated docs (ie silverbullet.md docs)

Related community threads:
- https://community.silverbullet.md/t/tag-styling/1022
- https://community.silverbullet.md/t/tag-styling-improvements/1254

### DIFF
--- a/common/markdown_parser/parser.ts
+++ b/common/markdown_parser/parser.ts
@@ -584,15 +584,13 @@ const NakedURL = regexParser(
   },
 );
 
-const Hashtag = regexParser(
-  {
-    firstCharCode: 35, // #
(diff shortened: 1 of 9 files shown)


zefhemel:
Seems to work as advertised. Thanks a lot!

PR #1183, opened 2024-12-15
Fix Swift syntax highlighting

davidchiles:
Fixed swift syntax highlighting and updated example in website as well.


zefhemel:
Hah. Thanks!

PR #1186, opened 2024-12-16
Fix encoding of inline content

### LINKED ISSUE #1182: Images with colons in their name not displayed
1. In Windows I select a png file in the file manager and Copy it using context menu
2. I paste it into Silverbullet (shift-insert)
3. silverbullet asks for a filename
4. if i keep the filename as-is, the image appears. But if I change the filename and it includes a colon (`:`) in the filename, the link for the image is inserted into the document correctly, but the image is not displayed.

image
image

Pistahh:
Inline content with the file containing e.g. a `:` cannot be displayed as the file name is set as the img src, and the browser treats the part before the : as the protocol, instead of as a file name.

This change encodes the file name so `:` becomes `%3A`, fixing the issue.

Fixes #1182.

### DIFF
--- a/web/cm_plugins/inline_content.ts
+++ b/web/cm_plugins/inline_content.ts
@@ -56,15 +56,17 @@ class InlineContentWidget extends WidgetType {
       return div;
     }
 
+    const url = encodeURIComponent(this.url)
+
     if (mimeType.startsWith("image/")) {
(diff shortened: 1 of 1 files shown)


zefhemel:
Awesome, thanks!

PR #1198, opened 2025-01-03
Prevent remote space script reload when syncOnly is enabled

mufeedali:
This also fixes a side effect where the library update status notification would never show up.


zefhemel:
Nice!

PR #1199, opened 2025-01-03
Correctly extract hashtag content in links and Page Picker filter

### LINKED ISSUE #1196: Hashtags rendered in templates create incorrect links to tag pages
Hashtags rendered in templates create incorrect links to tag pages (they include `< >` in the address). Related to #1072 and
Found by @mjf in https://community.silverbullet.md/t/my-first-hours-with-sb/743/9?u=maarrk, a minimal example below:

~~~
#tag1 #<second tag>

```template
{{#each tags}}
#<{{.}}>
{{/each}}
```
~~~

image

Probably need to change a `slice(1)` into `extractHashtag` (introduced by #1116) somewhere in template rendering chain.

Maarrk:
This fixes #1196

Also went through the code and looked at all instances of `replace("#`, `replace('#`, and `slice(1` to find other locations where `extractHashtag` should be used. I think now it should be good.

Also, is there a preference between imports starting with `../../plug-api` and `@silverbulletmd/silverbullet`? I see both throughout the codebase, and seems to me that Deno uses one or the other for automatic imports. I'd say that the second option is easier to copy into other files.

### DIFF
--- a/plugs/markdown/markdown_render.ts
+++ b/plugs/markdown/markdown_render.ts
@@ -15,6 +15,7 @@ import { Fragment, renderHtml, type Tag } from "./html_render.ts";
 import { isLocalPath } from "@silverbulletmd/silverbullet/lib/resolve";
 import type { PageMeta } from "@silverbulletmd/silverbullet/types";
 import * as TagConstants from "../../plugs/index/constants.ts";
+import { extractHashtag } from "@silverbulletmd/silverbullet/lib/tags";
 
 export type MarkdownRenderOptions = {
   failOnUnknown?: true;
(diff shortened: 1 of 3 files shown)


plugs/markdown/markdown_render.ts
```diff
@@ -15,6 +15,7 @@ import { Fragment, renderHtml, type Tag } from "./html_render.ts";
 import { isLocalPath } from "@silverbulletmd/silverbullet/lib/resolve";
 import type { PageMeta } from "@silverbulletmd/silverbullet/types";
 import * as TagConstants from "../../plugs/index/constants.ts";
+import { extractHashtag } from "../../plug-api/lib/tags.ts";
```
zefhemel, line 18: Yeah let's use @silverbullet/... style imports everywhere

zefhemel (changes requested):

zefhemel:
Looks good, but indeed let's use the @silverbulletmd/silverbullet style imports where possible

Maarrk:
Changed to `silverbulletmd/silverbullet` where possible, but not everything is exported and available this way, for example `import * as TagConstants from "../../plugs/index/constants.ts";` had to stay this way.

zefhemel:
Awesome!

Issue #1188, opened 2024-12-19
Allow setting indentation or tab size


mufeedali:
Currently, SilverBullet seems to default to two spaces for indentation with no option to use my preferred 4 spaces. Is there any way to change this right now? If not, is it possible to implement this?

Thanks :)

zefhemel:
Markdown specifies using 4 spaces to have a different meaning than a tab size of 2, so I don't think we want to have this configurable. When you indent text with 4 spaces, it becomes a code block. 

Observe (in Github):

  Text with two space indent

    Text with four space indent

PR #1212, opened 2025-01-16
Allow setting an indentation using indentMultiplier

mufeedali:
This is kind of a proof of concept for #1188 . Here, the indentation will be set to `2 * indentMultiplier` where `indentMultiplier` is an optional `number` type option the user can set in `SETTINGS`. The paddings in the CSS are also calculated on this basis.

This also doesn't cause any conflicts or other issues with the "four spaces = code block" behavior.


mufeedali:
Any ideas to do this more cleanly would be appreciated. :) (Especially the css and how to set `--editor-indent-multiplier` from the code.)

zefhemel:
While I still don't really see one would want it, I don't have strong objections to merging it and it seems to work as advertised 👍🏻

mufeedali:
> While I still don't really see one would want it

I just like 4 spaces better and have a ton of existing notes using 4 spaces. 😅 Thanks for merging!

mufeedali:
@mjf thanks for reminding me :)

mufeedali:
@zefhemel has this been removed in v2?

eizanprime:
Has it been removed ? Looks like a good feature that would have been useful when using non monospace fonts (I don't like using monospace fonts when not coding)

mufeedali:
@eizanprime Yeah it got removed along with the v2 changes.

I think @zefhemel isn't keen on this feature and I understand why. The current spacing calculations in CSS heavily rely on the assumption of 2 spaces indentation. With nested indentation, spacings aren't even always correct for the default 2 spaces indentation, so something like `indentMultiplier` just makes it worse and more noticeable. And that's setting aside the assumption of a monospace font too.

`indentMultiplier` worked fine for me in v1 (at least as well as 2 spaces) but I haven't been using SilverBullet much since then, part of the reason being that all my older notes (from Obsidian with 4 spaces) look weird now in v2. I think if I was going to do something like this again, I would probably want to actually fix how visual indentation is done here, rather than a workaround like `indentMultiplier`.

PR #1225, opened 2025-01-30
fix: strike through links in tasks

jim-fx:
When the body of a task contains a wiki style link [[]] and is checked, the link is not styled:

image psd

This uses the sort of new css `:has()` selector which is supported by most browsers: https://caniuse.com/css-has


zefhemel:
Awesome, this has been bothering me for quite a while.

Issue #1228, opened 2025-02-02
Template widgets open links in same tab


lutzky:
Expected: External links should open in a new tab, always. (Or in the same tab always, which is worse, but at least consistent)

Actual: External links open in a new tab **except for in a template widget**:

![Image](https://github.com/user-attachments/assets/2274f9bd-9f74-4dd7-8572-0741bb16f770)

Apparently what's going on is that outside of a template widget, [`editor.openUrl`](https://github.com/silverbulletmd/silverbullet/blob/c27c83bc44da8510c49f053936ab7adfafa86b09/web/syscalls/editor.ts#L79) is called when clicking on a link, apparently with the default of `existingWindow = false`, which adds `target = "_blank"`. However, the underlying HTML, generated in `markdown_render.ts`, never has `target = "_blank"`.

lutzky:
That's not due to my change, it was that way before. My change only applies to links in rendered live templates.

PR #1229, opened 2025-02-02
Open external links in a new tab

lutzky:
Fixes https://github.com/silverbulletmd/silverbullet/issues/1228. This matches the generated HTML to the current behavior of editor.openUrl.


zefhemel:
Nice!

Issue #913, opened 2024-07-02
FR: Custom editors


MrMugame:
Custom editors (which would need a good name, maybe `processor` or `manipulator` or smth) are editors provided by a plug ,similar to a `codeWidget`, for specific file types. - Note that the file type is determined by the extension, not by some MIME type. This has the reason that sometimes formats aren’t distinguishable (many text based formats for example). This also makes it easy for the user to know which editor is going to be used with a file. - This would look something like this:
```yaml
name: excalidraw
functions:
  excalidrawEditor:
    path: ./excalidraw.ts:editor
    editor: [".ecalidraw", ".ecldrw"]
```
The called method would then look something like this, again similar to `codeWidget`:
```js
export async function editor(): Promise<{ html: string; script: string }> {
  // return excalidraw editor html/script
}
```
This should just return a barebones editor without a file yet[^1]. The file will be provided to the iframe via messages, which will then trigger a custom event on the iframe `window`. The specific events here will be:

* `file_open`: This event is dispatched when a file is opened, meaning it is called once the editor loads up and can also then be called to reopen a file, this means the editor iframe doesn’t have to be regenerated when switching between different files of the same type.
* `file_change`: This event is dispatched when the current file is changed. This will just pass the new changed file for now, maybe later we could update the files more granularly although live collaboration is not really a goal of silverbullet anyways
* `editor_close`: This event is dispatched when the editor is closed to give it time to save. The iframe internals will when answer back when finished (all event handlers ran) and the iframe will then be deleted. Depending on performance we could probably parallelize closing and opening. 

Saving is handled via a global function either as an `api` call or a custom message.
All this will integrate into the silverbullet experience pretty seamlessly. The attachment files for which an editor (provided by a plug) exists could just be added to the page navigator with some visual mark (or a custom navigator). [^2]
![Dialog](https://github.com/silverbulletmd/silverbullet/assets/40832361/624f546f-6202-4c55-992b-1997b2a528d4)
This would obviously also need the features you already proposed, mainly attachment renaming and attachment deletion. There won’t be attachment creation, this will be handled by the plugs themselves using commands and syscalls (`writeAttachment`) if needed.

Advantages:
* Multiple editors/panels should be easy to implement
* Support for different markup formats (e.g. `Asciidoc`, `Org`)
* Editors for special formats like `stl`, `pdf`, `jpg`.
* Custom markdown editor using xterm and vim or monaco for the crazy will be possible
[...]

PR #1230, opened 2025-02-07
Implementing #913

MrMugame:
So I have some time right now and thought about implementing #913. This is going to be a rather big pull request which is why it's going to be a draft for now.

Added (so far):
- [x] An attachment picker (Currently under `Ctrl-o`) which is another flavor of the page picker and allows the user to (later) open the attachment or if there is no custom editor available it gives the options to rename or delete the attachment.
- [x] Add the new plug type
- [x] Fix lots of TODOs (eg. saving, etc.)

I'd love to get feedback on the decisions as a lot of them are there opinionated


zefhemel:
Looking good so far, no comment yet. A bigger, more philosophical question to think about is whether "attachment" is really the term we should keep using when other file types become more first class citizen. Perhaps we need a different term, open to ideas.

MrMugame:
So after a lot of code mangling, I've got somewhere, but there is still *a lot* to do.

https://github.com/user-attachments/assets/0c1657db-d80e-43e8-ac7e-45b0acdcf0d3

(This is the most minimalistic custom editor possible, but I'm still impressed how fast it is. We'll see later on if caching editors is necessary, it probably is tho)

And to continue further I need some decisions made/advice:
- There are currently some syscalls (mainly the editor syscalls like `editor.setSelection`, `editor.insertAtCursor`), which are depended on the codemirror editor and and are only meaningful to the codemirror editor. And some syscalls also depend on those syscalls. So you kind of get a dependency tree of syscalls and as soon as you switch the editor it all falls apart. As far as I can tell - you know Silverbullet a lot better so maybe you can think of a case - the top of the this dependency tree is only made up of *commands*. Meaning the if the user can't run commands which depend on a specific editor the problem is "gone". Clearly the syscalls will still exists, but they *shouldn't* be called when a different editor is open.  This is why I would propose something along the lines of:
```yaml
name: jpgeditor
functions:
  JPGEditor:
    name: "jpg-editor"
    path: ./editor.ts:editor
    editor: ["jpg", "jpeg"]
  invertColor:
    path: "./editor.ts:invertColor"
    command:
      name: "Image: Invert Color"
    dependsOnEditor: "jpg-editor"
```
Unsure about the concrete syntax here? Can we use the yaml name of the function (so `JPGEditor` in this case)? Does that make sense?
(Just noticed there are also click events like `page:click`, which trigger further syscalls and panelWidgets also triggers further syscalls. But panelWidgets don't exist for other editors, so no problem. And a **page**:click event also shouldn't exist anymore. At least for me a page is a markdown page)

- Shortcuts are currently handled through codemirror. If codemirror is gone so are shortcuts. There are a couple solutions to this problem.
    - Keep a "ghost" codemirror instance to which we send all key strokes
    - Create a custom keybinding handler
    - Use a third party keybinding library

Unsure whats best, but I'm not a big fan of keeping a ghost codemirror instance, because if we ever want to move all the default editor stuff into a plug we would still need to depend on codemirror for some weird background key handler.
 - Naming is still a big problem, like you said. Currently attachments are still called attachments and "custom editors" are called *dedicated editors*, which has the sole reason that I felt like that's an easy to search term when I replace it later. So I don't really have good ideas here. But that should be the least of our problems right now.

MrMugame:
So, next problem... and I don't have any ideas on possible solutions for this one.
How would one send the user to the editor for a specific attachment (e.g. jpg) when he goes there using the link directly. If I just send all file requests to the frontend, stuff like favicons and _plugs stop working (and probably other stuff), because the raw file isn't sent anymore. But I can't really filter, because I don't know which file types have custom editors. Just using a whitelist for favicon/plugs seems like an awful approach

zefhemel:
To decide which commands are available in the command palette, you can do something like is already implemented for `requireMode` (for commands), right now the options there are `rw` and `ro`, but perhaps you make it so that an editor name can be specified there (`requireMode: jpeg_editor` or whatever).

I'd assume you keep the currently active editor mode in the AppViewState somewhere (web/type.ts). You would then be able to switch between those from a plug with some new syscall, which invokes a reducer. Have a look how this works to e.g. toggle vim mode (AppViewState has uiOptions.vimMode). If you have that, you probably don't need additional magic manifest fields.

As to how to decide which editor to enable, here I would indeed implement a new Hook, probably, which perhaps specifies a new `pathPattern` (or similar) regexp field, which would determine what editor to open given the path (so, e.g. `\.(jpg|jpeg|png|gif)$` for an image viewer).

zefhemel:
> How would one send the user to the editor for a specific attachment (e.g. jpg) when he goes there using the link directly. If I just send all file requests to the frontend, stuff like favicons and _plugs stop working (and probably other stuff), because the raw file isn't sent anymore. But I can't really filter, because I don't know which file types have custom editors. Just using a whitelist for favicon/plugs seems like an awful approach

Yes this is a big one, and the original source of all the head aches of "why can I not put file extensions in my page name?" The reason for this is indeed so that SB can distinguish between loading with SB UI _loading_ a particular page, vs the content of that actual page. `/hello` opens the UI which in turn loads /hello.md. However, to support various auth proxies (very long story) what currently happens is that if you load /hello.md from the browser directly, it actually redirects you to `/hello`. It does so if no `X-Sync-Mode` HTTP header is set (fetching this file from the SB client _always_ sets this header). This is not the case for non `.md` files however, because that would break e.g. loading attachments (e.g. when loading an image from an  tag). Which puts us in a bit of a pickle, that honestly I also don't really know how to solve...

MrMugame:
So yes, I was anticipating the problem with syscalls as I didn't know if there were any that ran without input. So this is good ✅. My main "issue" was that some syscalls would look something like this then:
```ts
"editor.getText": () => {
  if (client.isDedicatedEditor()) return ""; // or undefined

  return client.editorView.state.sliceDoc();
},
```
We could just leave the implementation like it was but then the behaviour would be undefined if a custom editor is open, which shouldn't be a problem but is still not super ideal especially for plug development.

---

Putting it under `requiredMode` is probably a good idea. ✅

---

Not really sure what you mean here
> I'd assume you keep the currently active editor mode in the AppViewState somewhere (web/type.ts). You would then be able to switch between those from a plug with some new syscall, which invokes a reducer. Have a look how this works to e.g. toggle vim mode (AppViewState has uiOptions.vimMode). If you have that, you probably don't need additional magic manifest fields.

---

We could do regex or something similar. I did it like this. Wouldn't be *to hard* to change later.
```yaml
name: jpgeditor
functions:
  JPGEditor:
    path: ./editor.ts:editor
    editor: ["jpg", "jpeg"]
```

MrMugame:
Could we take an educated guess using the accept header? I think this could work most of the time

MrMugame:
So while implementing I noticed, that putting the required Editor under the `requiredMode` field is probably a bad idea. So (I don't know how you imagined it, but this is how I would have done it). commands, which specify `"rw"` and `"ro"` would just be for pages and e.g. `"jpg-editor"` would be for custom editors. This however has the problem that some jpg-editor commands don't work in readOnly mode so they should also be able to specify this. Same goes for commands like `Upload File`, which work in every mode but needs write access. So these are actually two completely separate problems. Now we could create weird syntax (e.g. `jpg-editor+rw`),  but this would probably break legacy support in some way if we don't have edge cases etc. etc. I would just opt for a new field.

MrMugame:
So coming back to the http server problem: This actually seems to work pretty well
```ts
this.app.get(filePathRegex, async (c, next) => {
  //....
  if (req.header("Accept")?.includes("text/html")) {
    return next();
  }
  //...
}
```
I haven't really tested everything yet (tbh I don't even know what even needs testing) and I don't know in what way this will clash with other plugs/tools/scripts/forks/etc. people have made. So I don't know if this is a solution we can roll with but it seems to work. Only problem I see right now is if someone ever wanted to keep raw html files in his space and maybe have a custom viewer for html files.

MrMugame:
Short question:
*page_ref.test.ts*
```ts
  assertEquals(parsePageRef("user@domain.com"), {
    kind: "page",
    page: "user@domain.com",
  });
```
What is this testcase for?

MrMugame:
So I'm pretty far now, @zefhemel. Most of the basic features are implemented. You can test it using this super basic [plug](https://github.com/MrMugame/silverbullet-txt) for txt files if you want. I haven't done a whole lot of testing so far, so there are probably still some bugs.
I've collected the exposed changes so far.
- `Ctrl-o` Opens attachment navigator
- `requireEditor` manifest property for attachments, possible values are (“page” | “any” | “notpage” | some editor name)
- The `PageRef` object passed to `editor.navigate` is now of type `LocationRef` and *should* (doesn’t have to) contain a `kind` property of type `"attachment" | "page"`
- The behaviour of (most) editor syscalls will be undefined, while a dedicated editor is open
- Added two new syscalls `editor.getCurrentPath` `editor.getCurrentEditor`
- New command `Attachment: Rename Attachment Page`
- Added events `editor:attachmentSaving`, `editor:attachmentLoaded` and `editor:attachmentReloaded`
- If request for a file like path comes in and it has an accept header it will be sent the UI
- New parameter for attachment path: `?raw=true` returns the raw file regardless of the accept header

So going forward there are still some TODOs/Questions:
- [ ] Write documentation
- [ ] Decide on decent naming. I would just opt for *page*, which would be a markdown file and *document* which is everything else i.e. replace *attachment*/*dedicated editor* with *document*. This is the clearest ~~Chat-GPT~~ I could come up with
- [ ] Fix a few minor TODOs
- [ ] Should we check for two plugs trying to create an editor for the same extension? I don't think slash commands check for this for example. Just take the one of the plug, which is iterated last?
- [ ] Should we move ~~attachments~~documents along with the page? I don't think this makes sense anymore

Yeah so maybe if you have time you can take a look, because this PR is getting close to being feature complete

zefhemel:
I have absolutely no idea where it comes from...

MrMugame:
> I have absolutely no idea where it comes from...

If you are referencing the test case, then I'm just going to remove it because it kind of interferes with the new logic because it also has an "extension".

And I would love to get this into a finished state. If you are ok with calling attachments documents from now on, I would start renaming everything: Only problems is that a lot of syscalls and events have "attachment" in their name. Should we do a breaking change or have legacy syscalls or something else?

zefhemel:
I'm thinking if there's likely to be any external plugs that is using any syscalls related to attachments, so I think you can safely rename it. I think "documents" makes sense, so let's go with that.

MrMugame:
Super simple [PDF Viewer](https://github.com/MrMugame/silverbullet-pdf)

zefhemel:
Just had a skim of the Changes and it looks very promising (and very big). What happened with the "F Deno" thing in your pdf repo? 😆

zefhemel:
Yep: https://github.com/silverbulletmd/silverbullet/commit/280fa2cd3c0dbe43513a5102bf8ce6423ef6a673

zefhemel:
About this `requireEditor: page` annotation. Did you specifically add these to commands that break without it, or more generally? There's a bunch of commands that I would assume should just work with no editor active (e.g. Library: Update) but perhaps they don't (not a big deal). Since there's so many page-editor only commands, would it be worth considering making this the default, and only setting requireEditor: something else if you need it?

website/API/space.md
```diff
@@ -1,4 +1,4 @@
-The Space API provides functions for interacting with pages, attachments, and files in the space.
+The Space API provides functions for interacting with pages, documents, and files in the space.
```
zefhemel, line 1: I understand you did a find & replace and that this was redundant before, but I think you can just leave out the "and files" here.

common/space.ts
```diff
@@ -11,7 +11,7 @@ const pageWatchInterval = 5000;
 export class Space {
   // We do watch files in the background to detect changes
   // This set of pages should only ever contain 1 page
-  watchedPages = new Set<string>();
+  watchedFiles = new Set<string>();
```
zefhemel, line 14: I figure you want to use this for documents too now? So to auto reload when changes are detected? If so...

common/space.ts
```diff
@@ -22,9 +22,11 @@ export class Space {
     eventHook: EventHook,
   ) {
     eventHook.addLocalListener("page:deleted", (pageName: string) => {
```
zefhemel, line 24: Maybe this has to change to become a `file:changed` event?

zefhemel (commented):
Not done going through this, but some intermediate comments.

website/API/space.md
MrMugame: But there is a separate api for files, so I think this sentence isn't really wrong, we could change it to something like this
> The Space API provides functions for interacting with pages, documents, or files in general inside the space

common/space.ts
MrMugame: I would think `file:deleted`, if a file just changed we'd still wanna watch it?

common/spaces/evented_space_primitives.ts
MrMugame, line 1: Would we also want to emit `document` events here? I wouldn't really know of a use case right now?

MrMugame:
> About this `requireEditor: page` annotation. Did you specifically add these to commands that break without it, or more generally? There's a bunch of commands that I would assume should just work with no editor active (e.g. Library: Update) but perhaps they don't (not a big deal). Since there's so many page-editor only commands, would it be worth considering making this the default, and only setting requireEditor: something else if you need it?

I started without adding this extra property and then back paddled, which made the transition a little messy. I think I forgot to properly go over everything one more time in the end. It should be better now.

I thought about making `page` and not `any` the default, but I decided against it, because
1. I think semantically it makes more sense
2. It definitely doesn't break backwards compatibility for plugs (You could also argue here that most plugs implement stuff which need the page editor anyways)

I can see both arguments here

common/space.ts
zefhemel: Yes. That's what I meant.

zefhemel:
What's the state of the PR (beside the merge conflict). Would it be ready to go from your end?

MrMugame:
It would be ready to go from my end.

I've been using it the past couple days a little bit and I haven't noticed any major bugs, if bugs come up I'll be in here.

zefhemel:
Ok, I looked at the code. It's a lot so couldn't really ingest all the details but overall it makes sense to me and there's plenty of good stuff in there. I also ran it for a bit locally and tried your txt and PDF examples. I really like that the txt editor even saves content. Nice work.

What I'll do is just merge this, this will allow edge users (including me) time to test this and report issues.

What I'd be interested in is to have at least 1 document viewer baked in. I'm thinking perhaps one for images? Doesn't need to be fancy, but just to have some code that actually uses this without having to resort to install plugs. Thoughts?

Very nice work here @MrMugame thanks for putting so much effort into this.

PR #1232, opened 2025-02-09
Feature/move lines up down

scambier:
I'd like to be able to move _all_ lines up/down with the same shortcut (alt+up/down) as "Outline: move up/down".

The problem is that Silverbullet doesn't support having 2 commands bound to the same shortcut (is that correct?), so this PR alters the `outline.moveItemUp()` to fallback on the simpler "move line up/down" commands when possible.

Even though it works, I'm not super satisfied with that solution, so advice is welcome :)

[silverlineup.webm](https://github.com/user-attachments/assets/e54b1a2c-b4b7-4f34-842d-8077f6bc0c29)

### DIFF
--- a/plugs/editor/editor.plug.yaml
+++ b/plugs/editor/editor.plug.yaml
@@ -351,14 +351,14 @@ functions:
   outlineMoveUp:
     path: ./outline.ts:moveItemUp
     command:
-      name: "Outline: Move Up"
+      name: "Move Outline Item or Line Up"
       key: "Alt-ArrowUp"
       requireMode: rw
 
   outlineMoveDown:
     path: ./outline.ts:moveItemDown
     command:
-      name: "Outline: Move Down"
+      name: "Move Outline Item or Line Down"
       key: "Alt-ArrowDown"
       requireMode: rw
 
--- a/plugs/editor/editor.plug.yaml
+++ b/plugs/editor/editor.plug.yaml
@@ -351,14 +351,14 @@ functions:
   outlineMoveUp:
     path: ./outline.ts:moveItemUp
     command:
-      name: "Move Outline Item or Line Up"
+      name: "Outline: Move Up"
       key: "Alt-ArrowUp"
       requireMode: rw
 
   outlineMoveDown:
     path: ./outline.ts:moveItemDown
     command:
-      name: "Move Outline Item or Line Down"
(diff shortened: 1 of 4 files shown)


plugs/editor/editor.plug.yaml
```diff
@@ -351,14 +351,14 @@ functions:
   outlineMoveUp:
     path: ./outline.ts:moveItemUp
     command:
-      name: "Outline: Move Up"
+      name: "Move Outline Item or Line Up"
```
zefhemel, line 354: Honestly even though not technically correct, I'd just keep the old command name here.

plugs/editor/editor.plug.yaml
```diff
@@ -351,14 +351,14 @@ functions:
...
       key: "Alt-ArrowUp"
       requireMode: rw
 
   outlineMoveDown:
     path: ./outline.ts:moveItemDown
     command:
-      name: "Outline: Move Down"
+      name: "Move Outline Item or Line Down"
```
zefhemel, line 361: And here

plugs/editor/outline.ts
```diff
@@ -1,11 +1,18 @@
 import { editor } from "@silverbulletmd/silverbullet/syscalls";
+import { moveLineDown, moveLineUp } from "../../plug-api/syscalls/editor.ts";
```
zefhemel, line 2: Since you're already importing `editor` the line above. You can just call `editor.moveLineDown` etc without importing it again.

plugs/editor/outline.ts
```diff
@@ -69,7 +76,13 @@ export async function moveItemDown() {
   const text = await editor.getText();
 
   try {
-    const currentItemBounds = determineItemBounds(text, cursorPos);
+    let currentItemBounds: ReturnType<typeof determineItemBounds> | undefined;
+    try {
+      currentItemBounds = determineItemBounds(text, cursorPos);
+    } catch {
```
zefhemel, line 82: This is _slightly_ ugly, but perhaps ok as it is. I would add a comment explaining in what situations we would expect this to fail (specifically when not operating on bulleted lists etc.)

zefhemel (changes requested):

zefhemel:
Yep, looks good. thanks!

PR #1240, opened 2025-02-14
Vim Mode and Read Only Fixes

allie-wake-up:
Currently in vim mode if the editor isn't selected and I hit `j` it will cause the editor to gain focus and insert the j character before entering vim normal mode. Read only mode is broken when toggling vim mode.  This PR attempts to fix those issues.  I don't know if there's a better way to fix the read only mode than calling reloadPage but it seems to work

- remove client.focus call from global keydown handler. This call is what breaks vim mode

- add client.reloadPage call to setUiOption syscall. This appears to fix read only mode that is completely broken without the above focus call. It also seems to fix the issue when toggling vim mode

- add client.focus call to CommandPalette onTrigger if no cmd is selected. Without the client.focus call above, the editor is no longer selected when exiting the command palette with escape - this call fixes that

- add watch-all task to deno.json to run all 3 watches in 1 terminal window

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -48,6 +48,7 @@
     "watch-web": "deno run -A --check build_web.ts --watch",
     "watch-server": "deno run -A --unstable-kv  --unstable-worker-options --check --watch silverbullet.ts",
     "watch-plugs": "deno run -A --check build_plugs.ts -w",
+    "watch-all": "deno task watch-web & deno task watch-server & deno task watch-plugs",
 
     "bundle": "deno run -A build_bundle.ts",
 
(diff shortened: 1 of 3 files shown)


deno.json
```diff
@@ -47,6 +47,7 @@
     "watch-web": "deno run -A --check build_web.ts --watch",
     "watch-server": "deno run -A --unstable-kv  --unstable-worker-options --check --watch silverbullet.ts",
     "watch-plugs": "deno run -A --check build_plugs.ts -w",
+    "watch-all": "deno task watch-web & deno task watch-server & deno task watch-plugs",
```
zefhemel, line 50: I'm ok with adding this, but then let's add deno task watch-server at the end so you can pass it arguments (like a folder), for instance `deno task watch-all ~/tmp`

zefhemel (changes requested):
I have to look at this more closely and test a lot of scenarios. It was kind of a pain to get this to work again last time, and moving the client.focus call may break yet other cases. Some patience :D

zefhemel:
Ok, this seems to work as advertised. Thanks!

PR #1243, opened 2025-02-15
Add option to personalize toc header

KitAmbraid:
Hellow,

I'm setting up silverbullet to see if it can replace Obsidian in my life, and I needed to change what heading level the table of content header was. So, here's a PR that **is** backward compatible (since we could remove `header` and instruct to set `headerText` to `""` to achieve the same effect)

I'm not happy with the wording I came up with for the documentation, suggestions very much welcome.


zefhemel:
Looks good

PR #1244, opened 2025-02-16
Fix spell checking for code blocks and inline code

MrMugame:
This is related to this [post](https://community.silverbullet.md/t/disable-spell-checking-in-code-blocks/1688/2). I was kinda right. It has to be set for `InlineCode` and `CodeText`, but inline code doesn't have to be less than a line and we don't want to disable spellchecking for the whole line so this has to be moved out of the the linewrapper codemirror plugin.


zefhemel:
This does seem to add an additional span around these elements in the DOM, but hopefully it doesn't break anything. Let's see, thanks!

PR #1246, opened 2025-02-16
Fix href check for action buttons

fitzgeraldkd:
## Context

Custom action buttons are currently being wrapped in an anchor tag, even though there is no `href` value. The anchor element (32x39.2) is larger than the button element (28x28).

The anchor element:
image

The button element:
image

## Issue

If you click the anchor element without clicking the button it refreshes the entire page instead of performing the intended action.

Tested on Windows in both Firefox and Chrome.

## Fix

Instead of strictly checking if `actionButton.href !== undefined`, we can just check the href's truthiness.

With this change, the buttons are not wrapped in the extra anchor tag:
image

### Alternate Fix

The href for custom buttons seems to be always set to an empty string per in `editor_ui.tsx` (see snippet below), so another fix for this issue would be to set it to undefined instead of an empty string. Happy to update the PR if this would be preferred over what I currently have committed!
https://github.com/silverbulletmd/silverbullet/blob/5111b5a3803262e163658b74d662da1365905b44/web/editor_ui.tsx#L304-L314

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Setup Deno
         uses: denoland/setup-deno@v1
         with:
-          deno-version: v2.1.9
+          deno-version: v2.2.0
 
       - name: Run bundle build
         run: |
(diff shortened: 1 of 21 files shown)


zefhemel:
Thanks! Backported a change from the main branch to resolve the conflict.

PR #1250, opened 2025-02-18
Space lua: os.date add support for %U, %V and %W

KitAmbraid:
Add support to format dates using the week number format specifiers:

- `%U`: US format - week 01 of the year starts on the first Sunday of the year. Days in the year prior to that week are in week 00. (00 to 53)
- `%W`: week 01 of the year starts on the first Monday of the year. Days in the year prior to that week are in week 00. (00 to 53)
- `%V`: EU / ISO 8601 format - week 01 of the year is the week containing January 4th. Days in the year prior to that week are in the last week of the year prior. (01 to 53)


zefhemel:
Nice!

PR #1254, opened 2025-02-21
add rebuildEditorState function to the syscalls

numanhg:
This commit exposes `client.rebuildEditorState()` function to the syscalls.


zefhemel:
Thanks!

PR #1263, opened 2025-02-25
Remove `editor.setText` syscall from client

### LINKED ISSUE #1262: Browser Reload and page breaks - TypeError: globalThis.syscall is not a function
I noticed that when accessing certain pages via URL directly they break and don't show. 
Page exists with original content as md file, but is not loaded. If I start typing something onto page, original contents get lost and rewritten.

Accessing the same page via "Open Page" button (Command+K) loads the page with no problem. 
However, clicking the reload button and the page does not load again.

This only happens for some of the pages, not all of them. 
Looking at the browser console, when the page does not load the following error is thrown:
async.ts:99 TypeError: globalThis.syscall is not a function
    at ft (syscall.ts:16:30)
    at Object.xE (editor.ts:35:10)
    at Lp.loadPage (client.ts:1147:20)
    at async client.ts:479:7
    at async navigator.ts:125:1

I'm deploying silverbullet via docker. Notes (md files) were there before, not necessarily created via SB GUI.
I tried "System: Reload" and "Space: Reindex", but the issue persists.

MrMugame:
Fixes #1262

### DIFF
--- a/web/client.ts
+++ b/web/client.ts
@@ -5,6 +5,7 @@ import type {
 import type { Compartment, EditorState } from "@codemirror/state";
 import { EditorView } from "@codemirror/view";
 import { indentUnit, syntaxTree } from "@codemirror/language";
+import { isolateHistory } from "@codemirror/commands";
 import { compile as gitIgnoreCompiler } from "gitignore-parser";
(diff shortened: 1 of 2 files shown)


zefhemel:
Yes, thank you :)

PR #1287, opened 2025-03-11
Use writer to ensure the entire file is written

lutzky:
Using `file.write` directly once doesn't guarantee the entire file is written; see https://docs.deno.com/examples/writing_files/

Noticed this while trawling through the code working trying to reason about sync issues


zefhemel:
Yikes that's scary. Thanks!

PR #1288, opened 2025-03-12
fix(transclusions): don't slice if it is the last heading on the page

### LINKED ISSUE #1086: h2 transclusion does not render if it's the last section of the page
section transclusion (e.g. `![[page name#title]]`) does not render in case it's the last section on the page and it's a h2 or higher (h3, h4, ...) heading.

janssen-io:
The check if a match was found by the heading regex was incorrect. `string.search(regex)` returns `-1` which is "truethy". 
Instead use a check if it's greater or equal than 0 (technically correct, but in practice it will even be bigger, because a next heading is at least a newline away).

Fixes #1086

### DIFF
--- a/common/query_functions.ts
+++ b/common/query_functions.ts
@@ -135,10 +135,9 @@ export function buildQueryFunctions(
               `[^#]#{1,${headingLevel}} [^\n]*\n`,
               "g",
             );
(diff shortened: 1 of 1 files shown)


zefhemel:
Nice catch!

PR #1289, opened 2025-03-13
Add image viewer

MrMugame:
This is going to be the last feature pull request related to document editors.

- I've added a simple image viewer for all file types supported by the `` tag. It uses the [panzoom library](https://github.com/timmywil/panzoom) for well panning and zooming as implementing this myself would probably drive me crazy. I've just inlined it to avoid any build steps.
- Fixed some minor bugs & reworked a bit of code
- Add a new syscall `editor.sendMessage`, which allows the user to send a message from e.g. a command to the document editor. This pretty much allows any type of communication between the too, so this is ~~probably~~ going to be the only syscall added for document editors. I'm really open to new naming suggestions here tho. (I'm using this for a new `Image Viewer: Show Info` command here)


zefhemel:
Nice, I'll merge this as is. 

Question though: the image flip buttons, they flip the picture but I'd also expect those flips to somehow persist and they don't. If this is the case, is it really useful to even show them?

MrMugame:
What I think would be more interesting is adding indexing to images. So you could query images taken in a specific time slot or at a specific location, but I don't directly have a use for this.

zefhemel:
I don't think it needs editing capabilities, but perhaps then don't show the controls to flip the image?

As to indexing: muah, don't have a use case either.

PR #1292, opened 2025-03-14
fix(tables): only escape pipes outside SB context

### LINKED ISSUE #943: Links in query results break when adding an alias
I'm querying some random data and want to add links to the resulting table, which works fine. However, when I add an alias to that link, it breaks because a slash is added at the end of the link.

image

image

### LINKED ISSUE #1016: Commands in tables with custom labels don't work
Example:

```
| One | Two |
| --- | --- |
| {[Flash: Custom Message\|Hello]("Hello")} | bug | 
```

With escape on | in {[ ]}:
image
image

Without escape:
image

janssen-io:
- Fixes #943
- Fixes #1016 

The transformer used to escape any pipe (`|`), which lead to incorrect links (#943) and broken command buttons.
To fix the command buttons (#1016), the check to see if we are inside the command-syntax was added in a similar way as the check of the wiki-links.

### DIFF
--- a/common/markdown_parser/parser.test.ts
+++ b/common/markdown_parser/parser.test.ts
@@ -240,3 +240,43 @@ Deno.test("Test NakedURL parser", () => {
     "http://abc.com?e=2.71",
   ]);
 });
+
+const tableSample = `
+| Header A | Header B |
+|----------|----------|
+| [[Wiki|Alias]] | 1B |
+| 2A             | 2B |
+| 3A | {[My: Command|Alias]("args", 1)} |
+`;
+
+Deno.test("Test table parser", () => {
+  const tree = parseMarkdown(tableSample);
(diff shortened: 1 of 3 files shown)


janssen-io:
I think the fix for 1016 should be small as well, so I will add it in this PR.

janssen-io:
The fixes don't feel quite elegant, but they do work. Any suggestions are welcome!

zefhemel:
In time, I think command links will go away in favor of using Lua widgets (creating a lua widget implementing a command link would be trivial). However, until then I'm ok with this solution. What would make me a bit more confident of this could would be some test cases. Could you look if you can add those e.g. somewhere here? https://github.com/silverbulletmd/silverbullet/blob/main/common/markdown_parser/parser.test.ts

zefhemel:
Excellent, thank you!

Issue #847, opened 2024-04-13
Table alignments are not working


katiuskt:
According https://silverbullet.md/Markdown/Extensions I would expect tables [alignment ](https://www.markdownguide.org/extended-syntax/#alignment) to be supported by SilverBullet but it looks like they don't (is always left aligned regardless your MD text):

```
| Not specified alignment| Centered aligned | Aligned to the right |
| --- | :---: | ---: |
| Test1 | Test1 | Test1 |
| Test2 | Test2 | Test2 |
```

SilverBullet (docker installation, latest release available 0.7.6)
![imagen](https://github.com/silverbulletmd/silverbullet/assets/75622100/9dd2f093-09ca-4d44-ac47-dbe6ffe8609d)

Same code, another MD editor (snap from obsidian):
![imagen](https://github.com/silverbulletmd/silverbullet/assets/75622100/b3fe0299-5ce0-490e-9aec-d17d4796fcc2)

zefhemel:
Indeed, this has not been implemented yet.

djm2k:
I'm looking at this one, I found another issue with [this example](https://github.github.com/gfm/#example-199) where missing `|` at the start/end of a line should still be parsed as a table.
Will open a PR to address the following:

- [ ] parseRow does not handle colons `:` to indicate `align` HTML properties for table columns
- [ ] parseRow does not support a mismatched number of pipes in data rows

PR #1293, opened 2025-03-16
Justifying tables

probablySophie:
This isn't a 100% serious pull request, because the code I wrote is hot garbage, but I've kept it in its own file so it can be pulled apart easier.

This follows the [markdownguide specifications](https://www.markdownguide.org/extended-syntax/#alignment) aligning row cells and header cells.

This is *not* the best way of doing this (inline style code 😞), but with an iteration or two probably could be decent?

There's a relevant issue (#847) and I'll keep working on this when I have time


zefhemel:
@mschmidtkorth took this over the finish line I think, thanks!

Issue #481, opened 2023-07-26
Switch to dark/light mode based on OS setting


zefhemel:
But still allow overrides

pihentagy:
To query the dark mode once:

```js
if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
    // dark mode is enabled
    document.documentElement.classList.add('dark-mode');
} else {
    // dark mode is disabled
    document.documentElement.classList.remove('dark-mode');
}
```

And to watch it reactively:

```js
window.matchMedia('(prefers-color-scheme: dark)').addListener((e) => {
   if (e.matches) {
       // Enable dark mode
      document.documentElement.classList.add('dark-mode');
   } else {
       // Disable dark mode
       document.documentElement.classList.remove('dark-mode');
   }
});
```

zefhemel:
Implemented here #1294

Issue #1290, opened 2025-03-13
UI with light styles still flashes momentarily when dark theme is selected


isti115:
With dark mode enabled, the default styles still get applied for a few frames while the page is loading, which results in a quite jarring flash. With the devtools open I can see that the `<html>` tag initially gets added to the DOM with no `data-theme` property, then gets set to `data-theme="light"`, and only after does it receive the actual `data-theme="dark"` value.

There are three approaches that I can think about right now:
- Loading the saved theme value early enough for it to be available by the time the first frame is rendered
- If the value is not yet available, respect the `prefers-color-shceme` value provided by the browser (implementing #481 would probably solve this as well)
- Render a gray background until the required information becomes available (this seems like an unnecessary compromise, but would still be better than the current behavior in case you are against implementing either of the two suggestions from above)

Personally I would probably go for the second option, if you'd also be happy with that, I'll send a PR.

---

P.S. Is it just me, or are `code blocks` barely visible in the current dark theme? Since everything is already monospace, the font doesn't help in distinguishing them, and their background is so subtle, that I can only see them if I know exactly where to look.

Issue #657, opened 2024-01-26
Dark mode goes away when you switch to sync mode


devzero:
I'm not sure if this is intended or not, but is surprising.  On a fresh install, switch to dark mode with the command, then click the sync mode button to go offline, once it syncs, it is no longer in dark mode (interestingly it switches back to dark mode when you switch back to online mode).  Does it pivot all local editor configs?  I noticed that the dark mode settings isn't in settings by default but your state for it sticks with you.  Could that be an issue going forward?

zefhemel:
Hmm odd, would have to check where I track that.

RLB2310:
Just wanted to add, It's not just going to Sync mode. Generally I've noticed if I ever clear the cache in my browser it removes the theme until I turn it on again in both modes. If i don't touch the host IP on the server or remove the cache the theme is consistent.

zefhemel:
The reason is that it's stored per client by design so that you can have different clients with different settings. If you wipe the client the setting is gone. 

The sync mode issue I haven't looked into still.

lilotter45:
I experienced this issue a while ago, so I'll add my experience if it is helpful:

I was able to produce the same behavior on iOS using Silverbullet as a PWA by doing the following:

1. Change mobile client to dark mode
2. Turn on sync mode
3. Client reverts to light mode

I didn't file a bug because I found that if I reloaded the client, either by fully exiting and reopening or running `System: Reload`, it would reapply dark mode.

PR #1294, opened 2025-03-16
[#481] If user hasn't explicitly set a dark mode pref in Silverbullet, fall back to browser pref

niels:
This addresses #481.

To differentiate between a preference having been set explicitly (by toggling dark mode on or off _within_ Silverbullet) or not, the `darkMode` property across the various components can now either be a boolean, or `undefined`.

The initial view state sets it to `undefined`.

When loading the editor UI, the client store is inspected to find an explicitly set preference. If one is found, that preference gets applied. Otherwise, the browser setting is inspected (via a media query) and the UI theme gets set accordingly.

If/once an explicit preference has been set, subsequent changes to the browser setting no longer affect Silverbullet.

One caveat: If the user hasn't set a preference within Silverbullet and then changes their browser setting, the change only gets applied upon page reload. (Toggling the preference within Silverbullet itself already triggers a page reload, thus applying the changed setting automatically.)

As this is my first time interacting with this codebase, my approach may be suboptimal. I would be happy to iterate as feedback is received. For now, with my limited understanding, this seemed like the most straightforward path to take.


janssen-io:
Can we add an explicit 'system' option next to 'light' or 'dark'. Just so that once a user has set something, it's easy to go back to system?

niels:
I agree that a "theme" setting with string values would, ultimately, be the best approach. We would initially have system, dark and light; and possibly others in the future.

What I struggle with is the current command UI. You can only "Toggle Dark Mode". For that UI, I don't think a "system" setting makes sense. The app loads in either light or dark mode, and toggling should then simply switch to the other. A simple toggle has no inherent concept of "system" vs. "user-set". By it's very nature, once I perform the toggle, I have set a user preference.

So I think we would have to add the new preference to the SETTINGS page instead. When doing so, would we want to retain the toggle? Would the toggle change what's set in SETTINGS? Changing back to the system theme would then be an "advanced feature" requiring the user to modify their SETTINGS manually.

There are a number of fairly closely related issues:

- https://github.com/silverbulletmd/silverbullet/issues/10
- https://github.com/silverbulletmd/silverbullet/issues/657
- https://github.com/silverbulletmd/silverbullet/issues/1290

It seems to me that it would make sense to discuss them together to come up with a design that addresses them all.

This PR here would then simply be an iterative improvement of the current functionality, rather than a more holistic approach to theming—with the latter shelved for now.

Would that be acceptable?

janssen-io:
That's an excellent point. If it would be up to me then it still would be great to add an extra command to the core library that clears the setting (instead of having to dig into the client storage myself).

But other than that, it would be awesome to have this as soon as possible.

zefhemel:
Sorry for the delay on this, adding it to my todo list. Will need to do some rebasing based on recent changes.

zefhemel:
Manually merged this in here, thanks a lot! https://github.com/silverbulletmd/silverbullet/commit/e0fff9d07c0846aec4b71196299429b2190681bb

PR #1298, opened 2025-03-17
feat(menu): Make menu responsive

### LINKED ISSUE #1033: When viewing Silverbullet on mobile and there are more than a few buttons on the header, the page name is very truncated
Maybe it would be ideal if the buttons were converted to a hamburger menu when the screen size is small enough using CSS. Then when clicking the menu icon, the buttons drop down and can be selected. This way more of the page name would be visible on small screens.

image

janssen-io:
Fixes #1033

Adapted the [implementation](https://community.silverbullet.md/t/collapsible-vertical-action-menu-for-mobilescreens-but-not-only/1552/7) by [Mr.Red](https://community.silverbullet.md/u/Mr.Red) and [Varenia](https://community.silverbullet.md/u/s-varenia) from the SB Community.

1. The mobile menu button is added directly in the Main UI ViewComponent's actionButtons property.
2. The styling for `#sb-top` is moved to its own file, so that the media query can stay close to it. Hopefully this makes it less likely that the mobile menu is forgotten when updating the top bar styling.
3. Use CSS nesting (original implementation did not use nesting)

### DIFF
--- a/plugs/editor/editor.plug.yaml
+++ b/plugs/editor/editor.plug.yaml
@@ -37,7 +37,7 @@ config:
     hideSyncButton:
       type: boolean
       nullable: true
-    menuClass:
+    mobileMenuStyle:
       type: string
       nullable: true
     actionButtons:
--- a/plugs/editor/editor.plug.yaml
+++ b/plugs/editor/editor.plug.yaml
@@ -37,6 +37,9 @@ config:
     hideSyncButton:
       type: boolean
       nullable: true
+    menuClass:
+      type: string
+      nullable: true
     actionButtons:
       type: array
       items:
--- a/type/config.ts
+++ b/type/config.ts
@@ -74,7 +74,7 @@ export const defaultConfig: Config = {
   actionButtons: [], // Actually defaults to defaultActionButtons
   autoCloseBrackets: "([{`",
   indentMultiplier: 1,
-  mobileMenuStyle: "mobile-vertical",
+  mobileMenuStyle: "hamburger",
 
   schema: {
     config: {
--- a/type/config.ts
+++ b/type/config.ts
@@ -50,7 +50,7 @@ export type Config = {
   autoCloseBrackets: string;
   smartQuotes?: SmartQuotesConfig;
   indentMultiplier?: number;
-  menuClass: string;
+  mobileMenuStyle: string;
 
   schema: SchemaConfig;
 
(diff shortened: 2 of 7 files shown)


janssen-io:
You make a great point, mjf. SilverBullet is all about extensibility, so your approach would fit that vision better.

Initially, I thought this would be a great improvement, because the current solution is just meh and extensibility often comes with the cost of complexity as well. 

I'll have a look at how we can maybe still incorporate this solution and make it opt-out for the sake of extensibility (by changing the class).

janssen-io:
Added a bottom bar (when `menuClass` is set to `mobile-bar`) option, which looks like this:

image

- The black bar on the bottom is from the browser itself.
- When the virtual keyboard pops up, the bar stays behind it. But the keyboard can easily be closed again (on Android and iOS).

zefhemel:
My experience with implementing bottom bars is that they... don't work — like they are hidden behind the keyboard when they unfold etc. But perhaps this was recently fixed in more recent Safari versions, let me test a bit.

zefhemel:
I just tested this and honestly I quite like it. I'd request just one change: to change the name of the setting to `menuStyle` to focus less on that this may use CSS classes under the hood (but still hint that it might 😉 )

The bottom bar indeed hides under the keyboard, which isn't great, but may be a reasonable trade off for people. Personally I wouldn't use it, but power to the people.

zefhemel (changes requested):
I just tested this and honestly I quite like it. I'd request just one change: to change the name of the setting to `mobileMenuStyle` to focus less on that this may use CSS classes under the hood (but still hint that it might 😉 ) and emphasize that it's a mobile only feature.

The bottom bar indeed hides under the keyboard, which isn't great, but may be a reasonable trade off for people. Personally I wouldn't use this mode, but power to the people.

zefhemel:
Or actually perhaps `mobileMenuStyle`? Either that, or it should apply to both desktop and mobile, but not that having this same UI is great on desktop.

janssen-io:
Sounds good. My reason for naming this `menuClass` is that you could use it to more easily style the regular menu as well.
But then again, it already has the `sb-actions` class. So I think it's very reasonable to give it a more apt name for its current use. :D I'll update it right away!

janssen-io:
If we name the setting `mobileMenuStyle` then the value doesn't make much sense.
Shall I rename the styles (and thus the classes) to `hamburger` and `bottom-bar`?
Optionally, we can still prefix the class with `mobile` or something similar to indicate it's meant for mobile/responsive styling.

PR #1301, opened 2025-03-18
Don't spread page meta in page navigator

MrMugame:
This fixes a bug where page attributes could overwrite internal variables leading to unwanted behaviour

https://community.silverbullet.md/t/page-picker-always-opens-a-new-browser-tab-when-selecting-certain-pages-like-person-joe/1997

(Someday this should be properly typed)


zefhemel:
Awesome, thanks!

PR #1305, opened 2025-03-21
Fix indexing of missing table cells (fix #1104)

### LINKED ISSUE #1104: Table rows are indexed incorrectly when some cells are empty
The columns in a given row are assigned in order of their definition, skipping empty columns.

Minimal example:
~~~
|Description|Foo|Bar|
|-----------|---|---|
|Row 1      |A1 |B1 |
|Row 2      |   |B2 |

```query
table where page = @page.name
```
~~~

Produces the following output (notice that second row has "B2" assigned to "foo"):

|Description|Foo|Bar|
|-----------|---|---|
|Row 1      |A1 |B1 |
|Row 2      |   |B2 |

|tableref|ref|tag|tags|page|pos|description|foo|bar|
|--|--|--|--|--|--|--|--|--|
|Indexing issue@1|Indexing issue@45|table||Indexing issue|45|Row 1|A1|B1|
|Indexing issue@1|Indexing issue@67|table||Indexing issue|67|Row 2|B2||

Maarrk:
The missing cells are undefined, just like in the past for shorter rows. Refactored the indexing function a bit to make it easier to test, and added unit tests. Changed local imports to import map where applicable.

### DIFF
--- a/plugs/index/table.test.ts
+++ b/plugs/index/table.test.ts
@@ -1,0 +1,48 @@
+import { assertEquals } from "@std/assert/equals";
+import { extractObjects } from "./table.ts";
+import { parseMarkdown } from "$common/markdown_parser/parser.ts";
+
+Deno.test("Table object data", () => {
+  // Not checking the references, only that the object properties are assigned correctly
+  const tree = parseMarkdown(`
+|Description|Foo|Bar|
+|-----------|---|---|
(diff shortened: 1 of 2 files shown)


zefhemel:
Looks decent, and I'm not familiar enough with the table based indexing stuff myself (don't use it). But I'm going to trust you on this one.

PR #1307, opened 2025-03-21
Implement table formatting in Space Lua (close #12)

### LINKED ISSUE #12: Markdown table reformat
For queries, there’s [attempt to make markdown tables look “pretty”](https://github.com/silverbulletmd/silverbullet/blob/d57a9e79dec6d18338d0ab7a3e6726ff29c6db45/packages/plugs/query/util.ts#L54). It would be cool to have this functionality as a command basically reformatting the markdown table you’re in with the appropriate amount of spacing to make it look good.

So e.g. turn this:
```
|Name|Age|
+——-+——+
|Pete|38|
```

Into
```
|Name|Age|
+----+---+
|Pete|38 |
```

Or similar

Maarrk:
Aligns column separators as proposed to close #12.

Preserves alignment specifiers and takes them into consideration for column width (see also #847, #1293)

### DIFF
--- a/plugs/core/Library/Std/Table.md
+++ b/plugs/core/Library/Std/Table.md
@@ -1,0 +1,138 @@
+#meta
+Implements utilities for working with Markdown tables
+
+# Table
+## Commands
+
+```space-lua
+command.define {
+  name = "Table: Format",
+  run = function()
+    position = editor.getCursor()
+    tree = markdown.parseMarkdown(editor.getText())
+    tableNode = nodeParentOfType(tree, position, "Table")
+    if tableNode then
+      formatMarkdownTable(tableNode)
+    else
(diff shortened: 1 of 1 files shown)


zefhemel:
Oh my, you people 😆 

This very cool. In my mind I was still thinking how to efficiently expose tree traversals in Lua, but I suppose you can already just... do it. Let me pull this in as is for now. We can iterate on style and what is builtin and not later.

As to general style, I'd indeed go for keeping things as much `local` as necessary. Every function or variable not local will be accessible from anywhere in the space. This may or may not be desirable depending on the case. Things like traversal functions you'll likely want to make publicly accessible, but I'm not sure what that API should look like.

And unit tests: my thinking for now is that part of the "Lua, hacker mindset" culture is not to have them. But maybe we have to introduce them later when this whole thing blows up.

Issue #1325, opened 2025-03-30
[v2] [regression?] SB_INDEX_PAGE `{{today}}` no longer works


vxnick:
I've switched from the `edge` Docker tag to `v2` for a more consistent experience but this seems to have broken the `SB_INDEX_PAGE=Journal/{{today}}` that I have configured.

If I go to my Home page in SB, the page title is simply `Journal/{{today}}` without being evaluated to the date.

vxnick:
@zefhemel This change works but turns out it's only evaluated at startup, not dynamically.

I've tried in vain for a few hours with Cursor to get this working but I have to admit defeat. Can you give any pointers or take this over please?

I know some others use `SB_INDEX_PAGE` in v1 so I think it's a worthwhile change to make.

vxnick:
I've pushed another commit courtesy of Cursor that fixes dynamic evaluation.

vxnick:
Implemented with `space-lua` per https://community.silverbullet.md/t/edge-some-upcoming-breaking-configuration-changes/1927/13?u=vxnick

PR #1329, opened 2025-04-02
Add support for Lua templating in SB_INDEX_PAGE [#1325]

vxnick:
Example with Docker Compose (the `$$` escapes a single `$`):

```yaml
environment:
  - SB_INDEX_PAGE=Journal/$${os.date('%Y-%m-%d')}
```

### DIFF
--- a/cmd/server.test.ts
+++ b/cmd/server.test.ts
@@ -1,0 +1,78 @@
+import { assertEquals } from "@std/assert";
+import { LuaStackFrame, LuaTable } from "../common/space_lua/runtime.ts";
+import { luaBuildStandardEnv } from "../common/space_lua/stdlib.ts";
+
+// Helper function to set up a Lua environment with standard library functions
+function setupLuaEnv() {
+  const env = luaBuildStandardEnv();
+  const sf = new LuaStackFrame(env, null);
+  sf.threadLocal.set("_GLOBAL", env);
+  return { env, sf };
+}
+
+Deno.test("SB_INDEX_PAGE template interpolation", async (t) => {
+  await t.step("simple string - no interpolation", async () => {
+    const { env, sf } = setupLuaEnv();
+    const result = await env.get("spacelua").get("interpolate").call(
+      sf,
+      "Journal",
+      new LuaTable(),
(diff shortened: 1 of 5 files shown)


zefhemel (changes requested):
Hello! Thanks for working on this.

A few thoughts:

1. I had kind of hoped to keep any Lua execution from the server, but I see there's value in it in this particular case. So let's be pragmatic about, this is fine 👍🏻 We should document that the Lua environment here will be limited, it will not be able to call custom defined Lua functions for instance, just the standard library stuff.
2. In the way this is done, there's a lot of code duplication (AI generated?) it would be nice to centralize the rendering of the index page name.
3. Since the indexPage can't really change at runtime I think this Lua expression only really needs to be evaluated once on boot, no?

vxnick:
Thanks for the review, to your points:

1. That makes sense, I will add documentation around this
2. This is AI generated yes - I will have a closer look at what it’s done and look for opportunities to refactor
3. That was my original implementation but then (with my use case) `os.date` was only being evaluated on boot so when the next day came my index page would still point at the previous day

zefhemel:
Actually. I think there may be a way to achieve what you need without this PR. Let me noodle a bit with that.

zefhemel:
Not as simple as I would have liked, but seems to do the job. To use use it simply keep SB_INDEX_PAGE set to "index", this space lua will automatically redirect you to a custom page whenever you're sent there.

~~~
```space-lua
local function redirectToCustomIndex()
  if editor.getCurrentPage() == "index" then
    editor.navigate({kind = "page", page = "Bla/" .. date.today()})
  end
end

event.listen {
  name = "editor:init",
  run = redirectToCustomIndex
}
event.listen {
  name = "editor:pageLoaded",
  run = redirectToCustomIndex
}
```
~~~

vxnick:
Fantastic idea! I will give this a go, thanks

zefhemel:
If this works for you, I prefer this solution over your PR, this keeps the backend cleaner.

vxnick:
Works for me, thanks for coming up with this solution!

Issue #1351, opened 2025-04-22
Hashes in code blocks interpreted as titles in ToC


chuskywalker:
https://community.silverbullet.md/t/hashes-in-code-blocks-interpreted-as-titles-in-toc/2215

For example, this kind of md, which is a fenced code block (/code, ``` ``) with a common sh comment in it (“#”):

```
# This is a comment in a BSD configuration file:
resolvconf=NO
```

Gets interpreted as a level 1 tile in the ToC:

![Image](https://github.com/user-attachments/assets/d6f96b9f-b5d0-4567-b947-011616ae32d0)

For context, both “Networking” and “Firewall” are legit lvl1 titles.

Code blocks made with 4 spaces (not fencing) don’t cause that issue in the ToC.

Comment from Zef Hemel: "...I just “simplified” the implementation of the TOC in v2, and it no longer takes into account fenced code blocks. So this is a bug. Could you create a github issue for this?"

wbhouston:
I can take a look at this.

zefhemel:
Fixed with #1354

PR #1354, opened 2025-04-24
Fix hashes in code blocks interpreted as titles in ToC

wbhouston:
Use `markdown.parseMarkdown` and `markdown.renderParseTree` to rely on finding headers. This has the added benefit of providing the starting position for a header in the file.

https://github.com/silverbulletmd/silverbullet/issues/1351


zefhemel:
Thanks you! I knew ultimately we'd go back to using the actual AST, but I didn't have time to put thought into a Lua friendly API on that yet. But... in this case you don't really need fancy traversals, so this should be good. Thanks!

PR #1357, opened 2025-04-25
Adding support for hosting with URL prefix

### LINKED ISSUE #163: Support hosting at a URL path

flowkclav:
Closes #163. Adding support for hosting with URL prefix to make reverse-proxy setup on sub-path possible.

Starting silverbullet server specifying a "prefix" (or "context root") in environment variable `SB_URL_PREFIX` will cause the application to be hosted on the specified prefix. This allows sub-path configuration when setting up reverse proxies (e.g. `https://myserver/sb/` -> `http://localhost:3000/sb/` )

Summary of changes:
- server:
  - Hono configured to serve on the prefix, if specified
  - Prefix is removed from incoming URLs in server endpoints for the application to function normally
  - Prefix is added to outgoing redirects
  - Prefix is set as path in authentication cookies 
  - Injection of the prefix in the `base` element of `index.html` and `auth.html`
- client:
  - Switching from "absolute" URLs (i.e. host-relative, e.g. `/myresource`) to page-relative URLs (e.g. `myresource`)
  - Using `document.baseURI` instead of `location.origin|host` when constructing URLs
 
NOTES:
  - Running the formatter modified a file I did not otherwise touch, I did not include it in this PR.
  - Likewise, the test I added and modified pass but another test is failing, at least on my setup (`JSON utils => ./plug-api/lib/json.test.ts:4:6` fails with `Actual 2023-05-02 / Expected 2023-05-03`. Probably a timezone thing). I left it alone.


zefhemel:
Thanks for this. I need the time to properly tests this. I anticipate a lot of edge cases and places where being hosted at the root is implicitly assumed.

flowkclav:
Absolutely!  I am fairly new to silverbullet so it is very possible that I missed something.   As if to illustrate my point, I just realised I never tested in offline mode. I just did and sure enough, the service_worker.ts will need tweaking. And unlike the rest of the client code, it doesn't have access to document.baseURI to derive the prefix.   I'll try to take a closer if I have some more time.

zefhemel:
Thanks a lot for this effort, a lot of people have asked for it! I'm sure there may be subtle issues, but based on some testing this seems to work great!

PR #1361, opened 2025-04-27
Dynamic App Name Loading for Auth Page

lochel:
This pull request updates the `auth.html` file to dynamically fetch and display the application name from the manifest file.


zefhemel:
Thanks!

PR #1365, opened 2025-04-30
Fix index page naming during server initialization

lochel:
This pull request ensures that the correct name for the index page is used during server initialization. This is relevant only when the default name is overridden using the `SB_INDEX_PAGE` environment variable. In such cases, we should create `<SB_INDEX_PAGE>.md` instead of `index.md`, using the default template content.


zefhemel:
Awesome, thanks

PR #1366, opened 2025-05-02
Add notes on Space Lua Pattern Matching

wbhouston:
* Allow for escaped hyphens at the end of Space Lua patterns.
* Update the order of year, month, day in tests and examples.


zefhemel:
Thank you for this 🙏

PR #1369, opened 2025-05-03
feat: configure vim in CONFIG

MatthiasBenaets:
Alright, hear me out here.
I know what you are thinking; why do it via the CONFIG file when it can be done via VIMRC?

Well, the new Lua implementation is great, and I like customising SB with custom commands.
If I want to invoke these, I could use shortcuts, but then I'm limited to only some key bindings. I prefer custom vim keybindings.

To my knowledge, there is no way to map Vim keybindings to SilverBullet commands. (otherwise, this pull request is pointless)

So, this commit allows you to define a custom _ex_ command that invokes a SilverBullet command, which you can then map to a custom vim keybind.
Since the [CodeMirror Vim Mode API](https://codemirror.net/5/doc/manual.html#vimapi) also allows you to `map`, `unmap` and `noremap`, I've implemented these as well.

The CodeMirror API has some type inconsistencies, so it's not a 1 to 1 implementation; but the resulting functionality should be complete and consistent with expectations.

Consult the JSON schema and `Vim.md` on how to use it.

One thing that can be considered for the future, is to add a `Vim.mapclear` before running `loadVimRc` and `loadVimConfig`. If it is not cleared, old key bindings will stick around until a page reload. This can have unexpected behaviour when new bindings partially match old binding. 
You can't call mapclear in each function because it will leak and clear already mapped binding from the other function.
So in the future, these can maybe be combined into a single function/syscall. But I guess this is not high prio.

### DIFF
--- a/Library/Std/Config.md
+++ b/Library/Std/Config.md
@@ -142,4 +142,79 @@ config.define("actionButtons", {
   }
 })
 
+config.define("vim", {
+  description = "Vim mode configuration",
+  type = "object",
+  properties = {
+    unmap = {
+      description = "Keys to unmap",
+      type = "array",
+      items = {
+        oneOf = {
+          { type = "string" },
+          {
+            type = "object",
+            properties = {
+              key = { type = "string" },
+              mode = {
+                type = "string",
+                enum = {"normal", "insert", "visual"}
+              }
+            },
+            required = { "key" },
+            additionalProperties = false
+          }
+        }
+      }
+    },
+    map = {
+      description = "Custom mappings",
+      type = "array",
+      items = {
+        type = "object",
+        properties = {
+          map = { type = "string" },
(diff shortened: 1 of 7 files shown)


MatthiasBenaets:
@zefhemel, I fully agree.
Personally, I don't need the VIMRC file anymore. I've replaced it with the CONFIG file, which does the same and more.
It would also be much better during development since you don't have to work around both functions. (for example, the mapclear issue I mentioned).

I already like how extendable this is by creating a custom command via `command.define` and mapping it to a custom _ex_.

The only reason why I did not remove it in this commit, is because many people probably rely on.
I think a lot of people are currently running SB on the v2/main branch, so it might cause some confusion.

Indeed, it would maybe be good to discuss this on the discourse or proceed with this in a minor version when there is the v2 release.

PR #1372, opened 2025-05-04
Add missing case to math.random and check arguments

lochel:
* add missing case `math.random(0)` which produces an integer with all bits (pseudo)random
* check arguments and throw `LuaRuntimeError` for invalid input


zefhemel:
I would almost ask to add tests, but... I see the problem there 😉

PR #1374, opened 2025-05-05
Generate version from git describe at build time

lochel:
Generate version from git describe at build time using:

```bash
git describe --tags --long --dirty
```

This makes it easy to identify the exact version a particular instance is running.

Calling `${system.getVersion()}` now returns a string like `"2.0-beta (0.10.4-209-gcdbee2fa)"`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ silverbullet
 deploy.json
 *.generated
 tmp_playground
+/version.ts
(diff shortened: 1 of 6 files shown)


lochel:
Just noticed that GitHub Actions aren't fetching the version string — probably because `git` isn't available in the environment. We may need to update the build pipeline, at least for releases and package builds.

I'm not going to work on that yet though, since I’d rather wait and see if these changes get accepted.

lochel:
@zefhemel Just wanted to make sure you saw my earlier comment - the Docker image doesn’t include the version as intended. We need to add git to the build pipeline to fix that.

PR #1377, opened 2025-05-06
Allow non-zero shell exit codes

lochel:
In the current implementation, an exception is thrown when a shell command exits with a non-zero code. While there may have been a good reason for this behavior, in my use case, non-zero exit codes are expected and should not result in an exception.

This pull request proposes removing the exception and returning the `code`, `stdout`, and `stderr` as-is. If there was a specific scenario this was meant to safeguard, I’m happy to adapt the pull request - but in that case, I’d need more context on what it’s protecting against.


zefhemel:
It seems like a good idea, but I'm ok to remove it.

PR #1379, opened 2025-05-07
fix: bundle external dependencies

costowell:
Silverbullet currently tries to download dependencies at runtime. I'm not sure what the intended reason is, but when deploying in immutable containers or on NixOS, this becomes a problem. I'm not experienced with deno so please let me know if I'm missing anything. If at all possible, it would be nice to see a minor version bump with pre-built files having this fix.


zefhemel:
There was  a reason for introducing this, if only I could remember...

PR #1382, opened 2025-05-08
Run tests commits and PRs to the "v1" branch

fflorent:
:warning: Some important notes:
 - I have not tested this patch even if I am comfortable this will work;
 - I think this commit should also be cherry-picked on the `v1` branch.


fflorent:
Deleting this PR in favor of #1383

Issue #1343, opened 2025-04-18
Ability to escape Lua expression directives


zefhemel:
For instance for use in templates, see: https://community.silverbullet.md/t/page-templates-in-v2/2037/21?u=zef

Potentially: `\${...}`

wbhouston:
> Markdown itself uses backslash for escaping characters, it may get lost while interacting with other programs, the forum etc.

I thought this would be a plus because the parse tree would mark these as `Escape` types that we could find and replace. I wrote [some code to do that](https://github.com/wbhouston/silverbullet-dev/commit/fbac37511a585ed5cc399a9e266db00e2e2bcff9) and encountered a problem where lua expressions inside WikiLink nodes are not parsed as such.

The string traversal in the current interpolation is probably the way to go, in case anyone else has the same thought.

marad:
> The [@marad](https://github.com/marad)'s solution is indeed nice trick. But I would also vote to have the escaping working too, it's far more intuitive for most users.

I agree. While my "trick" is working it's far from being a good solution. The main problem is that you are editing the code without any code-related features (i.e. autocomplete). Other is that SB by default changes the quote openning and closing signs (`"`) to their typographic equivalents that should be used for prose (`“` and `”` respectively). This messes the code up!

You can, of course, add the missing `$`, edit the code, and remove the `$`. But that's not the quality of experience that a polished tool like SB would have :)

PR #1386, opened 2025-05-08
Utility function to handle empty query results and templates

Maarrk:
More general implementation of #1359

This also documents current workaround for #1343

### DIFF
--- a/website/Space Lua.md
+++ b/website/Space Lua.md
@@ -57,6 +57,8 @@ A new syntax introduced with Space Lua is the `${lua expression}` syntax that yo
 
 For example: 10 + 2 = ${adder(10, 2)} (Alt-click, or select to see the expression) is using the just defined `adder` function to this rather impressive calculation.
 
+Note that you can have the text {lua expression} in your page and it won't be evaluated. So writing `${"$"}{lua expression}` can be used to “delay” evaluation by one pass, for example to escape expressions in templated content.
+
 ## Queries
 Space Lua has a feature called [[Space Lua/Lua Integrated Query]], which integrate SQL-like queries into Lua. Here’s a small example querying the last 3 modifies pages:
 
--- a/Library/Std/Pages/Maintenance.md
+++ b/Library/Std/Pages/Maintenance.md
@@ -10,7 +10,7 @@ ${template.each(query[[
   limit 20
 ]], template.new[==[
     * [[${ref}]]: broken link to [[${name}]]
-]==], "No aspiring pages, all good!")}
+]==])}
 
 # Conflicting copies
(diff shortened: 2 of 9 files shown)


website/Space Lua.md
```diff
@@ -127,10 +129,17 @@ There's a magic `_CTX` global variable available from which you can access usefu
 # Lua implementation notes
 Space Lua is intended to be a more or less complete implementation of [Lua 5.4](https://www.lua.org/manual/5.4/). However, a few features are (still) missing:
```
wbhouston, line 130: I think the lead in here should be more inclusive of the following headers. Maybe:

> Here are some of the expected differences and planned and not planned features:

website/Space Lua.md
```diff
@@ -127,10 +129,17 @@ There's a magic `_CTX` global variable available from which you can access usefu
...
 
-* `goto` and labels (not planned, goto considered harmful)
-* coroutines (not planned, not useful in the SilverBullet context)
-* _ENV (planned)
-* Full metatable support (only partial now, planned)
+## Differences
+* empty table `{}` tests false, while in [Lua 5.4](https://www.lua.org/manual/5.4/manual.html#3.3.4) "All values different from **nil** nad **false** test true". However this has a really nice interaction with empty [[Space Lua/Lua Integrated Query|query]] results
```
wbhouston, line 133: Typo on `nad` in the docs quote.

zefhemel:
Hmm. Need to think if the bug isn't really the incompatibility with Lua here. Not sure if we should double down on this being useful.

website/Space Lua.md
```diff
@@ -127,10 +129,17 @@ There's a magic `_CTX` global variable available from which you can access usefu
...
+
+## Planned
+* _ENV
+* Full [metatable](https://www.lua.org/manual/5.4/manual.html#2.4) support (only partial now)
+* Complete [[API/string]] API (some patterns in `gmatch` don’t work correctly)
```
wbhouston, line 138: I added a section to `[[API/string]]` about how Lua patterns and Space Lua patterns have some key differences, not just for `gmatch`. What do you think about removing:

> (some patterns in `gmatch` don’t work correctly)

and adding a bullet to the Differences section above, maybe:

> * lua patterns are translated to Javascript regex, so there are some [[API/string|key differences]].

wbhouston (commented):
I think this will be a nice change! I added some minor comments.

zefhemel:
I like the `some` name and approach. I wouldn't do weird metatable operator hacks (for now). I think being as Lua-y as reasonable is a good way to bring people in and to also allow them to use gained skills in other Lua environments.

zefhemel:
Very nice. I like this solution. And indeed I hope/assume people will not have used the extra template.each argument yet, and if so they'll figure it out 😄

PR #1387, opened 2025-05-08
Render a predefined string for empty markdown table instead of pipe characters

Maarrk:


zefhemel:
Yes thank you. This has been bothering me too.

PR #1388, opened 2025-05-08
Fetch tags and history for proper version string generation

lochel:
This fetches tags and history to enable proper version string generation using git describe, as introduced in #1374.


zefhemel:
Right, this actually works now. Just saw it in my version. Thanks! 👍🏻

PR #1390, opened 2025-05-09
Fix string.find

lochel:
This is an attempt to fix this problem: https://community.silverbullet.md/t/test-if-string-find-outputs-nil-not-working-as-expected/2342


zefhemel:
Thanks!

Issue #1316, opened 2025-03-25
Query Order By sorts A-Z then a-z


iamdangry:
Queries using `Order By` sorts A-Z then a-z, unsure if this is expected behaviour? Observed in 0.10.4 (legacy and LUA queries), Edge and v2.

**Space** comes before **index**, but **space** will come after **index**
![Image](https://github.com/user-attachments/assets/e3de9db6-35c1-45db-866b-df51050364c2)

Issue #614, opened 2024-01-02
ÄÖÜ german umlauts wrong sorting


v411e:
Example:
Multiple query objects are ordered by name asc

**Current sorting implementation**
- Arzt
- Arbeit
- Bus
- Zeit
- *Ärzte*

**Correct sorting**
- *Ärzte*
- Arzt
- Arbeit
- Bus
- Zeit

According to [Wikipedia](https://en.wikipedia.org/wiki/Alphabetical_order#Language-specific_conventions), the german umlauts can be interpreted for sorting like this:
- Ä = Ae
- Ö = Oe
- Ü = Ue
- a = ae
- ö = oe
- ü = ue

wizzor:
It should be noted that speakers of various languages will have a different preference.

Finland and sweden use WXYZÅÄÖ, so umlauts are placed at the end of the alphabet.

PR #1391, opened 2025-05-09
Order strings in queries using collation algorithm

Maarrk:
Configurable with queryCollation, default value maintains old behaviour. Added schema, example settings, and documentation to query page.

Close: #614
Close: #1316

### DIFF
--- a/common/space_lua/query_collection.ts
+++ b/common/space_lua/query_collection.ts
@@ -10,6 +10,7 @@ import { asyncQuickSort } from "$common/space_lua/util.ts";
 import type { DataStore } from "$lib/data/datastore.ts";
 import type { KvKey } from "@silverbulletmd/silverbullet/types";
 import type { KvPrimitives } from "$lib/data/kv_primitives.ts";
+import type { QueryCollationConfig } from "@silverbulletmd/silverbullet/type/client";
 
 export function buildItemEnv(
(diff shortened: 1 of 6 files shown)


common/space_lua/query_collection.ts
```diff
@@ -97,6 +97,20 @@ export async function applyTransforms(
...
+      false, // Old behaviour, also default sort in Lua and JavaScript
+    );
+    const locale: string | undefined = client.config.get(
+      "queryCollation.locale",
+      undefined, // Same as not passing this optional argument
+    );
+    const options: object | undefined = client.config.get(
+      "queryCollation.options",
+      undefined,
+    );
```
zefhemel, line 111: Relatively minor, but could you please just do a `client.config.get("queryCollation")` once to fetch the whole object (with an empty object as a default value, probably) and read from that instead of 3 separate calls?

zefhemel: Ah and wow indeed, you should not be using `client` here directly, but how should you pass in configuration then...

zefhemel: Ok, not great (and can't really test this now) but what will probably work is to navigate through LuaStackFrame: `sf.threadLocal.get("_GLOBAL").get("config")` to access the Lua version of the API

PR #1392, opened 2025-05-09
Add optional priority attribute to PageCreatingContent

lochel:
This PR introduces a way to override existing event listeners by adding an optional `priority` attribute to event callbacks. If multiple callbacks respond to the same event, the one with the highest priority is chosen.

I may have missed a more obvious or built-in solution, but this approach worked well in my case - and since the `priority` attribute is optional, it doesn't break any existing behavior.

Here's an example of how this can be used to override the default tag page:

```lua
event.listen {
  name = "editor:pageCreating",
  run = function(e)
    if not e.data.name:startsWith("tag:") then
      return
    end
    
    local text = "My custom tag-page\n"
    
    return {
      text = text,
      perm = "ro",
      priority = 1,
    }
  end
}
```

The main drawback is performance. Right now, all potential page variants are computed before selecting one, which isn’t ideal. It’s not an issue for me at the moment, but it could be optimized in the future.

### DIFF
--- a/plug-api/types.ts
+++ b/plug-api/types.ts
@@ -203,6 +203,8 @@ export type PageCreatingEvent = {
 export type PageCreatingContent = {
   text: string;
   perm: "ro" | "rw";
+  // Default 0, higher is higher priority
+  priority?: number;
 };
 
 export type SlashCompletionOption = {
--- a/web/client.ts
+++ b/web/client.ts
@@ -1274,13 +1274,39 @@ export class Client {
           "editor:pageCreating",
           { name: pageName } as PageCreatingEvent,
         ) as PageCreatingContent[];
-        if (results.length === 1) {
-          doc.text = results[0].text;
-          doc.meta.perm = results[0].perm;
-        } else if (results.length > 1) {
-          console.error(
-            "Multiple responses for editor:pageCreating event, this is not supported",
+        if (results.length > 0) {
+          // find the result with highest priority
+          const highestPriority = Math.max(
+            ...results.map((r) => r.priority ?? 0),
           );
+          // check that there is only one result with lowest priority
+          const highestResults = results.filter(
+            (r) => (r.priority ?? 0) === highestPriority,
+          );
+          if (highestResults.length > 1) {
+            // multiple results with lowest priority, show error
+            this.flashNotification(
+              `Multiple responses for editor:pageCreating event, this is not supported`,
+              "error",
+            );
+            console.error(
+              "Multiple responses for editor:pageCreating event, this is not supported",
+            );
+            if (previousPath && previousRef) {
+              this.ui.viewDispatch(
+                previousRef.kind === "page"
+                  ? { type: "page-loaded", meta: previousRef.meta }
+                  : { type: "document-editor-loaded", meta: previousRef.meta },
+              );
+            }
+
+            return;
+          }
+          const highestResult = results.find(
+            (r) => (r.priority ?? 0) === highestPriority,
+          );
+          doc.text = highestResult!.text;
+          doc.meta.perm = highestResult!.perm;
         }
       } else {
         this.flashNotification(


lochel:
By the way, this is what I actually wanted to implement:

```lua
event.listen {
  name = "editor:pageCreating",
  condition = function(e)
    return {
      condition = e.data.name:startsWith("tag:"),
      priority = 1,
    }
  end,
  run = function(e)
    local text = "My custom tag-page\n"
    
    return {
      text = text,
      perm = "ro",
    }
  end
}
```

The idea is that the `condition` function would be evaluated first for all registered callbacks. Based on the results, only the most relevant `run` function would be executed to compute the page content. This approach would be more efficient overall.

That said, it would also make the space-lua snippet more complicated, which might be a downside. I'm not sure which tradeoff is more important here - simplicity or efficiency.

Unfortunately, I don't know TypeScript or the internals of SilverBullet well enough to implement this myself. But if this would be the preferred solution, I'd be happy to give it a shot with a bit of guidance from someone more experienced.

wbhouston:
Alternatively, what if the [Library tag page](https://github.com/silverbulletmd/silverbullet/blob/b7f5fc174bb4e90daf4f49670b6d123410c4c4b5/Library/Std/Tag.md) was rewritten to include a priority and a separate function to handle the page construction? I feel like that would match the convention for widgets more closely. Something like:

```lua
-- priority: 10
pageEventHandlers = pageEventHandlers or {}
pageEventHandlers.tagPage = function(name)
  -- a similar function to the original that can be overwritten by users
  return {
    text = text,
    perm = "ro",
  }
end

event.listen {
  name = "editor:pageCreating",
  run = function(e)
    if not e.data.name:startsWith("tag:") then
      return
    end

    return pageEventHandlers.tagPage(e.data.name)
  end
}
```

`pageEventHandlers` is a little verbose, so that could use some work. I don't think `templates` is the right location, though, because this isn't returning a template.

zefhemel:
And a third option is to use the config system. Naming here is the hardest always hard, but we could have a `disabledBuiltinPages` config with a `tag` and `search` key that once set to `true` would simply not listen to the event.

Something along the line of 

```lua
config.get("disabledBuiltinPages.tag") or event.listen { 
   -- current tag page implementation
}
```

Also not in love with this solution, though...

lochel:
I think @wbhouston's proposal fits best within the current SilverBullet `Library/Std` structure. I'd just suggest forwarding the entire event argument to the handler for maximum flexibility:

```space-lua
-- priority: 10
pageEventHandlers = pageEventHandlers or {}
pageEventHandlers.tagPage = function(e)
  if not e.data.name:startsWith("tag:") then
    return
  end

  local text = "..."

  return {
    text = text,
    perm = "ro",
  }
end

event.listen {
  name = "editor:pageCreating",
  run = function(e) return pageEventHandlers.tagPage(e) end,
}
```

wbhouston:
My preference would be to have it both ways: a way to override the built in function to build the tag page and an approach to prevent the listener from being registered.

Setting a higher priority for the config is the only part of @zefhemel 's proposal that I am hesitant about. I might be in the minority on this, but--outside of changes in the code base--I ignore script priority. If there aren't issues with setting *all* config to a high priority, we could update the documentation to include that as a best practice and add a slash command for config boilerplate.

There are two other options I thought of:

1. Add an event API method to remove listeners. This would also provide more control for plug/third party event listeners. My memory of how listener removal works in JS is that all events are removed for a specific event type, but it might be possible to identify listeners and target individual listeners.
2. Programmatically treat all config changes as higher priority. I don't see how this would possibly be implemented in a sane way, but I thought I'd bring it up.

lochel:
Based on the example from my previous post, here's how you can unregister the tag page listener:

```lua
pageEventHandlers.tagPage = function(e)
  return
end
```

wbhouston:
That's subtly different from what zef proposed. That is not unregistering the event listener, it is override the function that is called when the event is handled. Zef's proposal:

```
config.get("disableBuiltinPages.tag") or event.listen { 
```

would never call `event.listen` in the first place.

Issue #1402, opened 2025-05-17
Admonitions written in all caps is not recognize.


queeup:
```
> [!note]
> This is recognize as an admonition.
```

```
> [!NOTE]
> This is NOT recognize as an admonition.
```

The same goes for `[!warning] / [!WARNING]`

https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts

PR #1409, opened 2025-05-24
fix: case insensitive admonitions by default

MatthiasBenaets:
Fixes: #1402 

I don't really think it's a true valid issue since it can easily be overwritten using a custom `space-style` that is case-sensitive. But I can understand it can be a bit confusing since the docs point to a GitHub discussion that uses capitalization.

The CSS attribute is now explicitly set to be case-insensitive (this is also reflected in the updated docs), and the admonition attribute name has been standardized to lowercase to ensure the correct icon appears.

It's true that we lose case sensitivity here, but I don’t see a compelling reason for users to define multiple admonitions with the same name but different capitalizations.


zefhemel:
Works for me, thanks!

PR #1410, opened 2025-05-26
Update widget header to not overflow container

wbhouston:
- Increase the left, right, and top margins from -15px to -10px to match the container's padding.
- Add a 10px margin below the header to create consistent margin around the contents.
- Reduce horizontal padding to 10px.
- Increase bottom padding to 15px to match the top.


zefhemel:
Thanks!

Issue #1336, opened 2025-04-14
v2: Problems with Sync mode when there is no network connection


zeus-web:
I use Silverbullet on an Android phone (among different desktop systems). But I run into problems as soon as there is no network connection. I thought, that should be no problem with sync-mode.

I tried with different browsers: Chrome, Firefox and DuckDuckGo browser (where I'm not sure on which it is based).
I tested Silverbullet is either opened via browser directly or is "installed as app" (from the browsers).

As long as there is a network connection everything is fine. However, if I switch to "Airplane mode" on the mobile, the trouble begins:

Directly after switching to Airplane mode, most of Silverbullet is working fine. However, if I open a tag-page ("tag:blabla") page, it takes about 30 seconds until it is displayed. That happens on every tag page.
Switching to other pages is no problem, however. Even pages with queries or scripts on them.

But even worse:
As soon as I "reload" the browser or close the "app" and reopen, I only get a white page. 
If I put the app in background and foreground again, there is no problem

As there is no GUI element at all, no button or whatever, I can't try to "reload UI" or "reload system", either. 

Looks like Silverbullet needs a connection on first start after the page/browser/"app" was closed once?

During my tests, I **once** saw a popup saying "could not load config" and after that a black page saying "offline", only.

As far as I understand, working completely offline (at least from time to time) should be fine...

zeus-web:
Well... After some experiments I found out, that this behavior is NOT the same on latest stable with sync-mode enabled:

With latest stable: after the app was closed and reopened, I see the blank page, but this time with the button-bar showing (but not working).
After I wait about 10 seconds my testpage shows up . Looks like it tries to connect until a timeout occurs (I'm fine with that).

However in v2 the page never comes up, stays a white page

zefhemel:
Ok, there's different issues here but I'll look into it.

papadopa:
I am facing the issue of the app not initializing if I'm offline. If I open it while online and then lose connectivity, it's ok.

Looking at the code a bit, I don't think it can initialize offline (assuming my read is correct):
* https://github.com/silverbulletmd/silverbullet/blob/main/web/boot.ts#L8 tries to fetch a client config which involves a `/.config` request
* the service worker intercepts the `fetch` call which doesn't hit the cache (https://github.com/silverbulletmd/silverbullet/blob/main/web/service_worker.ts#L13); side question why not have `/.config` cached and refreshed whenever online?
[...]

PR #1415, opened 2025-06-04
Fixing service worker and local storage conflicts when hosting multiple instances on same origin, different URL prefix but same space path

flowkclav:
(I was waiting on #1336 and the URL prefix PR to be resolved to tackle this since this depended on their outcome)

I've been using URL prefixes smoothly for many weeks now, except for this:

  When hosting _multiple_ Silverbullet instances on the same origin on different URL prefixes **BUT** with the same space folder/path - Docker container's path is always '/space' - conflicts and weird issues arise because the indexedDB name is derived only from the space path.

In this PR, to address the issue described above:
- Introduction of the URL prefix (if any) in deriving the database name
- Limiting the scope of the service worker to the URL prefix (if any) - otherwise multiple service workers can end up being active at once
- Tweaking the localStorage key a bit to include the URL prefix (and also adding the app name to play nicer with other potential applications on the same origin)

### DIFF
--- a/web/boot.ts
+++ b/web/boot.ts
@@ -1,7 +1,7 @@
 import { safeRun } from "$lib/async.ts";
 import { Client, type ClientConfig } from "./client.ts";
 
-const configCacheKey = "config";
+const configCacheKey = `silverbullet.${document.baseURI}.config`;
 
 safeRun(async () => {
   // First we attempt to fetch the config from the server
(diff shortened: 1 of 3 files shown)


zefhemel:
Awesome, thanks!

zefhemel:
If you're interested, it probably shouldn't be THAT difficult anymore to somehow make a single SB instance host multiple spaces given all this pre-work that you did. If that's something you'd like to tackle, that's be super awesome.

flowkclav:
I'll admit that the thought of multi-space on a single instance does cross my mind every time I recreate my 3 separate Docker containers :)   I can't promise I'll have time, but I'll let the idea percolate in my brain and if I wake up one morning thinking "hey, what if it was done like this?" and can't fight the urge to try I might look into it :)

zefhemel:
I've written up some ideas on how this could work here: https://github.com/silverbulletmd/silverbullet/issues/1417

PR #1430, opened 2025-06-19
upd: extend mentions for additional context

mschmidtkorth:
Hello, this PR extends the number of lines shown in the references/mentions of a given page. It attempts to keep the page name centered with content left/right to it, while not including subsequent blocks.

_Rationale:_ With the current single line it becomes difficult to see in which context a page has been mentioned.

_Note:_ The existing algorithm is kept to cover for edge cases (e.g. very short references).

image


zefhemel:
Beatiful, thanks!

PR #1437, opened 2025-06-21
Improvements for Modals

MrMugame:
Modals had a background and border before, which made it hard to e.g. achieve rounded edges. I see no reason for this. If a modal should have a border/background the developer should add them themselves.
Also the check for rendering panels didn't allow for a `mode` of `0`, which was kind of weird imo. Took me a second to find that.


zefhemel:
Nice. And regarding the `0` — welcome to the wondrous world of JavaScript...

PR #1439, opened 2025-06-22
feat: collapsible mention section

mschmidtkorth:
Hello, this makes the mention section collapsible by click (it is collapsed by default, could be changed). It also shows the total mentions (could be removed).

Note there is no change to the loading behavior (posssible async loading etc.). I didn't investigate this because I never really had long loading times :) 

_Rationale:_ If you have many mentions, it unnecessarily bloats the page.

image

### DIFF
--- a/Library/Std/Widgets.md
+++ b/Library/Std/Widgets.md
@@ -140,9 +140,16 @@ function widgets.linkedMentions(pageName)
     order by page
   ]]
   if #linkedMentions > 0 then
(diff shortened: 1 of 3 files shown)


mschmidtkorth:
@zefhemel I noticed I missed fetching the latest merged changes – this has now been updated and should be ready to merge. I've also slightly improved the UI (buttons, icons, no more scrollbar etc.):

image

PR #1440, opened 2025-06-23
feat: floating Table of Contents

mschmidtkorth:
Hello, this PR replaces the static ToC widget at the top of a page with a floating, persistent navigation panel that follows users as they scroll through documents. I did touch the original ToC code so please have a good look :)

**Note** Let me know if this is better suited as a plugin… putting it as draft as I noticed some weird behavior with widgets being killed 'randomly' jumping to different AST sections.

_Rationale:_ You can quickly jump through the document without having to scroll back to the top.

**Implementation**
- Replaces `widgets.toc()` with `widgets.floatingToc()`
- Uses HTML instead of Markdown for better styling control (via `web/styles/main.scss` classes)
- Is responsive incl. mobile
- Hides border frame and refresh button of the original implementation (`class="sb-lua-top-widget"`) via `.toc-hidden-container` class, which might not be the best solution
- Can still be disabled via `disableTOC` frontmatter as built on top of existing implementation, but it is now a string property. I've kept the property name for legacy purposes, but would have preferred `displayTOC = true|false|always`
- Keyboard navigation (not yet committed)
- Search (not yet committed)
```yaml
---
  pageDecoration:
    disableTOC: true   # Default - shows TOC (opacity 0.1, hover to 1.0)
    disableTOC: false  # Disable TOC completely
    disableTOC: never  # Always visible TOC (opacity 1.0); example use case: index pages like https://silverbullet.md
---
```

**Possible enhancements**
- Indicate current heading (would require more extensive JS)
- Keyboard shortcuts to navigate to next/previous heading (again JS)

**Before**
toc3

**After**
(peeking but not hovered; deliberately barely visible; `disableTOC: false|never`)
image

(hovered)
image

(mobile)
mobile

### DIFF
--- a/web/styles/main.scss
+++ b/web/styles/main.scss
@@ -321,6 +321,18 @@ body {
     &:hover {
       transform: translateX(0px);
     }
+
+	&.floating-toc-always-visible,
+    &:not(.floating-toc-always-visible) {
+      width: 45px;
+      height: 45px;
+      right: 12px;
+
+      &:has(.toc-toggle:checked) {
+        width: calc(100vw - 24px);
+        max-width: 280px;
+      }
+    }
   }
 
   // Mobile collapsible for both always-visible and default ToC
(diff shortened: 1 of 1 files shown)


ilkersigirci:
@zefhemel could you please clarify why this issue was closed? I am looking to place the Table of Contents in the sidebar rather than the default Top View.

I have reviewed the documentation, GitHub issues, and the official forum, but I haven't found a method to achieve this outside of this PR. Could you please point me in the right direction? Thank you in advance.

PR #1441, opened 2025-06-23
fix: expose markdownToHtml

mschmidtkorth:
Hello, `markdownToHtml()` exists in Lua but is not exposed as a syscall for TypeScript plugins (stumbled over it [here](https://silverbullet.md/Library/Std/Export)).

This PR:
  1. Adds `markdown.markdownToHtml` syscall in `web/syscalls/markdown.ts` which converts a markdown string to HTML
  2. Adds the corresponding TypeScript API function in `plug-api/syscalls/markdown.ts`

  This ensures API parity between Lua and TypeScript environments, allowing plugins to render markdown
  content to HTML as documented.
  
  Do let me know if it is not expected to be exposed!


zefhemel:
Thanks!

PR #1443, opened 2025-06-25
Fix day of the week calculation

lochel:
I am pretty sure this was working some weeks ago ...


zefhemel:
I revised or "fixed" some stuff in dates, perhaps it regressed

PR #1445, opened 2025-06-30
Redirect to auth page if ping returns 401

makeshared:
Fix message thrown on 401/403 status in authenticatedFetch.

The client tests this string for an exact match in init() and if it's different assumes the server is offline rather than (as intended) doing a redirect to the auth page.


MrMugame:
Ultimately this is probably better solved using a custom error type, but nobody got time to implemented that lol

makeshared:
> Ultimately this is probably better solved using a custom error type, but nobody got time to implemented that lol

Agree - or maybe could just use the HTTP status (401, 403, ...) for a start

PR #1447, opened 2025-07-01
upd: group commands by category

mschmidtkorth:
Hello, this PR groups commands by category while maintaining FuseJS ordering. Category headers are sticky upon scrolling. Supports current search behavior (e.g. "New Window" or "Editor: New Window") and has high weight for category matching.

_Rationale:_ Less redundant information, easier to skim through commands, more aesthetically pleasing.

**Before**
image

**After**
image

### DIFF
--- a/web/components/filter.tsx
+++ b/web/components/filter.tsx
@@ -216,7 +216,7 @@ export function FilterList({
             const groupedItems = groupOptionsByCategory(matchingOptions);
             let optionIndex = 0;
 
-            return groupedItems.map((item, idx) => {
+            return groupedItems.map((item) => {
               if (item.type === 'category') {
                 return (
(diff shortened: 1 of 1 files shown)


MrMugame:
I would be interested in how this performs from a usability standpoint. Because the best search results aren't necessarily grouped by category like this, so I'm unsure if this maybe makes the search worse. It also reduces the information density 7 commands vs 4 commands in the same visual space. Unsure if better or worse for usability.
I agree in that it is definitely more appealing visually, but I think we shouldn't let that cut into usability.

zefhemel:
I see the screen use concern, but I do find it very appealing visually. Let's go with this and see how people like it in practice. Thanks @mschmidtkorth !

PR #1448, opened 2025-07-02
upd: page picker hierarchy

mschmidtkorth:
Hi, this is a very minor, visual only change for 'nested' pages to see the hierarchy more easily. Could keep the /, but I think the spaces help the eye.

**Before**
image

**After**
image

### DIFF
--- a/web/components/page_navigator.tsx
+++ b/web/components/page_navigator.tsx
@@ -128,7 +128,7 @@ export function PageNavigator({
           // Use the displayName or last bit of the path as the name
           name: pageMeta.displayName || pageMeta.name.split("/").pop()!,
           // And use the full path as the description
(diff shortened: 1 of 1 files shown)


nightscape:
An option would be to use ` / ` (slash with space before and after).
That would look more similar to what you need to write when you create the page.

mschmidtkorth:
@nightscape I'm not a fan of / purely from an aesthetic and typographical perspective, but I'm inclined to agree it would be better for consistency.

mschmidtkorth:
Replaced `>` with `/`.

MrMugame:
It's a little opaque to me. It kind of obfuscates the actual name.

The better solution in my opinion would be to wrap the slashes in spans, which could be styled however you want with space styles. This would allow for anything `>`, ` / `, etc.

PR #1449, opened 2025-07-02
feat: page hierarchy overview

mschmidtkorth:
Hello, this PR adds a bottom widget (after Linked Mentions) to display page hierarchy, if the current page is part of it. It always shows the immediate parent and any children for easy navigation, with each page being clickable individually.

**Note** Collapsible is the same as for #1439, breadcrumbs as for #1448

image

### DIFF
--- a/Library/Std/Widgets.md
+++ b/Library/Std/Widgets.md
@@ -187,7 +187,7 @@ function widgets.pageHierarchy(pageName)
           table.insert(linkParts, "<a href=\"" .. currentPath .. "\" data-ref=\"" .. currentPath .. "\">" .. segment .. "</a>")
         end
 
-        html = html .. "<li>" .. table.concat(linkParts, " > ") .. "</li>"
+        html = html .. "<li>" .. table.concat(linkParts, " / ") .. "</li>"
       end
     end
 


mschmidtkorth:
Changed `>` to `/` for consistency.

Library/Std/Widgets.md
```diff
@@ -123,6 +123,92 @@ event.listen {
...
+
+  -- Add descendant pages
+  for _, page in ipairs(descendantPages) do
+    local displayName = page.name:gsub("/", " / ")
+    table.insert(hierarchyItems, "[[" .. page.name .. "|" .. displayName .. "]]")
+  end
+
+  if #hierarchyItems > 0 then
+    -- HTML instead of Markdown to make it collapsible
+    local html = "<div class=\"collapsible-hierarchy collapsed\">" ..
```
MrMugame, line 164: Wouldn't the multiline string syntax be better suited here? `[[ ... ]]`

web/boot.ts
```diff
@@ -113,6 +113,31 @@ safeRun(async () => {
...
   // @ts-ignore: on purpose
   globalThis.client = client;
+
+  // Widget utilities for collapsible sections
+  // @ts-ignore: on purpose
+  globalThis.sbWidgets = globalThis.sbWidgets || {};
+  
+  // Toggle function for collapsible hierarchy
+  // @ts-ignore: on purpose
+  globalThis.sbWidgets.toggleHierarchy = function(header: HTMLElement) {
```
MrMugame, line 123: This seems a little misplaced imo. This could just be in the widget code as a script tag, couldn't it?

web/cm_plugins/lua_widget.ts
```diff
@@ -181,7 +184,18 @@ export class LuaWidget extends WidgetType {
         preserveAttributes: true,
       }, this.client.ui.viewState.allPages));
     }
-    if (html) {
+
+    // Combine HTML from both sources: Markdown and HTML
+    if (htmlFromHtml || htmlFromMarkdown) {
```
MrMugame, line 189: This seems like weird behaviour to me, it's not intuitive.
Imo, if the user decides he wants to export raw html, but still include markdown they should use the `markdownToHtml` syscall. 90% of the time just concatenating it in a div is not going to be what you want.

web/styles/main.scss
```diff
@@ -159,3 +159,63 @@ body {
 .sb-markdown-toolbar:hover {
   opacity: 1;
 }
+
+/* Collapsible hierarchy widget styles */
```
MrMugame, line 163: Could be put into a style tag in the lua. Imo, keeping as much out of SB itself is the main objective

web/styles/theme.scss
```diff
@@ -302,3 +302,9 @@ html[data-theme="dark"] {
   --admonition-icon: url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M10.29 3.86 1.82 18a2 2 0 0 0 1.71 3h16.94a2 2 0 0 0 1.71-3L13.71 3.86a2 2 0 0 0-3.42 0z"></path><line x1="12" y1="9" x2="12" y2="13"></line><line x1="12" y1="17" x2="12.01" y2="17"></line></svg>');
   --admonition-color: #ff9100;
 }
+
+/* Chevron icons for collapsible sections */
+html {
```
MrMugame, line 307: *See the other comments*

MrMugame:
Added my 2 cents

zefhemel:
I'm wondering if this should be included in the standard library or if it can just be a script to be shared on community.silverbullet.md. If it gets popular, we can include it in the standard distribution.

PR #1450, opened 2025-07-02
feat: keyboard shortcuts in action buttons

mschmidtkorth:
Hi, this PR provides the capability to show dynamic keyboard shortcuts in action button tooltips. It does so by implementing a tooltip component (rather than using the browser's native plain tooltips) and using the `commandName` prop to determine the command to dynamically derive the shortcut from. Provides user-controllable CSS variables (applied my theme in the example screenshot).

Open to feedback and ideas – might have overengineered this one…

I tried parsing 'run' to get the command rather than a separate prop, but as I don't know where the user defined their command (could be any markdown file) and only get the minified JS it didn't work out. 

**Before**
image

**After**
image

**Example Config**
```yaml
navigateHome:
    path: "./navigate.ts:navigateCommand"
    command:
      name: "Navigate: Home"
      key: "Alt-h"
      page: ""
```
```lua
config.set {
  actionButtons = {
    {
      icon = "home",
      description = "Go to index page",
      commandName = "Navigate: Home",  -- Shows "Alt h" in tooltip
      run = function()
        editor.invokeCommand("Navigate: Home")
      end
    }
  }
}
```

### DIFF
--- a/web/components/tooltip.tsx
+++ b/web/components/tooltip.tsx
@@ -1,0 +1,247 @@
+import type { ComponentChildren } from "preact";
+import { useRef, useState, useEffect } from "preact/hooks";
+import type { Command } from "../../type/command.ts";
+
+export interface TooltipProps {
+  text: string;
+  shortcut?: string;
+  macShortcut?: string;
+  commandName?: string;
+  commands?: Map<string, Command>;
+  position?: 'top' | 'bottom' | 'left' | 'right' | 'auto';
+  delay?: number;
+  disabled?: boolean;
+  className?: string;
+  children: ComponentChildren;
+}
+
+export function Tooltip({
+  text,
+  shortcut,
+  macShortcut,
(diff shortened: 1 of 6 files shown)


zefhemel:
This is nicely done! I do have one request for a follow-up PR, which is to somehow reuse this shortcut rendering code also in other places in the UI, I'm thinking in the command palette for instance. This will look nicer and will be more consistent.

zefhemel:
Ah now, dove a bit into the specifics a bit more. I see the problem now. Action buttons indeed are no longer directly tied to commands nor keyboard shortcuts... I see what you did there, but I'm wondering if it's worth all this code to get there. As I mentioned in the previous comment I do like the nice rendering of the keyboard shortcuts, but see more value in them in the command palette (this is what people actually look at), and _far_ less in the context of hovering over an action button.

Do you agree?

PR #1451, opened 2025-07-03
upd: defensive programming patterns (boot.ts)

mschmidtkorth:
Hello, this PR adds defensive programming/error handling to `boot.ts` (following a comment by @MrMugame). I know too little of Silverbullet's full architecture to make drastic changes, so I:

- Added proper null checks to prevent runtime errors 
- Implemented graceful service worker update notifications
- Added resilient config parsing with corrupted cache cleanup
- Improved network timeout handling with retry logic
- Added service worker registration error handling
- Made IndexedDB check non-blocking

### DIFF
--- a/web/boot.ts
+++ b/web/boot.ts
@@ -3,20 +3,67 @@ import { Client, type ClientConfig } from "./client.ts";
 
 const configCacheKey = `silverbullet.${document.baseURI}.config`;
 
+// Type declarations for global properties
+declare global {
+  var sbIndexedDBAvailable: boolean;
+  var pendingServiceWorkerUpdate: boolean | undefined;
+  var client: Client;
+}
+
+// Check IndexedDB availability early but don't block
+if (!globalThis.indexedDB) {
+  console.error(
+    "IndexedDB not available - app will have limited functionality",
+  );
+  // Set a flag that client can check
+  globalThis.sbIndexedDBAvailable = false;
+
+  // Still show alert but don't block execution
+  setTimeout(() => {
+    alert(
+      "SilverBullet requires IndexedDB for full functionality. Some features may not work correctly. Please use a recent version of Chrome, Firefox (not in private mode) or Safari.",
+    );
+  }, 1000);
+} else {
+  globalThis.sbIndexedDBAvailable = true;
+}
+
+// Helper function for config fetch with timeout
+async function fetchConfigWithTimeout(
+  timeout: number,
+): Promise<Response | null> {
+  try {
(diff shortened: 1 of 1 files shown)


web/boot.ts
```diff
@@ -83,16 +151,34 @@ safeRun(async () => {
...
+        console.log("New service worker activated, notifying user...");
+
+        // Notify user through the client if it's available
+        // Note: This event might fire before client is initialized
+        if (
+          globalThis.client &&
+          typeof globalThis.client.flashNotification === "function"
+        ) {
+          globalThis.client.flashNotification(
+            "A new version is available. Please refresh when convenient.",
```
zefhemel, line 175: I'd add "a new version of SilverBullet" is available, so to make it clear it's not about page content for instance.

web/boot.ts
```diff
@@ -3,37 +3,99 @@ import { Client, type ClientConfig } from "./client.ts";
...
 const configCacheKey = `silverbullet.${document.baseURI}.config`;
 
+// Type declarations for global properties
+declare global {
+  var pendingServiceWorkerUpdate: boolean | undefined;
+  var client: Client;
+}
+
+// Check IndexedDB availability early but don't block
+if (!globalThis.indexedDB) {
```
zefhemel, line 13: I'd be surprised if anything works at all without IndexedDB in its current version. Perhaps the wording can be even stronger as in: please use a modern browser.

web/boot.ts
```diff
@@ -107,17 +193,24 @@ safeRun(async () => {
       "Not launching service worker, likely because not running from localhost or over HTTPs. This means SilverBullet will not be available offline.",
     );
   }
-  const client = new Client(
-    document.getElementById("sb-root")!,
-    clientConfig!,
-  );
+  // Validate root element exists
+  const rootElement = document.getElementById("sb-root");
+  if (!rootElement) {
```
zefhemel, line 198: This seems VERY defensive, no, since this element is actually baked into the index.html page itself. This seems unnecessary.

web/boot.ts
```diff
@@ -3,37 +3,99 @@ import { Client, type ClientConfig } from "./client.ts";
 
 const configCacheKey = `silverbullet.${document.baseURI}.config`;
 
+// Type declarations for global properties
+declare global {
+  var pendingServiceWorkerUpdate: boolean | undefined;
```
zefhemel, line 8: Can this not just be a top-level `let` variable? Since this seems only to be used inside boot.ts?

PR #1466, opened 2025-07-18
Add a copy button

MrMugame:
This adds a copy button to the button bar of lua block widgets. It copies the content of the widget to the clipboard. For markdown widgets the content is the trimmed markdown content, for html widgets the content is the html.
image

The second commit is only half related and cleans up some code, I can revert it if you want. It just removes some doubled types and unused properties.

(Btw: What really speaks against dropping codeWidgets :eyes: ?)

### DIFF
--- a/type/client.ts
+++ b/type/client.ts
@@ -20,16 +20,11 @@ export type Notification = {
 export type PanelMode = number;
 
 export type CodeWidgetContent = {
-  html: string;
+  html?: string;
   script?: string;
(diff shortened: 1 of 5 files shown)


zefhemel:
Thank you once again. As to about dropping codeWidgets, I agree it's probably legacy at this point. There are a few plugs that use it, but honestly using Lua widgets is likely the better solution.

PR #1470, opened 2025-07-24
Only `scrollIntoView` when selection changes

### LINKED ISSUE #1454: [Bug] The Command menu skips when scrolling
This happens when I do a type-ahead search in the Command menu. If I scroll further down the list the position will jump around. I'm not sure if this is because the list of commands is being lazy-loaded or some other reason. I looked at the output on the console but didn't see any messages that looked out of the ordinary.

I'm running "ghcr.io/silverbulletmd/silverbullet:v2" (last updated today) via Docker and am viewing in Vivaldi (7.5.3735.41 (Stable channel) stable (64-bit)) PWA. This has been happening since I installed though so probably a few weeks at this point.

MrMugame:
This fixed #1454 and [this post](https://community.silverbullet.md/t/page-picker-scrolls-to-selection-every-time-sync-occurs/2035/3)

I opted for the else if here, because the switch was just getting confusing if i didn't instantly return and we already had doubling cases.

### DIFF
--- a/web/components/filter.tsx
+++ b/web/components/filter.tsx
@@ -158,50 +158,39 @@ export function FilterList({
             return false;
           }}
           onKeyDown={(view, e) => {
-            switch (e.key) {
-              case "ArrowUp":
-                setSelectionOption(Math.max(0, selectedOption - 1));
-                return true;
-              case "ArrowDown":
-                setSelectionOption(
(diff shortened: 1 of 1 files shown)


zefhemel:
Awesome, I also noticed this happening but didn't get around to look at it. Thanks a lot!

PR #1471, opened 2025-07-24
Add `__call`

### LINKED ISSUE #1463: v2: Space Lua: Implement `__call` metamethod for tables
Please, implement `__call` metamethod for tables.

MrMugame:
This just adds the call method and thus fixes #1463

- I added a very small test
- It's nestable (Which I think isn't possible with `__index` right now, but should be implemented as it is part of the spec)
- Removed the `Attempting to index a non-table` error, as it was just plain wrong and always just showed `... non-table: null`. null cases now fall through to the call function and will just give a `Attempted to call nil as a function` error, which aligns with the behavior of the lua implementations I tested
- The `Attempting to call ${prefixValue} as a function` was also just doubled. It's handled in the `luaCall` method and evaluating the args before evaluating the callee aligns better with the common lua implementations

### DIFF
--- a/lib/space_lua/eval.ts
+++ b/lib/space_lua/eval.ts
@@ -345,12 +345,7 @@ function evalPrefixExpression(
         if (e.name) {
           selfArgs = [prefixValue];
           prefixValue = luaIndexValue(prefixValue, e.name, sf);
-          if (prefixValue === null) {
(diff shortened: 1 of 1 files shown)


zefhemel:
Thank you for adding this and cleaning this up 👍🏻

PR #1473, opened 2025-07-25
Make `stripIndent` an option for `template.new`

### LINKED ISSUE #1459: SBv2: Page Template expansion removes indentation
When instantiating a [Page Template](https://silverbullet.md/Library/Std/Page%20Templates) like this:

```
* Foo
    * Indented Foo
```

the resulting page will lose all indentation. That is, it will look like this:

```
* Foo
* Indented Foo
```

This doesn't happen in v1 where the resulting page keeps all indentation, as expected:

```
* Foo
    * Indented Foo
```

MrMugame:
This fixes #1459

I feel like just adding the option to template new is the best and most sensible option, even if it makes a core api a little more complex.

### DIFF
--- a/Library/Std/Template.md
+++ b/Library/Std/Template.md
@@ -20,15 +20,17 @@ function template.each(tbl, fn)
 end
 
 -- Creates a new template function from a string template
-function template.new(templateStr)
+function template.new(templateStr, stripIndent)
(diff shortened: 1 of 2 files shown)


zefhemel:
Yes this makes sense

PR #1474, opened 2025-07-27
Support mixing html and md widgets

lochel:
This addresses the issue of mixing html and md widgets, as reported in this community post: https://community.silverbullet.md/t/mixing-html-and-md-widgets/2652

I’m not deeply familiar with the codebase, so I approached this by testing a few different solutions and settled on the one with the smallest footprint. It resolves the issue effectively without introducing major changes.

That said, the widget code might benefit from a broader refactor, which could be tackled in a separate pull request by someone more familiar with the internals.

### DIFF
--- a/web/cm_plugins/lua_widget.ts
+++ b/web/cm_plugins/lua_widget.ts
@@ -174,7 +174,7 @@ export class LuaWidget extends WidgetType {
         trimmedMarkdown,
       );
 
-      html = parseHtmlString(renderMarkdownToHtml(mdTree, {
+      const mdHtml = parseHtmlString(renderMarkdownToHtml(mdTree, {
         translateUrls: (url) => {
           if (isLocalPath(url)) {
             url = resolvePath(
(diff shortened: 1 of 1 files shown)


MrMugame:
I'm going to be honest, I don't like it. It's seems very opaque, at least to me (e.g. if you have four hooks providing `md, html, md, html` in that order, it will be rendered as `html, html, md, md`). It's also [been suggested before](https://github.com/silverbulletmd/silverbullet/pull/1449/files/ff156a8ba06ece5def46cabbf8b8eea021f42769#diff-b91cd323e4d6b8c887dab234249f88ff4f819baa869427d5f6160460680407baR191-R194) as part of a bigger PR.

The usecase (as far as I can tell from your forum post and from the PR) seems to be to have multiple widgets with the `renderTopWidgets` hook. This hook already implements a [similar thing](https://github.com/silverbulletmd/silverbullet/blob/be2b6bf09ee14c42b6371862e0547d0315c63848/web/cm_plugins/top_bottom_panels.ts#L28-L47) for content of the same type. But this also seems like an ad hoc solution to the problem.
Imo it should just render multiple top widgets. I don't see a reason not to.

lochel:
Feel free to rewrite the widgets properly. I mentioned above that it would be a good idea indeed. 
Until then, this is a major improvement to the current status quo which simply drops some widgets.

zefhemel:
Thanks for attempting this, but I agree that @MrMugame's solution here #1476 is nicer so I've merged that one.

PR #1476, opened 2025-07-29
Allow multiple panel widgets

MrMugame:
This basically is my answer to #1474, it implements an `ArrayWidget` which just takes a callback for an array of widget content and renders the individual widgets. This class is used to render multiple top and or bottom widgets. Arguably this could be considered hacky, but the implementation really isn't that bad and decently small. The only real thing to consider is that it taps into the widgetCache and activeWidgets set, which was previously only used by lua widgets, but this shouldn't matter.
Another thing is, to return event results in order (the order the event handlers are traversed), this change shouldn't break anything for users (otherwise they are dependent on a race condition), but allows the user to use the typically lua priority to order the top or bottom widgets.
~~I also removed a dispatch in 88cc33ad57f2f664cdb597e2c7f7ec35816810fc, where I'm not exactly sure what it was for. It caused me issues, but removing it doesn't really change anything as far as I noticed. Maybe you can shed some light on that. If it's actually necessary I'm going to look into fixing my issues.~~
And I added a top margin to the bottom widget so it doesn't hug the text.

image

### DIFF
--- a/web/cm_plugins/lua_widget.ts
+++ b/web/cm_plugins/lua_widget.ts
@@ -211,9 +211,7 @@ export class LuaWidget extends WidgetType {
       );
       // Because of the rejiggering of the DOM, we need to do a no-op cursor move to make sure it's positioned correctly
       this.client.editorView.dispatch({
-        selection: {
-          anchor: this.client.editorView.state.selection.main.anchor,
-        },
+        selection: this.client.editorView.state.selection,
       });
     });
   }
--- a/web/cm_plugins/lua_widget.ts
+++ b/web/cm_plugins/lua_widget.ts
@@ -209,12 +209,6 @@ export class LuaWidget extends WidgetType {
           copyContent: copyContent,
         },
       );
-      // Because of the rejiggering of the DOM, we need to do a no-op cursor move to make sure it's positioned correctly
-      this.client.editorView.dispatch({
-        selection: {
-          anchor: this.client.editorView.state.selection.main.anchor,
-        },
-      });
     });
   }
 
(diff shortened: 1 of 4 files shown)


MrMugame:
Ok I now see why the dispatch exists. The cursor layer doesn't redraw after creating a widget, so the cursor sometimes stays in the wrong position and [the only way to trigger a redraw of the layer](https://github.com/codemirror/view/blob/3cea8dba19845fe75bea4eae756c6103694f49f3/src/draw-selection.ts#L84) is to either change the doc or selection.
~~First of all, that seems like weird behavior, I haven't found a thread about this in the codemirror forum, but I feel like it's common to draw a widget and the cursor being moved because of that~~. (It's probably because we are updating asynchronously, right?)
Second thing is that the bug I encountered wasn't actually caused by me it was there before. If you have a bottom or top widget, e.g. the linked mentions and it's offscreen and you try to select the document with `Ctrl-a` it suddenly deselects again (most of the time). This happens, because the bottom widget is rerendered for some reason and the rejiggering updates the selection, but only sets the anchor, so a selection, i.e. a range is lost (~~This normally isn't a problem~~ It is lol, see the video below). This can be easily fixed, by copying the complete selection. I did this in ac74d10
I still really don't understand why it rerenders the offscreen widgets tho?

https://github.com/user-attachments/assets/3bda1e52-e2b2-4b0a-a7d3-aecfd8995286

**tl;dr:** It's fixed now

zefhemel:
Great, yes this is more or less what I had in mind but didn't get to implement. Let me give it a quick test and then I'll merge it.

Issue #1043, opened 2024-08-18
ToC widget inheriting pageDecoration.prefix


cmdrSpectral:
TOC widget inherits _pageDecoration.prefix_ attribute. It gives a result I find improper. Having a prefix emoji for a page is one thing, but added to every header is:
- superflous (I can already see the emoji on the page title bar)
- untrue (those headers don't have emojis)
- makes using emojis for headers difficult (you see double emojis then)

Example below has a :brain emoji associated with #thinking tag, and you can see quite a lot of brains around:
<img width="833" alt="Screenshot 2024-08-18 at 10 28 40" src="https://github.com/user-attachments/assets/73810668-e9b4-4be6-9e5b-0ddff8ae1adb">

Shouldn't pageDecoration.prefix be ignored while using TOC?

If we had something like _pageDecoration.header.prefix_, that's fine for those who benefit from this decoration, but I don't see much use cases for my personal workflow.

imagoiq:
@cmdrSpectral have you found a workaround since last year?

Issue #1352, opened 2025-04-23
editor.getCurrentPageMeta() not returning tags and frontmatter on initial load


MatthiasBenaets:
Currently tested on v2.
When using editor.getCurrentPageMeta() inside a query or space-lua block, the function will initially only return the default page metadata.
Frontmatter metadata will not be made available.
On page reload or system reload, the correct metadata is returned.

Example
````
---
status: bug
---

a tag that is not escaped
\#problem

```space-lua
function metadata()
  return editor.getCurrentPageMeta()
end
```

Returns default metadata, after a page refresh/system reload will correctly display:
${metadata()}

Returns nil or nothing on initial page load:
${metadata().status}
${metadata().tags}
````

This is not really a big deal for tags since you can just query directly for a specific tag, but it is problematic in queries that match pages to the one opened using frontmatter metadata.

PR #1477, opened 2025-07-30
Fix page decoration prefix appearing in table of content's links

imagoiq:
Table of content's links contain page decoration prefixes when navigating between pages or when reloading the system (not on page refresh for some reasons, might be related to https://github.com/silverbulletmd/silverbullet/issues/1352).

This proposal avoids adding those prefixes when the link is having a `pos` in the file which indicates that it targets a heading. I think that it's an expected behavior as the heading could as well be prefixed with something (e.g., an emoji).

Note that it will as well change link's mention widget behavior.

closes https://github.com/silverbulletmd/silverbullet/issues/1043

### DIFF
--- a/web/markdown/markdown_render.ts
+++ b/web/markdown/markdown_render.ts
@@ -551,7 +551,7 @@ export function renderMarkdownToHtml(
         if (t.attrs!["data-ref"]?.length) {
           const pageRef = parseRef(t.attrs!["data-ref"]!);
           const pageMeta = allPages.find((p) => pageRef.page === p.name);
-          if (pageMeta) {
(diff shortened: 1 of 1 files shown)


MrMugame:
*Small Sidenote:*

The reason this doesn't happen for normal in document wiki links is, because it doesn't find the document meta. The document meta is [searched using the actual string url](https://github.com/silverbulletmd/silverbullet/blob/be2b6bf09ee14c42b6371862e0547d0315c63848/web/cm_plugins/wiki_link.ts#L74-L76) you type in (and it's never going to find e.g. the page `foo#Header`). This whole wikilink file is such a mess. Currently working on a PR to rewrite the whole ref system anyways.

PR #1479, opened 2025-07-31
Add MCP server

nightscape:


zefhemel:
I'm ok with building an MCP server for SB, but is there a way to do this separately and not built into the SB core?

nightscape:
You mean in a separate repository? That's basically what https://github.com/Ahmad-A0/silverbullet-mcp does.
I took the code and inlined it because it's much more convenient to have this integrated into a single server.

Maybe adding [Universal Tool Calling Protocol](https://www.utcp.io/) protocol support would be a middle ground?
Then the silverbullet repo would just need to expose a file that describes the available REST interfaces for the tools and then one could use [a UTCP-MCP translator](https://github.com/universal-tool-calling-protocol/utcp-mcp) to make these tools accessible to AI tools.

In order to be useful, one would need to create more REST routes for non-CRUD actions though, e.g. to
  - search notes
  - move note (otherwise the AI has to create a new note with the same content and delete the old one, which bears the risk of hallucinations during copying and wastes a lot of tokens)
  - Query metadata
  - Link graph analysis
  - Structured data queries

Would you be ok with adding such higher-level operations as REST endpoints?
Or maybe you have other ideas or plans of how to make AIs be able to interact efficiently with Silverbullet?

justinbadal:
I'm here to add that adding more routes would be much appreciated. This is the smoothest functioning MCP so far. I've had minimal to no issues using it with most LLMs (commercial or self-hosted). Thanks!

zefhemel:
My personal stance towards the AI industry is not very much on the positive side right now, and while I will not actively fight against any AI support/use in SilverBullet I will have limited personal interest to actively support it. However, if we add generally useful things to SB (not just for AI agents to use) I'd be happy with it, of course.

That said, I'm not sure the current architecture of SB can really support what you'd like. The things you mention like search, metadata query etc. are in the v2 model (where the backend is purely a "dumb" file store and knows nothing about what's in the files) are simply not available on the backend at all. All the indexing happens in the client, without the server being involved at all. To enable it, indexing would have to happen on the server (like it it did pre-v1) which would be a huge architectural pivot (back). Search is the same story, unless you'd implement it separately or with grep or something (which kind defeats the purpose of integrating it in SB).

Unless you have creative ideas on how this functionality may also run on the server?

zefhemel:
Since the server has now been rewritten in Go, this approach is no longer viable. I'll close it.

PR #1485, opened 2025-08-01
Recursively remove orphaned folders after file deletion

wzaldivar:
This PR adds a cleanup step after file deletion in DiskSpacePrimitives.deleteFile.

It traverses parent directories upward from the deleted file path, removing any empty folders until it reaches either a non-empty folder or the root space.

This helps prevent leftover empty directories (e.g. after renaming or deleting deeply nested files), which otherwise persist due to the current file-only deletion logic.

Details:

The method cleanOrphaned() uses Deno.remove() in a loop.

Stops on:

- ENOTEMPTY (non-empty directory)
- PermissionDenied
- unexpected errors

Skips missing directories (NotFound) to handle race conditions gracefully.

Cache is invalidated after deletion, as before.

### DIFF
--- a/lib/spaces/disk_space_primitives.ts
+++ b/lib/spaces/disk_space_primitives.ts
@@ -135,10 +135,40 @@ export class DiskSpacePrimitives implements SpacePrimitives {
     const localPath = this.filenameToPath(name);
     await Deno.remove(localPath);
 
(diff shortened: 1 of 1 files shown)


zefhemel:
Seems to work as advertised. Thanks!

PR #1486, opened 2025-08-02
Filter nullish values from dispatch

MrMugame:
Only noticed now, overlooked that functions have a default return. This should fix stuff like `tag:` pages not working anymore.

Before, sometimes it was deliberately compared to undefined and sometimes we checked if it's falsy. I now opted to only filter null and undefined, I think this is reasonable.


zefhemel:
Oh the joys of Javascript

lochel:
@MrMugame @zefhemel This change appears to break the widgets and doesn’t resolve the page creation event issue either.

PR #1487, opened 2025-08-04
Fix event dispatch

lochel:
This might fix #1486


MrMugame:
Totally, my bad. I don‘t even know how I accomplished that.

EDIT: Someone good at idiot prevention would probably change the return type to something pointing to `{}` or maybe even `LuaValue`, although that‘s just any again and events aren‘t limited to Lua

PR #1492, opened 2025-08-15
Make read only more robust

MrMugame:
Adds `EditorState.readOnly` so the page is *actually* read only and also makes the page read only if `client.clientConfig.readOnly` is set.


ToontjeM:
This works correctly in V2.

PR #1493, opened 2025-08-17
Update example Authelia resource regex to match literal dots

imchairmanm:
The current regular expressions for the example Authelia configuration don't escape the dot character, meaning they'd match requests for `/Xclient/manifestXjson` for instance.

This PR just escapes them so the expressions match literal dot characters, which I assume was the intention.


zefhemel:
Thanks!

Issue #1404, opened 2025-05-21
Regression on V2: link with header doesn't function properly.


LumenYoung:
Regression: I find that ![[something#header]] doesn't work on my instance. It renders the whole page instead of the content under header. This problem is confirmed reproducible in forum by others as well.

Version: v2 branch latest.

devonhk:
I don't see a [v2 branch](https://github.com/silverbulletmd/silverbullet/branches/all?query=v2). Can you share the actual commit?

LumenYoung:
> I don't see a [v2 branch](https://github.com/silverbulletmd/silverbullet/branches/all?query=v2). Can you share the actual commit?

No, what I mean by v2 is the docker tag that is constantly being updated with the newest commit from the master branch I think. So u can just refer to the newest master. I don't think there are any effective fixes right now applied to the master branch that can prevent this bug being reproduced yet. Unfortunately since it is a docker image it is hard to pinpoint which commit I was using when reporting this problem.

devonhk:
Gotcha, thanks for the clarification. I forgot about the `v2` docker tag since I pinned mine to a specific `sha256`.

PR #1496, opened 2025-08-21
The one about refs

### LINKED ISSUE #707: Enforce page naming rules
The [page on page naming](https://silverbullet.md/Page%20Name%20Rules) describes several rules, but when creating the page via the UI, those rules aren't enforced. It wouldn't be a bad idea to have checks in the UI that prevent creating pages with invalid names that could cause issues down the line. 

Could also potentially leverage aliases or displayName frontmatter to let it look like what you named it, but still be named properly under the hood? I'm not familiar enough with those features to comment on whether this'd work or not (or if it's a good idea)

https://github.com/silverbulletmd/silverbullet/assets/7729878/a06d61ce-1942-40ec-bbaa-7b292a1616bb

Places that need to check:
- Page Picker
- [[wikilinks]]
- ...elsewhere?

### LINKED ISSUE #860: In [link](url), clicking the end of the url will take to the URL.
For example, `[GitHub](https://github.com<click here>)` take you to the url.

This is a bit annoying when I have to place my mouse cursor at the end of the URL to edit it. It also prevents me from dragging the link from the end.

### LINKED ISSUE #935: Unable to link to attachment within template - page name error
While it is possible to link directly to an attachment via markdown, doing so within a template yields an error of "page name can not end with a file extension".  Noticed this when working with the now-indexed attachments.

In the example markdown below, the first link goes right to the image as expected while the second fails with error mentioned above.

``` 

    [[IMG_2534.jpeg]]

    ```template
    [[IMG_2534.jpeg]]
    ```

```

### LINKED ISSUE #1038: Links to pages with Square Brackets in filename do not work
Links to pages like `meetings/[ACME] My Meeting` do not work. Same for headings. The `[]` seems to break rendering links in general

image

### LINKED ISSUE #1060: Support standard Markdown heading anchors and links
Searching around, it seems you already create anchor links for headings and they can be linked to via wiki style links. Can you please add support for standard Markdown links as well?

some/page.md
```
[[some/page#Sub Heading]]        <-- this works

[Here](#sub-heading)             <-- this does not

# Heading

## Sub Heading
```

### LINKED ISSUE #1141: bug: percent sign in a page title breaks live queries
Steps to reproduce:

1. Start a new silverbullet space: `mkdir repro && docker run -it -p 3000:3000 -v ./repro:/space zefhemel/silverbullet`
2. Press cmd+k and create a page named `test %`. I entered “hello” as content, not sure if having content is necessary.
3. Press cmd+k and go back to the index
4. Run the Libraries: Update command.
5. Paste the following live query into the page:
```
` ``query
page
order by lastModified desc
select name 
render [[Library/Core/Query/Page]] 
` ``
```

Observe how the live query is rendered as an empty white space and in the developer console, I see:

```
URIError: URI malformed
    at decodeURI (<anonymous>)
    at Object.translateUrls (markdown_widget.ts:108:13)
    at markdown_render.ts:576:35
    at Ak (markdown_render.ts:547:3)
    at Ak (markdown_render.ts:553:7)
    at Ak (markdown_render.ts:553:7)
    at Ak (markdown_render.ts:553:7)
    at Ak (markdown_render.ts:553:7)
    at ld (markdown_render.ts:566:5)
    at t.renderContent (markdown_widget.ts:100:18)
```

Once I replace the % sign in the page title with an “a” and go back to the index, the live query starts working.

### LINKED ISSUE #1155: loosen page name validation to allow page names to end in dates
The validatePageName function prevents me from ending my page names with a date.

For example, I placed a file called “notes 14.11.2024.md” in my space.

I can link this page no problem:

Screenshot 2024-11-14 at 21 22 18

But actually opening the page results in a 404:

Screenshot 2024-11-14 at 21 22 23

Using cmd+k, I can see the page, but opening it by pressing Enter results in an error toast:

Screenshot 2024-11-14 at 21 22 30

It would be good to make the name validation more consistent, but more importantly, I would suggest to loosen or entirely remove name validation.

I can’t find an explanation of what the validation is good for. What’s wrong with using page names that end in file extensions? 

And even if validation must be kept for some reason, I think the special case of using a date as part of a page name should be considered reasonable enough to be added as a special case.

### LINKED ISSUE #1161: Media with early colons is counted as URL protocol
# Bug description
I am using [johnny.decimal](https://johnnydecimal.com) to name pages, except I foolishly decided that colons look better than blank spaces so I end up with page names like `S01: Skyler Grey/40-49: Travel/41: Leisure/S01.41+2024-10: USA trip`

Fine, great, pages work fine... except when uploading images. If I drag and drop an image into the page, it doesn't display at all, like this:

[Enregistrement d'écran_20241120_115658.webm](https://github.com/user-attachments/assets/5c39d071-b968-489b-be78-663bbe797878)
(as demonstrated you can refresh the page/etc. with no change)

If you remove the exclamation mark and try to click on the URL, you instead get a dialog asking you if you want to let your silverbullet instance open the page with a handler for the "protocol" before the colon ('scuse my French, this says "Allow https://silverbullet.starrysky.fyi to open this s01 URL with system handler")
image

As a workaround, you can add `./` to the front of the URL, although this does make the URL display as if it doesn't exist:
[Enregistrement d'écran_20241120_120402.webm](https://github.com/user-attachments/assets/44a87f4f-94af-4be0-b209-590a55ec9e49)

You can also get away with this if you have a slash in the URL before the colon. For example, if the space were called `Skyler Grey/S01: Personal` instead this doesn't occur

# Reproduction steps
- Call a page `S01: Foo/Bar`. This has to be at the top of your workspace (you can't have `Baz/S01: Foo/Bar`)
- Drag and drop an image onto the page

# Actual outcome
- The image isn't displayed

# Expected outcome
- The image should be displayed even if the first part of the page name contains a colon

# Versions
Silverbullet Version: Occurs on both 0.9.4 and 0.10.1
Browser Version: Occurs on both Firefox 132.0.1 and Chromium 130.0.6723.116
Sync or online mode: Occurs on both

### LINKED ISSUE #1169: Clicking links in tables
I've noticed that when I click a link to another file that is in markdown table, the entire new page is reloaded and i can see the base theme of silverbullet for a second until the entire page is rendered. This only happens when the link is in a table and i go to that file. If i take the link to the file out of the table and click on it, this does not happen and it loads smoothly and retains my theme the entire time.

### LINKED ISSUE #1214: Page picker space completion does not complete properly in two cases
Hey, I open Page Picker with Ctrl-k and hit space to complete the name of the current page (my main use case is that I want to easily be able to create pages nested under the current page).

When the current page name contains no spaces, everything works as expected.

Image

When the page name contains a space, only the first word is completed. In this case, I was expecting to get `test page/` when I hit space.

Image

When I have pages nested even deeper, it only completes to the first page name. Here I would expect to get everything completed, all levels. Hitting space for each level of page name completion would also be ok.

Image
Image

Thanks

### LINKED ISSUE #1414: lost a note's content 😢
I don't have quite precise steps to reproduce, but I've spent more than an hour on a note called `mastodon-comments.md`, then went back to my home page (`index`), added a link to `[[the/dir/that/had/mastodon-comments.md]]`, clicked it, and it replaced the content of `mastodon-comments.md` with the content of the home page. I realize that I shouldn't have called it `mastodon-comments.md` but `mastodon-comments`, but I think the content is lost forever now, unless there's some hidden version control in Silverbullet that I would've missed? :(

### LINKED ISSUE #1416: Page name cannot have all characters
Pages are saved as files and filesystems have restrictions on what characters are valid.

As a result if you try to create a page named e.g. `&...><`, you'll get:
```
Writing file &...><.md
Error while writing file &...><.md Error: The filename, directory name, or volume label syntax is incorrect. (os error 123): open 'C:\Users\kjk\src\silverbullet\data\&...><.md'
    at Object.open (ext:deno_fs/30_fs.js:541:21)
    at DiskSpacePrimitives.writeFile (file:///C:/Users/kjk/src/silverbullet/web/spaces/disk_space_primitives.ts:83:31)
    at eventLoopTick (ext:core/01_core.js:178:7)
    at async file:///C:/Users/kjk/src/silverbullet/server/http_server.ts:609:24
    at async dispatch (https://jsr.io/@hono/hono/4.7.4/src/compose.ts:51:17)
    at async dispatch (https://jsr.io/@hono/hono/4.7.4/src/compose.ts:51:17)
    at async dispatch (https://jsr.io/@hono/hono/4.7.4/src/compose.ts:51:17)
    at async https://jsr.io/@hono/hono/4.7.4/src/hono-base.ts:434:25
    at async mapped (ext:deno_http/00_serve.ts:399:18) {
  code: "ENOENT"
```

I ran into this problem myself and solved it with encoding filenames.
See `toFileName()` (https://github.com/kjk/edna/blob/main/src/filenamify.js#L69) and `fromFileName()` (https://github.com/kjk/edna/blob/main/src/filenamify.js#L93) 

Feel free to use this code, I hereby declare it public domain / MIT / whatever.

### LINKED ISSUE #1453: Page names with a "file-extension" can't be browsed to
# Description

page names with a "filename extension" in them, don't work.

It seems to me, that trying to browse to them, always either redirects you to the page of the origin-link
or the index page if that is not known.

## How to reproduce

1. add this markdown
```markdown

- new aspiring-page does not work [[newpage.md]]
- other new aspiring-page also does not work [[newpage.xx]]
- works fine [[newpage]]

```
2. hover over the links, check their URL
3. try clicking on the links, see where they take you
4. create a non-empty  file `fromfilesystem.md.md` in silverbullet's space on the filesystem
5. watch it appear in your lua-script list of recent pages
6. try to navigate to it.

## Considerations on proper handling

- At the very least, Silverbullet should not silently fail but mark the page as unreachable or the page-name as erroneous
- It should be noted, that such files can be created via the filesystem, containing content that is then unreachable via Silverbullet.
- Generally, a filename like `newpage.md.md` or `newpage.xx.md` is not a problem for filesystems on any operating system. So it should be ok to just create these pages.
- It surely could be confusing for the user though. Especially if they want to create page `abc` and wrote `abc.md` by habbit. While pages can be renamed later, avoiding mistakes initially is surely the better strategy.

In conclusion I feel the best way to handle this, is to just accept these page names, but show them in a warning colour or with a warning text if they are still only aspiring-pages.

### LINKED ISSUE #1490: remote image embed does not work with query parameters
### Description

This shows up as the actual image

```markdown
![](https://community.silverbullet.md/uploads/default/original/1X/33cc440d7d9f8d54e1c4fc2fe05f2592aa138083.jpeg)
```

but this is not rendered at all

```markdown
![](https://community.silverbullet.md/uploads/default/original/1X/33cc440d7d9f8d54e1c4fc2fe05f2592aa138083.jpeg?foo=bar)
```

The external server is fine with either and delivers the same file, but silverbullet only shows the one without `?foo=bar` as an embed.

### Expected

Both should render the same (and maybe embeds that don't work should show some placeholder?)

I have quite a few files with such links that I want to move over to silverbullet. In many cases the query parameters could just be dropped, but in some they are actually needed to get the correct response.

Maybe similar to https://github.com/silverbulletmd/silverbullet/issues/1285?

MrMugame:
Ok, this is a big one, I hope it's not too hard to review

# The one about refs
## Motivation
The first thing that sparked me to do this is a bunch of issues about unclear naming rules, weird behavior when linking to "odd" names, etc. An example issue would be #707. Kind of the issue this is all about.
The main thing though was the experience programming, a lot of times I just didn't really know what the correct behavior should be. What should happen when a user navigates there? Is that a correct path? Shit breaks? probably not a me issue. What's a path? What's a ref? What's a name? What the fuck is this `editor.navigate({ kind = "page", page = "foo" })` (Shhh... I know I created this, it's awful). 

## Prerequisites
- One should be able to link to every page in their space with every possible link format. Namely wikilinks and markdown links.
- A user should in no way able to create a page with an "invalid" page name.
- It should be clearly defined what an "invalid" page name is.
- The "code" which determines validity should be as small as possible.
- Be as backwards-compatible as possible

## What's new
Let's define some terms so we are talking about the same stuff.

### Names
Not going to be to rigorous here, but every page and every document has a name. This name has to follow some rules.

- Names cannot be empty
- Names cannot start with a `.` nor `^`
- Names cannot contain the following characters `|@#`
- Names cannot contain the sequences `]]` or `[[`
- (Names are case-sensitive)
- (`/` is allowed, contrary to what most filesystems allow)

### `Path`/path
A path is pretty close to what we talked about so far when using that term. It's a relative or absolute file path (*duh*) to a file (page or document) inside your space. This entails that a path always contains a `.`. A path is very closely related to a name, if a name ends in an extension, i.e. a dot followed by one or more letters or digits[^3], which is not `.md`, it's the path. If the name doesn't end in an extension it has to be added for it to be a path. SB will assume it's a markdown file and add `.md`. [^1].
In code the path will look like this.
```ts
export type Path = `${string}.${string}` | "";
```
Obviously a path is also constrained by the naming requirements mentioned above. So this type isn't as restrictive as it should be, but it is restrictive enough for programmers to not pass *any* string (which in the worst case is user generated) into a function which expects a `Path`. Notably a `Path` can also be the empty string. This behavior of this case differs. Either it points to the current page (for wikilinks) or it points to the index page (for `editor.navigate`). 

[^1]: This means one could create a markdown page ending in “extension” by e.g. creating the page `foo.png.md`. This should be discouraged as you won’t be able to link to it easily without typing out the full name.
[^3]: Only the last dot is considered, e.g. the extension of `foo.tar.gz` is `gz`

### String refs
String refs or links are basically what every user knows - it's what you type into a wikilink `[[<stringRef>]]` or a markdown link `[alias](<stringRef>)`. A string ref also has to follow some specific formatting.

- A string ref starting with `^` links to a meta page
- The core (After a possible `^` and before a `@` or `#`) of a string ref is a name
- A string ref can end three ways:
  - `#` followed by a string (which can notably also contain `#`). This will point to the header equal to that string. Notably you can't link to every possible header, because stuff like `]]` will restrict you in a wikilink for example.
  - `@` followed by an upper or lowercase `L` followed by a number, optionally followed by an upper or lowercase `C` followed by a number (e.g. `@l12c13`). This will point to the corresponding line and column
  - `@` followed by a number. This will point to the corresponding character.

A major change introduced here is that you now can't have multiple "endings", this was possible before, but totally useless and introduced weird edge-cases. Notably The first character to appear will be prioritized. E.g `foo@bar#header` is invalid, as the `@` is neither followed by an `l` nor a digit, but `foo#header@bar` is valid. This hasn't changed, but should be explicitly stated: Documents can also be linked to with a position or header[^2]. One more thing is that the **`ref` property on Objects (i.e. indexed stuff), has to be set to a valid string ref.**

[^2]: This isn’t implemented right now, but in the future we could for example jump to specific headers in a pdf or similar

### `Ref`
A ref is the representation of a string ref in code. It looks likes this.
```ts
export type Ref = {
  path: Path;
  meta?: boolean;

  details?:
    | { type: "position"; pos: number }
    | { type: "linecolumn"; line: number; column: number }
    | { type: "header"; header: string };
};
```
It now solves the problem of being able to point to invalid paths, because the `kind` didn't align with what was in the `page` property. It obviously now also can't represent multiples `details`. To turn a string ref into a ref, you can use the `parseToRef` function, which uses the following regex to parse it. This is *the* core regex and encodes all requirements I just mentioned.
```regex
^(?<meta>\^)?(?<path>(?!\.|\^)(?!.*(?:\]\]|\[\[))[^@#\|]*)(@(?<pos>\d+)|@[Ll](?<line>\d+)(?:[Cc](?<col>\d+))?|#\s*(?<header>.*))?$
```

This is now enforced throughout the code (not the regex, nobody should touch that). I won't mention every case here, but it mostly results in flash notifications for the user when he tries to navigate somewhere "illegal". Visually wiki links also show it with a different color (Open to suggestions on the color choice).
image
Some more detailed stuff is written in the changelog. I hope I catched everything.

## Changelog
- Links can now be invalid. On pages they will be highlighted in purple/pink and you won’t be able to navigate to these pages.
- `cleanPageRef` was removed. You now won’t be able to wrap your `SB_INDEX_PAGE` in brackets, like this `[[index]]`. <mark>breaking</mark>
- `editor.navigate` now takes in an object of the new `Ref` type or a string ref. It also still supports the old way `editor.navigate({ kind = "page", page = "yourpage"})`, but using this is heavily discouraged and the user will be warned in the console. For nearly all cases the above example can now be written like this `editor.naviagte("yourpage")`. If you want *exact* compatibility the code has to be rewritten like this `editor.navigate("yourpage" + ".md")` (<mark>breaking</mark>)
- `isLikelyHandled` isn’t called anymore for Wikilinks, So if a plug PlugSpacePrimitive has a valid path the link won’t be highlighted. <mark>breaking</mark>
- Wikilinks will now show the `pageDecoration.prefix`. They do this as proposed in #1477. This means the prefix won’t render if a position is added to the link
- `Navigate to Page` command will now allow to navigate to header or pos for example. It’s hidden from the user so it doesn’t really matter.
- Removed `sb-wiki-link` because [it can’t ever render (it will always have child classes)](https://codemirror.net/docs/ref/#language.HighlightStyle). Renamed the css class `sb-wiki-link-page` to `sb-wiki-link` <mark>breaking</mark>
- `editor.getCurrentPath` doesn’t take an argument now and always returns the full path (i.e. the behavior is now as it would have been with `editor.getCurrentPath(true)`) <mark>breaking</mark>
- Linking to headers now uses the markdown tree to determine the header, not a find and replace. The user will also be navigated to the end of the header <mark>breaking</mark>
- `editor.getCurrentPage/Meta` now “correctly” returns for documents. It didn’t before. <mark>breaking</mark>
- Invalid page links won’t be indexed <mark>breaking</mark>
- The `ref` property of indexed headers will now only contain the position. The `ref` won’t contain the header anymore. <mark>breaking</mark>
- The desired index version was bumped.
- If you navigate to `/`, SB will redirect you to your index page, which will now also be reflected in the URL.
- Removed [this](https://github.com/silverbulletmd/silverbullet/blob/3268b7923ea5c54b8e9323a648d82381f7e20c60/web/reducer.ts#L58-L63). Probably unnecessary anyways
- The index page has to conform to the new page rules, otherwise SB won’t start <mark>breaking</mark>
- Rewritten the code for inlined content. It now also uses `LuaWidget` under the hood. This means that inlined images will now have a border and rounded corners. It will try to show errors if it fails to inline content for some reason. It’s now also possible to transclude markdown from external sources.

## Issues
- closes #707 THE ISSUE
- closes #1414
- fixes #1490
- fixes #1453
- won't fix #1416 (Probably a won’t fix now, as we don’t care about OS restrictions)
- fixes #1038 (I fixed this in another PR already, at least partially I think)
- partially fixes #1214 (Partially both cases were desired behavior as it seems, only fixed the first one, the second one seems correct)
- closes #1155 Not specifically what the creator of the issue envisioned, but a good compromise
- fixes #1169 (In the future tables should maybe also use `LuaWidget`)
- fixes #1060

Already fixed issues, I stumbled upon, while collecting issues.
- resolved #860 Not me
- resolved #1141 This was already fixed not really related to this PR
- resolved #935 Also already fixed
- resolved #1161 (Also fixed for a longer time now)

### Inlined content
As mentioned in the changelog I've rewritten a lot of the inlined content code. They now use `LuaWidget`. Two questions ( @zefhemel )
- We probably want #1404 back. Should we allow that for external sources? It would basically be the hash of the URL.
- Since they are now lua widgets, should we fire an event for unknown mime types and request widget content? This way you could write plugs or lua to transclude documents. (We should probably not do mime types as our library for example won’t return for `excalidraw`).

### Incorrectly named pages
If a page already exists on the file system and is incorrectly named, the user can see it in the picker, can delete or read it using syscalls, but can't navigate to it. This behavior kind of arose from the implementation, but I'm unsure how useful it is. We should heavily discourage users having invalidly named pages, but especially looking at users migrating from e.g. obsidian or logseq, we should at least allow a rename to a proper path. This would be possible using syscalls, there just isn't a user accessible way. Maybe we should add a command `Rename Page` - Or another possibility, when the user tries to navigate to an invalid page we ask if he want's to rename or delete it, similar to the menus you  get when "navigating" to documents which don't have an editor (Actually thinking about it now this is probably the best solution). @zefhemel

### Filesystem restrictions
I.e. case-inensitivity, specific disallowed characters, etc. I thought about this and my conclusion is: don't give a sh*t. This is also what we already do except for [a few exceptions](https://github.com/silverbulletmd/silverbullet/blob/b7e8d4219526e305bd4b74da01a266543858f791/plugs/index/refactor.ts#L235-L240) (We don't even consider different path separators on DOS). SB is designed for sane file systems like `ext4` and everything will work fine on there. No guarantees for the rest.

### Some points I collected while scouring the code
- Drop `PlugSpacePrimitive`. I didn’t know this existed and I’m very sure nobody else knows either and I don't see it used anywhere.
- Consider renaming `editor.getCurrentPage` to `editor.getCurrentName` (also `getCurrentMeta`). Not really important would just align better.
- The regex replace for `pageDecoration.cssClasses` shouldn’t be copy pasted.
- `task.ts` is an absolute mess.
- Creating documents through links?
- Having an empty file open (which thus isn’t yet saved) causes constant errors in the console. This seems dumb. The `FallbackSpacePrimitive` just shouldn’t log the errors, no reason to.
- The empty path i.e. `""`, currently has two meanings. It can ever point to the index page or it can point to the current page, depending on the context. Maybe it would make sense to differentiate these cases. Unsure ...
- Consider dropping internal markdown links, so WikiLinks are internal links and markdown links are www links.
- `pageDecoration.renderWidgets` is not documented and also doesn’t work for LuaWidgets


zefhemel:
Wow! I'm going to give this a close look later today. Thanks for this!

zefhemel:
First of all: this is probably the most elaborate and most beautiful PR I have seen in a long time. Thanks the the amount of time, investment and thinking you put into this. 

Overall I agree with the overall change and think it's a huge improvement. Let me just comment on a few details:

> If you navigate to /, SB will redirect you to your index page, which will now also be reflected in the URL.

I'm not a big fan of this. Is there anyway to avoid having to do so, how fundamental is it? I know there were a few specific code paths to make this happen, but I think URL visuals are worth something. Not a deal breaker, but something that I'd like to keep if at all possible.

> We probably want https://github.com/silverbulletmd/silverbullet/issues/1404 back. Should we allow that for external sources? It would basically be the hash of the URL.

It would be very nice to have that back, I didn't jump on it because I was wondering if we can perhaps somehow generalize it. For refs with headers we can come up with reasonable semantics of what it means to transclude it (all content up to the next header at that same level?), but what if people add e.g. `@1234` what would that do, transclude everything _from_ that position to the end? Perhaps it should just be limited to support headers.

> Since they are now lua widgets, should we fire an event for unknown mime types and request widget content? This way you could write plugs or lua to transclude documents. (We should probably not do mime types as our library for example won’t return for excalidraw).

Honestly I don't have a strong opinion on this. I still use SB 99% purely with text content. Whatever you think is best. It would of course be nice to make this extensible.

> Drop PlugSpacePrimitive. I didn’t know this existed and I’m very sure nobody else knows either and I don't see it used anywhere.

First reaction: but it is used a lot! Then I did a code search, and indeed it isn't anymore. It was there to implement full text search pages, but that's done different now also. I still like the idea and the power of this, but indeed something worth removing. Maybe not yet, though.

> Consider renaming editor.getCurrentPage to editor.getCurrentName (also getCurrentMeta). Not really important would just align better.

I'd say, let's do it but in a backwards compatible way. Add a new syscall with the new name, point all known references to that, but keep the old one around with a comment about deprecation.

plug-api/lib/ref.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  assertEquals(parseToRef(" .foo"), { path: " .foo" });
+  assertEquals(parseToRef("foo."), { path: "foo..md" });
+  assertEquals(parseToRef("/.md"), { path: "/.md" });
+  assertEquals(parseToRef("/foo/.bar.md"), { path: "/foo/.bar.md" });
+  assertEquals(parseToRef("foo[bar"), { path: "foo[bar.md" });
+  assertEquals(parseToRef("foo]bar"), { path: "foo]bar.md" });
+  assertEquals(parseToRef("foo(bar"), { path: "foo(bar.md" });
+  assertEquals(parseToRef("foo)bar"), { path: "foo)bar.md" });
+
+  assertEquals(parseToRef("^.foo"), null);
```
zefhemel, line 29: Now I remember your comment about exceptions being the worst idea ever ;)

plug-api/lib/ref.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  encodeRef,
+  isValidName,
+  isValidPath,
+  parseToRef,
+} from "./ref.ts";
+import { assert, assertEquals } from "@std/assert";
+
+Deno.test("parseToRef() default cases", () => {
+  assertEquals(parseToRef("foo"), { path: "foo.md" });
+  assertEquals(parseToRef("/foo"), { path: "/foo.md" });
```
zefhemel: Thought as I'm reading this (may get back to it later): what does starting with a `/` mean? Is it the same as just leaving it out entirely, all refs are absolute, no?

plug-api/lib/ref.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    details: { type: "linecolumn", line: 42, column: 69 },
+  });
+  assertEquals(parseToRef("foo@L42"), {
+    path: "foo.md",
+    details: { type: "linecolumn", line: 42, column: 1 },
+  });
+
+  assertEquals(parseToRef("foo@ 123"), null);
+  assertEquals(parseToRef("foo@c69"), null);
+  assertEquals(parseToRef("foo@123#header"), null);
```
zefhemel, line 80: The drawback of using `null` values here is that there's no opportunity what the actual problem is (I know what it is in this case, but how will the user?)

plug-api/lib/ref.ts
```diff
@@ -0,0 +1,273 @@
...
+  renderToText,
+} from "@silverbulletmd/silverbullet/lib/tree";
+
+/**
+ * Represents a path with an extension. This is a little cursed, but enforces
+ * that people check the path before setting it. For navigation logic the empty
+ * path will point to the index page. This could differ for e.g. for the
+ * wikilink logic where it points to the currentPage
+ */
+export type Path = `${string}.${string}` | "";
```
zefhemel, line 14: I like it, it also gives us a way to more clearly communicate when we pass a path around and not random strings.

plug-api/lib/ref.ts
```diff
@@ -0,0 +1,273 @@
...
+ */
+export type Ref = {
+  path: Path;
+  meta?: boolean;
+
+  details?:
+    | { type: "position"; pos: number }
+    | { type: "linecolumn"; line: number; column: number }
+    | { type: "header"; header: string };
+};
```
zefhemel, line 27: Even though we probably don't want to create a `Name` type, would it make sense to put your very clear definition of what a name is (as well as the other spec-y stuff you put in the PR) into this source file as comments? Perhaps also in the website content somewhere appropriate.

plug-api/lib/resolve.test.ts
```diff
@@ -1,54 +1,40 @@
...
+  assertEquals(
+    "bar@123#456",
+    resolveMarkdownLink("foo", "/bar@123#456"),
+  );
+  assertEquals(
+    "baz.jpg",
+    resolveMarkdownLink("foo/bar", "/baz.jpg"),
+  );
 
   // Relative paths
```
zefhemel, line 23: It would be nice to have some test cases with parent paths, like `../something` which apparently SB supports (couldn't even remember) but I don't see tested here.

plugs/index/page_links.ts
```diff
@@ -148,70 +160,6 @@ export async function indexLinks({ name, tree }: IndexTreeEvent) {
       return true;
     }
 
-    // Also index links used inside query and template fenced code blocks
```
zefhemel, line 151: Was this an explicit decision to no longer support?

web/editor_ui.tsx
```diff
@@ -156,26 +156,37 @@ export class MainUI {
...
-                  document.name === name
-                );
+                const ref = parseToRef(name);
+
+                // Check beforhand, because we don't want to allow any link
+                // stuff like #header here. The `!ref` check is just for
+                // Typescript
+                if (!isValidName(name) || !ref) {
+                  // Realistically this should only happen when creating a new page
+                  client.flashNotification(
```
zefhemel: Should be an `error` (second arg)

zefhemel:
Moving on to testing:

I'm not clear on what the current stance on relative links is. Do we support those? If so, how?

I cannot use a link like `../hello` because this is invalid now due to the "can't start with ." restriction. I've never attempted a relative page link, but they somehow work so this is a consideration.

If I link: `[[hello/../index]]` it appears in the editor in red (doesn't exist), but when I click it, it navigates to the index page (but `hello/../index` in its name). That's not great. Perhaps paths like that need to be normalized, or any attempt at such relative links should just not be there?

Once upon a time I invested in doing stuff around path resolving in order to support federated links (where under those links I assume I'd have to rewrite links), but that's no longer a thing. You dug deeper into this, do we need to resolve paths and support relative links in any way?

Bug: when I drag & drop an image (.png) on this branch it seems to suggest a `Inbox/2025-03-16/file.txt` path. In the main branch it correctly pulls the file (.png). Something regressed here?

Next issue: auto complete from transclusions now seems broken, when I do `![[typytype]]` the completion box instantly disappears.

plugs/index/page_links.ts
MrMugame: The indexing was only done for `template` and `query` blocks. I don't think that's needed anymore. Haven't really thought about doing this for every fenced codeblock, would there be a usecase?

MrMugame:
> Bug: when I drag & drop an image (.png) on this branch it seems to suggest a Inbox/2025-03-16/file.txt path. In the main branch it correctly pulls the file (.png). Something regressed here?

Is your file name by chance an invalid name? It fallbacks to `file.txt` in that case, I improved it a little bit now, so it tries to honor the extension.

> Next issue: auto complete from transclusions now seems broken, when I do ![[typytype]] the completion box instantly disappears.

Urggghh, good catch, it seems to happen, because LuaWidget fires the no-op transaction to fix the cursor layer, which intern cancels the autocomplete. Maybe it wasn't the best idea to show error messages - or I'll ask the codemirror wizard what's the best solution here is.

MrMugame:
> I cannot use a link like ../hello because this is invalid now due to the "can't start with ." restriction. I've never attempted a relative page link, but they somehow work so this is a consideration.

I think it should work for markdown links, the resolving is done before the url is parsed. Arguably now saying that whatever is in a markdown link is a *string ref* is incorrect. (Also the resolving only works for leading `..`, not `..` inbetween)

> If I link: [[hello/../index]] it appears in the editor in red (doesn't exist), but when I click it, it navigates to the index page (but hello/../index in its name). That's not great. Perhaps paths like that need to be normalized, or any attempt at such relative links should just not be there?

Good point, ~~the "resolving" here seems to happen at the file system level.~~ News flash: URLs support these "dot segments", so the problem is created at the `fetch`. ~~I.e. you can't really create a folder named `..` in posix.~~ This is also seems to be a problem in the current SB version. There is a very similar situation for the `/foo` case, which I thought I handled, but I think i only checked it while SB was offline and with the datastore it's fine: `/foo` and `foo` are different cases. But on the filesystem it causes problem. This now leads me to think we should just handle both of these cases in the [`normalizePath`](https://github.com/MrMugame/silverbullet/blob/315ab8711a7ada9815470f170f1ceed62ab0f97e/plug-api/lib/ref.ts#L59) function. This obviously obscures the actual ref away from everyone using `parseToRef`, but I *think* that shouldn't matter. Or we just straight up disallow string refs starting with `/` or containing `/../`. I'd like to here your opinion on that. Imo too many restrictions are a bad thing, but I also wouldn't know why anybody would use `/../` in their links (especially because it's already buggy), really unsure.

(The `/./` case should be handled the same way)
(Another case that comes to mind here is `/foo/.bar/page`, but this shouldn't cause problems as we only hide top level hidden folders)

MrMugame:
> Once upon a time I invested in doing stuff around path resolving in order to support federated links (where under those links I assume I'd have to rewrite links), but that's no longer a thing. You dug deeper into this, do we need to resolve paths and support relative links in any way?

Unsure about the federated stuff. But for relative paths see the last comment, the only place I found them supported is markdown links.

plug-api/lib/ref.ts
MrMugame: Writing docs is on my todo list, for the webstie there already is a `Page Name Rules` file

mjf:
@MrMugame All of this is very cool! We all owe you our big "THANKS for your great effort and big deal of work!" Really.

-----

> Since they are now lua widgets, should we fire an event for unknown mime types and request widget content? This way you could write plugs or lua to transclude documents.
> :+1: 

The more ways to hook into something with Space Lua the better! :)

> I cannot use a link like `../hello` because this is invalid now due to the "can't start with ." restriction. I've never attempted a relative page link, but they somehow work so this is a consideration.
> :+1:

I would suggest not allow the leading dot for *names* only and if it really must be allow relative paths - `./` and `../` (with obvious common semantic). I would not resist if even `.name` was allowed, why not? Consider this page: `/My/Configs/.bashrc` (file `.bashrc.md` in fact) instead of page: `/My/Configs/dot.bashrc` (uff) or something like that - that's a pity. ;)  (I presume the motivation for this had been to exclude "hidden" files at some point in time in the history of the project but why? Let's support some `.sbignore` and you are all well.)

-----

> Only the last dot is considered, e.g. the extension of `foo.tar.gz` is `gz`.
> :+1:

Excellent! Finally! :)

To note, the very same principle could "restrict" the page naming restrictions even more! You are going to exclude the `@` character (because it collides with something aparently) but the same principle can hold for it as for the extension, and the same may apply for the `#` char, and perhaps even the `|` char (I've seen URLs with pipes inside!) There is, in fact, no need to exclude these from the path/name, all what's needed is to setup the right semantics and describe and demostrate that to the user saying if you want a page `/Foo@Bar` you must type `Foo@Bar@` (in which case the special last `@` part is empty) because the last `@...` express `<something>` else, or, `[[/Foo?|Bar|Baz]]` will be rendered as `Baz` linking to `/Foo?|Bar` page. Or just allow user to `\\` escape things. :)

plug-api/lib/ref.test.ts
MrMugame: Yeah, but using a regex doesn't really allow for much more. If we wanted to return proper error messages, we'd have to write a handwritten parser, which would probably also be magnitudes slower

MrMugame:
> I would not resist if even .name was allowed, why not?

Mhm, good point, I didn't really consider this. I guess the main problem here is that there still is the JWT database and probably a bigger issue a possible `.git` that shouldn't be synced to the client (although interestingly enough there are wasm implementations of git, maybe syncing that could be useful to people :eyes:). So, if we allow names starting with a `.` something like `.sbignore` is needed, which kind of already exists  as `SB_SPACE_IGNORE`. Problem with that env variable right now being that it confused a lot of people. If you e.g. ignore `foo.md` it's not visible to you anywhere, but you can still read and write to `foo.md` and now you get weird edge cases when you try to create `foo.md` on the client (especially when you are offline, you may start to write into a file and once it syncs, you get problems).

So what I would propose is we introduce a `.sbignore` as a file and scrap the old env variable. If the file doesn't exist inside a space, SB will for backwards compatibility create a version, which just contains `.*` to disallow files starting with a dot. SB syncs that file with the client as part of the `ClientConfig` (meaning it can't be updated at runtime, so it makes sense to ignore `.sbignore` itself too). SB will both check these naming rules on the server and on the client. So even when you are offline you won't be able to create such a file. Implementation wise, you will be able to create a `Ref` which points to a disallowed file, the check will only happen in `CheckPathSpacePrimitives`, which should for all cases just bubble up to the correct place. Two issues:
- `.sbignore` seems like a bad name, because we are now actually disallowing these names (maybe `.sbexclude`?)
- We will need the gitignore-parser on the client. It doesn't seem to big, but I don't know how "good" that library is. I guess we can always replace it when bugs arise.
If I haven't missed something major, this should work. What does @zefhemel think?

zefhemel:
> Is your file name by change an invalid name? It fallbacks to file.txt in that case, I improved it a little bit now, so it tries to honor the extension.

🤦🏻 yes, it was an invalid name.

plugs/index/page_links.ts
zefhemel: Oh yeah, you're right. I didn't notice the template and query mentions.

zefhemel:
@MrMugame I like the .sbignore-style solution — I hadn't considered that, which is ironic because... well it's based on .gitignore's format.

Yeah `.sbexclude` with `.*` can work. The only thing I'm not a huge fan is to introduce more magically created files, but I suppose we already have `.silverbullet.db.json` also, so maybe it's ok.

web/editor_ui.tsx
MrMugame: Fixed in d0785ac

MrMugame:
> Yeah `.sbexclude` with `.*` can work. The only thing I'm not a huge fan is to introduce more magically created files, but I suppose we already have `.silverbullet.db.json` also, so maybe it's ok.

I guess using we could also keep the env variable:

Pro:
- Everything so far is configured using env variables
- (It wouldn't create files on it's own)

Con:
- Not very ergonomic
- When moving the space, you loose the ignore config

Or we use the file option and we just internally default to `.*` and not create the file. This seems opaque tho.

MrMugame:
~~AHAHHA, If we allow names starting with `.` we will consider the rest to be the extension, so we basically have an "empty name", this goes against the "no empty names" rule and also seems dumb.~~ Wait, this case is already handled, when normalizing the path. If it doesn't have any characters before the `.` it adds a `md`

EDIT
AHAHAH
We also get problems with url paths, like `.ping` or `.shell`

zefhemel:
Ah yes and there may be more. Let's not allow this.

MrMugame:
> Ah yes and there may be more. Let's not allow this.

Yeah, got a POC working, but I moved it to a different branch, maybe some day.

> I'm not a big fan of this. Is there anyway to avoid having to do so, how fundamental is it? I know there were a few specific code paths to make this happen, but I think URL visuals are worth something. Not a deal breaker, but something that I'd like to keep if at all possible.

Should be fixed with 150306f

MrMugame:
(Side-note, `lib/plugos/syscalls/fetch.ts` seems to be unused, same with `lib/plugos/syscalls/fs.deno.ts` and the corresponding tests)

zefhemel:
Correct, just deleted those 81b413f88c91eca543c41ace4497caa278c799b1

MrMugame:
We get into some troubling cases with `<>` in md links. For md links they can kind of "escape" the link if you e.g. have parens inside your link `[alt](<foo)bar>)`. But you can now construct pages which one isn't able to link to using markdown links like `foo>)bar`, If you don't escape it the `)` is a problem, if you escape it the `>` is a problem. Either we introduce some veeeerry obscure rules or we have to kind of drop this

> One should be able to link to every page in their space with every possible link format. Namely wikilinks and markdown links.

I'm still for dropping internal md links, but I guess there are users who wouldn't be happy with that.

MrMugame:
> To note, the very same principle could "restrict" the page naming restrictions even more! You are going to exclude the @ character (because it collides with something aparently) but the same principle can hold for it as for the extension, and the same may apply for the # char, and perhaps even the | char (I've seen URLs with pipes inside!) There is, in fact, no need to exclude these from the path/name, all what's needed is to setup the right semantics and describe and demostrate that to the user saying if you want a page /Foo@Bar you must type Foo@Bar@ (in which case the special last @ part is empty) because the last @... express <something> else, or, [[/Foo?|Bar|Baz]] will be rendered as Baz linking to /Foo?|Bar page. Or just allow user to \\ escape things. :)

I though about this. You can start writing out the cases that can happen here and you will quickly notice that the list is endless. Here are a few thoughts:
- `[[/Foo?|Bar|Baz]]` Why not `/Foo?` with `Bar|Baz` as the alias, the way it currently is. There is no reason to choose one or the other, it's either pipes in the string ref or pipes in the alias. The way it's in the PR right now is the way it's in e.g. obsidian and the way it has been, so I don't really see a reason to change.
- `Foo@Bar@` This is frankly insane, this introduces a new "ending", for a hacky way to link to pages. Maybe I should specify a little more "Every path should be a valid string ref", this isn't given here. What we could say is that we don't consider endings, which don't exactly fall under `@123` or `@l69c69`. So `foo@bar` is allowed, but `foo@123` will just be `foo`
- If we allow the `#` in names, we will have to "disallow" it in headers. Again the tradeoff. I could see arguments here for "disallowing" it in the header

MrMugame:
I'm going insane, I don't know how I overlooked (Looking in my notes I think I considered that, but somehow it got lost) that case when writing all this, but my current problem is that not every name is a valid string ref, specifically the case of something like `foo.png.md`, where the name is `foo.png` and then interpreting that as a string ref, gives you trouble, because it looks like the document/image `foo.png`. The problem being that this is used quiet a lot throughout the code, all the indexing stuff just does `ref: name`, so in the end we get this ambiguity. The core problem here is that names aren't unique. So `foo.png` has the name `foo.png` and `foo.png.md` has the name `foo.png`
So either we roll with it and say ref != name and change it throughout the code (and probably some extensions and user code, I guess the pattern of `editor.navigate(name)` is used a lot, but I don't know, although it would only break for that edge case) or we decide to make names unique. This would mean that we'd have to disallow the case of `foo.png.md`, i.e. pages can only have one extension.

Also unsure if I like the idea of allowing `/../`, `/./`, etc. in stringrefs. I don't see a need to. It's also currently now supported, it just bugs out. I feel like forbidding that isn't too insane.

@zefhemel 

Ok I think I'm going to do the second thing and disallow the `foo.png.md` case. It is buggy in the current SB anyways. It shouldn't break any backwards compatibility and we can always rewrite the rest of the name -> ref cases in the future and lift the restriction. Also going to disallow the `/../` cases
This means that the confliced pages can't be called `.conflicted.314551285412` anymore

zefhemel:
Fun times!

> Ok I think I'm going to do the second thing and disallow the foo.png.md case. 

Sounds like the best option 👍🏻 

> Also going to disallow the /../ cases

Yeah, let's.

> This means that the confliced pages can't be called .conflicted.314551285412 anymore

Err, ok  you'll change the naming pattern for that then?

MrMugame:
@zefhemel I just noticed, the docs say that the markdown link syntax can only be external? Is that how it should be? Because how it was in code and how I now also do it is that markdown links can be internal and are just relative. Tbh only having wikilinks as internal would certainly make things easier.

MrMugame:
Ok, so I'm pretty much done now, I think. At least my todo list is empty, there are probably still bugs I would guess.

One thing still left open is the problem with transclusions not autocompleting. Not really getting any smarter from searching on that topic. Codemirror and async code don't really play nice.

And also small sidenote, having a file without an extension in your space can cause problems, wouldn't recommend, at least wouldn't try touching it using SB. Maybe we should just ignore those. 

Overall I think this should make SB much more robust *for the user*. Internally SB still needs overhaul. Best case in the future, there will be no code relying on names anymore. Only paths. This would also make it possible to get rid of a lot of doubled code for documents and pages. But this PR is already close to 100 commits and starting to touch stuff like primitives, would probably warrant updating like 100+ more files, but at least the "ground work" is now there to make those changes in the future.

lib/spaces/sync.ts
```diff
@@ -84,8 +84,8 @@ export class SpaceSync extends EventEmitter<SyncEvents> {
 
     let operations = 0;
     const revisionFileName = filePieces.length === 1
-      ? `${name}.conflicted.${pageData2.meta.lastModified}`
-      : `${fileNameBase}.conflicted.${pageData2.meta.lastModified}.${fileNameExt}`;
+      ? `${name}.conflicted:${pageData2.meta.lastModified}`
+      : `${fileNameBase}.conflicted:${pageData2.meta.lastModified}.${fileNameExt}`;
```
zefhemel, line 88: Since the new convention is to use a colon (hope this is OK on Windows machines), I've also updated the query in Maintenance.md

zefhemel:
Made one minor change (Maintenance.md updated query). For the rest looks good. I think this just needs some soaking time, have people use it and report issues. 

Merging. Thanks a lot!

lib/spaces/sync.ts
MrMugame: Good point, maybe something like a hyphen or underscore would be better. But this shouldn't be a too problematic change.

PR #1502, opened 2025-08-25
Add command to toggle clean mode

### LINKED ISSUE #1500: Ability to disable clean mode
I've assumed the name 'clean mode' from the name of [/web/cm_plugins/clean.ts : cleanModePlugins()](https://github.com/silverbulletmd/silverbullet/blob/041b1cdcfe175b56c0d9bc7a7acc10b6344db8a8/web/cm_plugins/clean.ts#L19). I mean the way markdown syntax disappears when your cursor isn't focused on it.

I prefer to see the various markdown marks when I use markdown editors, so I'd like an option to not have to highlight or alt+click to reveal the `*`s and `[]()`s on my pages. Is this a feature y'all'd be willing to have?

(If yes, someone extremely pretty and devastatingly clever has fortunately [already written some code to accomplish this](https://github.com/aphymi/silverbullet/commit/78118c55ac3bed0c5ded698c46c616df9ab94307); if y'all are good with it, I'll make a PR.)

aphymi:
This PR resolves #1500 by adding a built-in editor command to toggle 'clean mode', i.e. the feature that hides markdown syntax like `*`s when not focused by the cursor.

The goal of having clean mode disabled is to have *both* the markdown syntax and formatting simultaneously, except for places where that would be overly repetitive (e.g. tables), in which case the syntax is prioritized.

Implementation-wise, I largely mimicked the `Toggle Dark Mode` command for the toggling behavior, creating a `cleanMode` UI option, defaulting to true. When false, almost every instance of markdown syntax being hidden is disabled, as if everything on the page was alt-clicked or highlighted. The only exception is lua directives (`${...}`); their behavor is unchanged whether clean mode is enabled or not.

There are, I think, two questions left in my mind before I'd call this done:
1. *Should* lua directives be excluded from this? The other reasonable option, I think, is giving them the same behavior as images (and other inline content), rendering both the syntax and the result. I poked at that a bit, but it behaved oddly.
2. Should the color of syntax characters change with clean mode disabled? The current `--meta-color` in the dark theme is a rather angry red, giving the impression that there's some kind of error with syntax characters. It'd be nice if that was less the case with clean mode disabled, where syntax characters are an expected part of the page. This could be changed on a per-user/space basis with space styles, of course.

### DIFF
--- a/web/cm_plugins/wiki_link.ts
+++ b/web/cm_plugins/wiki_link.ts
@@ -19,6 +19,8 @@ import { wikiLinkRegex } from "../markdown_parser/constants.ts";
  * Plugin to hide path prefix when the cursor is not inside.
  */
 export function cleanWikiLinkPlugin(client: Client) {
+  const cleanModeEnabled = client.ui.viewState.uiOptions.cleanMode;
+
   return decoratorStateField((state) => {
     const widgets: any[] = [];
     // let parentRange: [number, number];
@@ -38,6 +40,11 @@ export function cleanWikiLinkPlugin(client: Client) {
 
         const { leadingTrivia, stringRef, alias, trailingTrivia } =
           match.groups;
+
+        const textWithoutMarks = text.slice(
+          leadingTrivia.length,
+          -trailingTrivia.length,
+        );
 
         const ref = parseToRef(stringRef);
 
@@ -116,7 +123,7 @@ export function cleanWikiLinkPlugin(client: Client) {
         widgets.push(
           Decoration.replace({
             widget: new LinkWidget({
-              text: linkText,
+              text: cleanModeEnabled ? linkText : textWithoutMarks,
               title: helpText,
               href: ref ? encodePageURI(encodeRef(ref)) : undefined,
               cssClass: "sb-wiki-link " + css,
@@ -143,7 +150,11 @@ export function cleanWikiLinkPlugin(client: Client) {
                 );
               },
             }),
-          }).range(from, to),
+          }).range(
+            // If clean mode is disabled, don't replace the marks
+            cleanModeEnabled ? from : from + leadingTrivia.length,
+            cleanModeEnabled ? to : to - trailingTrivia.length,
+          ),
         );
       },
     });
--- a/web/editor_ui.tsx
+++ b/web/editor_ui.tsx
@@ -128,6 +128,15 @@ export class MainUI {
     }, [viewState.uiOptions.darkMode]);
 
     useEffect(() => {
+      clientStoreSyscalls(client.ds)["clientStore.get"](
+        {},
+        "cleanMode",
+      ).then((storedCleanModePreference: boolean) => {
+        viewState.uiOptions.cleanMode = storedCleanModePreference;
+      });
+    });
+
+    useEffect(() => {
       // Need to dispatch a resize event so that the top_bar can pick it up
       globalThis.dispatchEvent(new Event("resize"));
     }, [viewState.panels]);
--- a/web/ui_types.ts
+++ b/web/ui_types.ts
@@ -39,6 +39,7 @@ export type AppViewState = {
   uiOptions: {
     vimMode: boolean;
     darkMode?: boolean;
+    cleanMode: boolean;
     forcedROMode: boolean;
     customStyles?: string;
   };
@@ -76,6 +77,7 @@ export const initialViewState: AppViewState = {
   uiOptions: {
     vimMode: false,
     darkMode: undefined,
+    cleanMode: true,
     forcedROMode: false,
   },
   isMobile: false,
--- a/plugs/editor/editor.plug.yaml
+++ b/plugs/editor/editor.plug.yaml
@@ -10,6 +10,10 @@ functions:
     path: "./editor.ts:toggleDarkMode"
     command:
       name: "Editor: Toggle Dark Mode"
+  toggleCleanMode:
+    path: "./editor.ts:toggleCleanMode"
+    command:
+      name: "Editor: Toggle Clean Mode"
   newWindow:
     path: editor.ts:newWindowCommand
     command:
--- a/plugs/editor/editor.ts
+++ b/plugs/editor/editor.ts
@@ -12,6 +12,9 @@ export async function setEditorMode() {
   if (await clientStore.get("darkMode")) {
     await editor.setUiOption("darkMode", true);
   }
+  if (await clientStore.get("cleanMode")) {
+    await editor.setUiOption("cleanMode", true);
+  }
 }
 
 export function openCommandPalette() {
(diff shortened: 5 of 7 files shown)


web/cm_plugins/wiki_link.ts
```diff
@@ -19,6 +19,8 @@ import { wikiLinkRegex } from "../markdown_parser/constants.ts";
  * Plugin to hide path prefix when the cursor is not inside.
  */
 export function cleanWikiLinkPlugin(client: Client) {
+  const cleanModeEnabled = client.ui.viewState.uiOptions.cleanMode;
```
MrMugame: Maybe pass this in as a parameter? Imo, that would be a little cleaner

web/editor_ui.tsx
```diff
@@ -127,6 +127,15 @@ export class MainUI {
       });
     }, [viewState.uiOptions.darkMode]);
 
+    useEffect(() => {
+      clientStoreSyscalls(client.ds)["clientStore.get"](
```
MrMugame: The `darkMode` example above sets a really bad precedent. I don't know what @zefhemel thinks of this, but IMO, this isn't at all how this should be done. (Also the `toggleDarkMode` is bad imo, it gets the option from the ui options and sets it in the clientstore). ~~At least `client.clientSystem.localSyscall` should have been used.~~

The clientstore should be solely used to cache the option in the browser, so it persists across reloads. It should never be touched inside of here and most definitely not by constructing some new syscalls. The code in here should only look at the ui option and update accordingly. The editor plug should set that ui option. (Now this is more of a personal opinion, but I think it should also be `darkMode: "system" | "dark" | "light"`, I think that reflects it better, but that really doesn't matter and I'm unsure if changing that may break some plug code). The `vimMode` option seems to do all of this properly.

This isn't your fault, I just think this should be changed. It may work right now, but that's not really an argument right here.

web/cm_plugins/wiki_link.ts
```diff
@@ -116,7 +123,7 @@ export function cleanWikiLinkPlugin(client: Client) {
         widgets.push(
           Decoration.replace({
             widget: new LinkWidget({
-              text: linkText,
+              text: cleanModeEnabled ? linkText : textWithoutMarks,
```
MrMugame: No reason to do all of this again I think. Just bail out at the `isCursorInRange` check.

web/ui_types.ts
```diff
@@ -76,6 +77,7 @@ export const initialViewState: AppViewState = {
   uiOptions: {
     vimMode: false,
     darkMode: undefined,
+    cleanMode: true,
```
MrMugame: This should be `false` by default

aphymi: To be clear: `true` is the *current* behavior, and setting it to `false` triggers the new behavior introduced by this PR. Are you saying you want this new behavior to be default?

MrMugame: Oh, ok, I interpreted `cleanMode` as in "cleaned from any styling". I guess that's a perspective thing.

PowerUser64:
Maybe it would be good to rename this option. _Clean mode_ might be what it's called someplace internally, but I think we might want to come up with a different name to present to users. I was confused when I saw #1500 at first - I thought "ooh, clean mode - what's that?" After seeing similar confusion in @MrMugame's comment [above](https://github.com/silverbulletmd/silverbullet/pull/1502#discussion_r2298012168), I think it's not just me. Perhaps calling it something like "inline rendering" it would be more immediately obvious. Thoughts? Feel free to share other name ideas too if you think of any.

MrMugame:
My two cents

> Should lua directives be excluded from this? The other reasonable option, I think, is giving them the same behavior as images (and other inline content), rendering both the syntax and the result. I poked at that a bit, but it behaved oddly.

How did it behave? My guess would be some `margin: -1em 0;` problems. It's all very finicky. But overall, that behavior would seem the most intuitive to me.

> Should the color of syntax characters change with clean mode disabled? The current --meta-color in the dark theme is a rather angry red, giving the impression that there's some kind of error with syntax characters. It'd be nice if that was less the case with clean mode disabled, where syntax characters are an expected part of the page. This can already be changed on a per-user/space basis with space styles, of course.

Maybe a data attribute on the html element similar to the theme, so people can style accordingly

web/cm_plugins/wiki_link.ts
aphymi: ...oh! You're right! Thanks for pointing that out!

web/cm_plugins/wiki_link.ts
aphymi: Hmmm, I agree that this placement is awkward, but I don't think passing it as a parameter would be better, given that the client is already being passed, with all its context. I've moved the definition closer to its only remaining use now, anyway.

web/editor_ui.tsx
aphymi: Thank you for the explanation! I understand the flow *much* better with that bit of info. Reorganised it in 3ab7113; how's it look now?

aphymi:
@PowerUser64 
> Perhaps calling it something like "inline rendering" it would be more immediately obvious. Thoughts? Feel free to share other name ideas too if you think of any.

Yeah, "clean mode" isn't necessarily intuitive; I'd be up to change that. 'inline rendering' feels too nonspecific. Maybe 'Toggle Markdown Syntax Rendering'? It's a bit of a mouthful, but it's probably the clearest option?

@MrMugame
> My guess would be some margin: -1em 0; problems. It's all very finicky.

Oh! Yes, that was exactly it. Don't know how I missed that.

> But overall, that behavior would seem the most intuitive to me.

Fair enough! I've added that in b1d856c. The only catch is that the rendered result still hides while you're editing the directive. Lua blocks seem to have a flash of an incorrect render every refresh before they settle down, but that gets really jittery and ugly when a refresh happens with every keystroke, so I chose to make them hide until your cursor leaves the directive.

MrMugame:
> Lua blocks seem to have a flash of an incorrect render every refresh before they settle down, but that gets really jittery and ugly when a refresh happens with every keystroke, so I chose to make them hide until your cursor leaves the directive.

That's something to consider with inlined content (``) in the current/"normal" mode, because it's facing the exact same problem and this also currently causes a bug with autocomplete not working. That's would be a pretty easy fix. Unsure what @zef thinks.

aphymi:
I've renamed the command to `Toggle Markdown Syntax Rendering`, and added a note about it to a couple relevant documentation pages. Absent a fix for that rendering weirdness, I think this feature is ready!

zefhemel:
Sorry for jumping in late. This looks good to me. Fun fact, this is actually largely what SilverBullet looked like by default for the first year or so (video evidence: https://www.youtube.com/watch?v=RYdc3UF9gok).

Question: would you really want to configure this per client, or is more of a personal (as in: space wide) type configuration? For dark mode and vim mode (that works similar) you can really argue that e.g. on mobile or one or time of day you may want different settings, whereas this one to me seems more of a "life choice" style thing. If you agree, perhaps this should just be a configuration (lua.set type of deal) rather than a thing you toggle with a command.

Thoughts?

aphymi:
I *do* expect to set it once and then barely touch it again, but I'd also like to keep it easy to flicker on and off, whether that's someone like me just wanting to check what the markdown output would look like, or someone who keeps syntax rendering on wanting to get a clearer sense of the page's text.

In *theory* we could have a boolean config option and a separate command that lets you choose between `"config" | "on" | "off"` or something of that sort, but that feels too complicated.

zefhemel:
Ok, tested this. Looks ok to me!

PR #1505, opened 2025-08-25
Improve transclusions

MrMugame:
- Move some code around for better organization
- Align the behavior between transclusions from the widget and `renderMarkdownToHTML`
- `expandMarkdown` won't run into endless loops anymore
- Markdown transclusions from external sources are removed (You get CORS errors 90% of the time anyways and I don't want to start proxing that stuff rn)
- We are back to determining the content type from external sources using the extension instead of a HEAD request. For the reasons see the comments in the code.


zefhemel:
Looks good, did some testing and seems to work as advertised. Thanks once more!

PR #1507, opened 2025-08-26
Expose file system API on /.fs instead of root

### LINKED ISSUE #1506: Expose file system API on /.fs instead of root
This was discussed at some point with @MrMugame to clean up the URL namespace, I think now is the time to do this.

The idea is that all files will be served from `/.fs`:

* When visiting page `/index` this will serve the HTML for the UI, and in turn fetch `/.fs/index.md` to get the content
* I'll probably rename `/all.json` to plain `/.fs` for the file listing

This may break some users who used the API directly.

zefhemel:
Fixes #1506


zefhemel:
Ok, this seems to be working. After upgrading you *have* to reload twice to get the up-to-date client other things break, but that's a one-time issue.

zefhemel:
@MrMugame if you have any comments let me know. Perhaps there's also some things we can change regarding page name restrictions as a result of this? You have a better understanding of this now I think :)

server/http_server.ts
```diff
@@ -482,94 +483,101 @@ export class HttpServer {
...
-        // This header signififies the requests comes directly from the http_space_primitives client (not the browser)
-        !req.header("X-Sync-Mode") &&
-        req.header("sec-fetch-mode") !== "cors"
-      ) {
-        // It can happen that during a sync, authentication expires, this may result in a redirect to the login page and then back to this particular file. This particular file may be an .md file, which isn't great to show so we're redirecting to the associated SB UI page.
-        console.warn(
-          "Request was without X-Sync-Mode nor a CORS request, redirecting to page",
+    // This is a good guess that the request comes directly from a user
+    if (
+      req.header("Accept")?.includes("text/html") &&
```
MrMugame: Can we remove that now? I added this when implementing document editors, so if the user navigates to a url he doesn't get the raw file (it's quiet hacky). I think we can also drop the `?raw=true` then (It's only used in the `loadDocumentEditor` function). We can just navigate the user to the `fs` path.

server/http_server.ts
```diff
@@ -482,94 +483,101 @@ export class HttpServer {
...
 
-    this.app.get(filePathRegex, async (c, next) => {
-      const req = c.req;
-      const name = req.param("path")!;
+    if (
+      name.endsWith(mdExt) &&
+      // This header signififies the requests comes directly from the http_space_primitives client (not the browser)
+      !req.header("X-Sync-Mode") &&
+      req.header("sec-fetch-mode") !== "cors"
+    ) {
```
MrMugame: Can this even still happen? Maybe I'm missing something, but I feel like this is a relic from offline mode times or something. (Same for service worker)

I can't reproduce the behavior

zefhemel: In theory this was still doing something, but because this exact same thing is also implemented in the service_worker it never actually makes it to the server. Removing it here.

PR #1517, opened 2025-09-02
Add tag rendering template to Query Templates

costalfy:
Hi,
I created this taskItem because I needed to have a list of all the tags that exist in my index space like this :
```
## Tags
${template.each(query[[
from index.tag "tag"
order by _.name desc
  select {name=name}
]], templates.tagItem)}
```

I hope it can help 

Regards

### DIFF
--- a/Library/Std/Query Templates.md
+++ b/Library/Std/Query Templates.md
@@ -27,6 +27,11 @@ templates.itemItem = template.new([==[
 templates.paragraphItem = template.new([==[
 * [[${ref}]] ${text}
 ]==])
+
+-- Renders a tag object
+templates.tagItem = template.new([==[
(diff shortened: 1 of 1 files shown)


agreenfield1:
I see this template in "Library/Std/Query Templates" and it works in the example at the bottom of that page, but for some reason I cant use it on anywhere else?  If I just start typing ${templates. the immediately previous item can autompopulate (paragraphItem) but not this one.  And any reference to templates.tagItem gives the error about calling nil as a function.

zefhemel:
@agreenfield1 are you on a build that has this integrated? It's not on the `:latest` yet, you'll have to use the `:v2` edge builds to get it. After an upgrade, always reload your client 2x to be sure to have the latest client files.

PR #1522, opened 2025-09-05
Add gcs storage backend

xmclark:
Definitely not working correctly right now.


zefhemel:
Hey! Thanks for putting in this work.

Two risks:

1. In the past I actually had an S3 backend, very similar to this one (it did have to rely on an additional database for keeping some meta data like file creation and modified dates which somehow couldn't reliably or efficiently be pulled from S3). Eventually I removed it because I didn't use it myself, some people ran into issues and I didn't have the time or mental space to maintain and test it a lot. That's why I deleted it. I'm mentioning this, because if we merge your PR, I will not have the time to test or maintain it either. If it causes problems for people and you (or other people) may not be around to support, I may delete it at some point.
2. I have a vague idea of perhaps rewriting the backend to either plain node.js (which would be ok, this would be easy to port) or Go. If that happens, some of this work may need to be redone. No timeline or decision on this, but just something to set expectations.

xmclark:
> Hey! Thanks for putting in this work.
> 
> Two risks:
> 
>     1. In the past I actually had an S3 backend, very similar to this one (it did have to rely on an additional database for keeping some meta data like file creation and modified dates which somehow couldn't reliably or efficiently be pulled from S3). Eventually I removed it because I didn't use it myself, some people ran into issues and I didn't have the time or mental space to maintain and test it a lot. That's why I deleted it. I'm mentioning this, because if we merge your PR, I will not have the time to test or maintain it either. If it causes problems for people and you (or other people) may not be around to support, I may delete it at some point.
> 
>     2. I have a vague idea of perhaps rewriting the backend to either plain node.js (which would be ok, this would be easy to port) or Go. If that happens, some of this work may need to be redone. No timeline or decision on this, but just something to set expectations.

Hey @zefhemel thanks for the comments. I meant to connect at some point. Totally cool with that position and not too surprising - supporting a feature for one user isn't time-well spent for a project like this. Since this feature may not be super desireable, I will close this draft PR. Bullet # 2 sounds interesting, looking forward to that. I think it would be great if the storage backend was definable in the the plug/extension model, that way this type of contribution doesn't need to be maintained long term.

PR #1523, opened 2025-09-06
support links in table queryable objects

ADobin:
I was having an issue where I was trying to use multiple tables to store names, links to external sites, and links to internal manuals I downloaded and then create a page that summarizes across those tables all of the items matching to display it in another table. The query would only return the specific text, not the full content of the cell, and I debugged and found what I think is the offending code.

I made sure to preserve what I believe was the original intent, to remove Hashtags from the result, since there are several examples of that behavior. But based on the original discussions in #725 and the implementation in #729 I'm not sure that was originally desired. Leaving it for now to isolate this change but happy to discuss more.

There are not any specific contributing guidelines, so I implemented what I thought looked like similar patterns that already existed. Happy to take feedback or additional examples for testing!

Before
image

After
image

### DIFF
--- a/plugs/index/table.ts
+++ b/plugs/index/table.ts
@@ -1,4 +1,5 @@
 import type { IndexTreeEvent } from "../../type/event.ts";
+import { renderToText, replaceNodesMatching } from "../../plug-api/lib/tree.ts";
 import { extractHashtag } from "../../plug-api/lib/tags.ts";
 import {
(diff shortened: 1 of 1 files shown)


zefhemel:
Thanks for this! I'm not a heavy user of tables myself and don't have strong opinions on how it should function. So let's merge this, and see what the people say :)

PR #1526, opened 2025-09-09
Update doc editors

MrMugame:
This removes the default html/skeleton for document editors and instead inserts the HTML directly.  All the api is added from the outside, which has some advantages and disadvantages:

**Disadvantages:**
- We can't sandbox the iframe anymore. It wasn't sandboxed anyways, we also couldn't sandbox it because of some key event issues anyways. We just now cannot sandbox it at all, because we are actively touching elements of the iframe from the outside. But I think this isn't really a goal anymore anyways. It would be easier to create some space lua via a syscall to escape the sandbox than to escape the iframe.
- The mental model when programming here isn't quite as clear.

**Advantages:**
- It's much much much easier to develop any document editor, because you don't have any weird load order issues anymore, because e.g. your style tags were inside of the body. Some things aren't even really possible the way it was before, especially proper ES6 module usage.
- Code size is reduced and we don't have this weird html inside a a string anymore. (The deno loader plugin style doesn't support import from text)
- The development experience is overall better, because now it's just like normal web development.

I made sure to be backwards compatible. The only plug I know that uses document editors, which isn't mine is the excalidraw plug and it works. I also removed an event `file-changed`, which will now just emit `file-open`, which shouldn't break anything, as nobody handled that event properly anyways.


zefhemel:
I like simplification, nice! Did some cursory testing and seems fine. I do see one warning in the console (Chromium), but I doubt we can do anything about that:

```
for image-viewer
about:srcdoc:1 An iframe which has both allow-scripts and allow-same-origin for its sandbox attribute can escape its sandboxing
```

PR #1541, opened 2025-09-13
Fix plug names

MrMugame:
The plug name and plug path were very intertwined in the code. This could cause problems, if the file path and plug name didn't align. The plug manager also modified the plug path for some plugs, which could be even more confusing.
I cleaned up some code around it, although there is so much code here that could just be dropped. Environments for plugs can be completely dropped for example (if there are not plans to reimplement this). The one place there name and path are still connected is for builtin plugs, because sometimes we need to know the name and sometimes the path. I just generated the paths from the names.
I also rewrote some stuff in the plug manager. It can now take a `{ code: string, name: string }` from the event and will use that name for the filename. This is used for the `ghr` case.
The only problem now is that plugs can overwrite other plugs, maybe this was possible before unsure. I added some protection against overwriting builtiin plugs.

### DIFF
--- a/cmd/test/runtime.test.ts
+++ b/cmd/test/runtime.test.ts
@@ -1,10 +1,10 @@
-import { createSandbox } from "../../lib/plugos/sandboxes/deno_worker_sandbox.ts";
 import type { SysCallMapping } from "../../lib/plugos/system.ts";
 import { System } from "../../lib/plugos/system.ts";
 import { assertEquals } from "@std/assert";
 import { compileManifest } from "../compile.ts";
 import * as esbuild from "esbuild";
 import { fileURLToPath } from "node:url";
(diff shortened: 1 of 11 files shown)


zefhemel:
Thanks for doing this. I think I see what you did here and the changes make sense. I did some local testing and all seems to work fine. I think Lazyily should be Lazily but I'll fix that in post.

Thanks!

zefhemel:
And indeed I think given the current state of plugs just running on the client (and I have no plans to change that), more cruft can be removed like all the environment stuff.

MrMugame:
> Thanks for doing this. I think I see what you did here and the changes make sense. I did some local testing and all seems to work fine. I think Lazyily should be Lazily but I'll fix that in post.

I do write like a toddler ~~sometimes~~ a lot.

PR #1542, opened 2025-09-13
Cleanup code around dark mode

MrMugame:
I don't want to offend anybody, but the code in the `useEffect` here, was really not it. This should also fix a bug, where you couldn't force white mode if your system is set to dark mode.
And maybe we should also add something to the docs about the use of the `clientStore` here, because this caused confusion multiple times now.

### DIFF
--- a/plugs/editor/editor.ts
+++ b/plugs/editor/editor.ts
@@ -9,8 +9,12 @@ export async function setEditorMode() {
   if (await clientStore.get("vimMode")) {
     await editor.setUiOption("vimMode", true);
   }
-  if (await clientStore.get("darkMode")) {
-    await editor.setUiOption("darkMode", true);
+  // Only set the darkmode value if it was deliberatly set in the clientstore,
+  // otherwise leave it so the client can choose depending on the system
+  // settings
+  const darkMode = await clientStore.get("darkMode");
+  if (darkMode != null) {
+    await editor.setUiOption("darkMode", darkMode);
   }
   const markdownSyntaxRendering = await clientStore.get(
     "markdownSyntaxRendering",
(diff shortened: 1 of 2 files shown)


plugs/editor/editor.ts
```diff
@@ -9,8 +9,12 @@ export async function setEditorMode() {
   if (await clientStore.get("vimMode")) {
     await editor.setUiOption("vimMode", true);
   }
-  if (await clientStore.get("darkMode")) {
-    await editor.setUiOption("darkMode", true);
+  // Only set the darkmode value if it was deliberatly set in the clientstore,
+  // otherwise leave it so the client can choose depending on the system
+  // settings
+  const darkMode = await clientStore.get("darkMode");
+  if (darkMode != null) {
```
zefhemel, line 16: The `!=` operator does not exist in my definition of JavaScript, surprised that lint even allows it 😆

zefhemel:
I have not put much thought into this, but I'm not sure we shouldn't deprecate clientStore altogether, since all our stores our now client-side anyway, maybe it should just be replaced with datastore everywhere? clientStore was always a bit weird because of the different key format (strings vs lists of strings).

Separate topic, I'm ok to merge this on its own.

plugs/editor/editor.ts
MrMugame: Yeah I wasn't 100% sure what this would/could return here. Because if you follow the definition, some are typed to return `null` and some return `undefined` and there are some parts were typescript doesn't check

MrMugame:
> I have not put much thought into this, but I'm not sure we shouldn't deprecate clientStore altogether, since all our stores our now client-side anyway, maybe it should just be replaced with datastore everywhere? clientStore was always a bit weird because of the different key format (strings vs lists of strings).

I guess it uses datastore under the hood anyways. I'm not sure how many people use it, I use it in silversearch, just because it's a little easier to use than the datastore api.

PR #1552, opened 2025-09-16
fix #1538 and all wikilinks in a frontmatter block are now navigable

### LINKED ISSUE #1538: Enable alt-clicking inside frontmatter links
@costalfy in other places in SB, when you Alt-click, you will always move your cursor to that location rather than activating e.g. a link. This is not currently the behavior in frontmatter links as of https://github.com/silverbulletmd/silverbullet/commit/4c95083403d7e5a4a25f3526a5ea030ac3c7bd14 could you have a look at this?

costalfy:
I think there is a better way to do this job, but until a future redesign, this feature can be tested.

image

### DIFF
--- a/web/cm_plugins/wiki_link.ts
+++ b/web/cm_plugins/wiki_link.ts
@@ -1,7 +1,7 @@
 import { syntaxTree } from "@codemirror/language";
 import { Decoration } from "@codemirror/view";
 import type { Client } from "../client.ts";
-import { decoratorStateField, isCursorInRange } from "./util.ts";
+import { decoratorStateField } from "./util.ts";
 import type { ClickEvent } from "@silverbulletmd/silverbullet/type/client";
 import { wikiLinkRegex } from "../markdown_parser/constants.ts";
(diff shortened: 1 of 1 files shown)


web/cm_plugins/frontmatter.ts
```diff
@@ -82,6 +99,124 @@ export function frontmatterPlugin() {
                 }).range(mFrom, mTo),
               );
             }
+
+            // 2) Internal links: WikiLinks [[...]] (make navigable)
```
zefhemel, line 98: This whole section seems to replicate largely what's already in wiki_link.ts do you see any way to extract this and reuse it in both? All this code duplication isn't great.

costalfy:
While waiting for a reply, I added this news admonitions slash command 
image

MrMugame:
? This seems very unrelated to the PR

What reply are you waiting for? I made a comment on community on ideas to avoid the code duplication. [See this comment](https://community.silverbullet.md/chat/c/-/2/2711)

zefhemel:
Ok, this seems good enough for me. The admonition stuff is pretty unrelated, but whatever. Thanks!

Issue #1549, opened 2025-09-15
Rewrite server in Go or Node.js


zefhemel:
Deno has served us well, but it is a constantly shifting ecosystem (regressions, new package manager approaches every few months) and a distraction that gives us little advantage over using more "boring technology". 

Currently, SilverBullet's server does very little:

1. Authentication (for those that use it)
2. Serve client (static) assets
3. Implement the HTTP CRUD API, which basically means listing, reading, writing and deleting files (on disk)
4. Implement a few specific endpoint like /.shell to run shell commands and /.proxy to proxy HTTP requests

These can be implemented in any technology. However there's a few nice to haves:

* Being able to compile into a single binary is a nice feature that Deno has provided. This make distribution easier.
* Boring technology: SilverBullet is challenge enough in itself, we don't need the distraction of just keeping up with the rapid ecosystem developments of the underlying technology

Options that I see:

1. Node.js: the obvious advantage here is that this is relatively mature, however the ecosystem keeps shifting. There are many server frameworks (express, fastify, next.js, ...) that constantly move from hot to not, this is rather tiring. It's a mild pain to use with TypeScript compared to Deno. The main advantage: this is still TypeScript/JavaScript so same stack as the client, _some_ code sharing can happen (although there's not that much code anyway). [node.js also appears to be getting a single binary option](https://nodejs.org/api/single-executable-applications.html).
2. Go: stable, mature, performant, boring. Compiles into a single statically linked binary (so we can also run it in smaller Docker images like alpine), low memory footprint. Disadvantage: another language for people need to know to contribute — although again, the server is small. 

I'm tempted to the Go option also because I have plans to work towards a hosted SilverBullet solution where I'd really like to handle as many clients (and spaces) on a single server using all of the server's resources (and not just a single thread like node.js). I also like to keep things as small as possible, and the packaged Go binary will probably be about 20MB (estimate) compared to about 85MB+ (guess based on the current Deno-size binary which is likely smaller than what node.js would be) and way lower on memory use.

zefhemel:
Backend is now rewritten in Go

PR #1555, opened 2025-09-17
Go backend

zefhemel:
Implements https://github.com/silverbulletmd/silverbullet/issues/1549

Should be functionally equivalent (only had to drop the `sync` sub command which I hope nobody used or even knew about anyway).

Fun facts:

* Server builds (single binary) are now about [~7MB zipped](https://github.com/silverbulletmd/silverbullet/releases/tag/edge-go) vs [~30MB zipped](https://github.com/silverbulletmd/silverbullet/releases/tag/edge) with Deno

This also changes a fair amount regarding development flow, please re-read the README on the tools and commands to use moving forward.

Docker builds will soon be available under the `:v2-go` tag.

Things to do after this is merged:

* [x] Reconsider the whole repo code structure, it's a bit messy now
* Consider building the main docker images from Alpine rather than ubuntu, because there's no longer a glibc dependency (maybe offer two versions, e.g. `-ubuntu` and `-alpine`)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - "main"
-      - "go-backend"
     tags:
       - "*"
 env:
@@ -63,7 +62,7 @@ jobs:
             # When pushing to main branch, release as :edge DISABLED FOR NOW
             # type=edge,branch=main
             # When pushing to main branch, release as :v2
-            type=raw,value=v2-go,enable=${{ github.ref == 'refs/heads/go-backend' }}
+            type=raw,value=v2,enable=${{ github.ref == 'refs/heads/main' }}
 
       - name: Build and push main docker images
         uses: docker/build-push-action@v4.0.0
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -4,8 +4,6 @@ on:
   push:
     tags:
       - "*"
-    branches:
-      - "go-backend"
 jobs:
   release:
     runs-on: ubuntu-latest


PowerUser64:
Let's Go! 🚀 Feel free to let me know when this is functional and I can provide beta testing.

zefhemel:
This is now ready for test @PowerUser64 you can build it yourself, or use the `:v2-go` docker tag. I've now switched to it on my own main instance and will test with it for a few days.

PowerUser64:
Beta test results are in!

It seems like I can't set the port for SB to run on. I do `make build` in the toplevel and then I run

```
SB_FOLDER=/...  ./silverbullet --hostname 127.0.0.1 --port 9073 --user u:pw
```

... and then it says:

```
SilverBullet is now running: http://localhost:3000
```

and then it exits with status code 0 (success), which is weird because I'm fairly certain it failed.

I have something running on port 3000, so I know it can't be using port 3000. Also, it says it's running on localhost when I told it to run on 127.0.0.1, so I think there's something going on here.

It also says that I can pass the `SB_FOLDER` path as an argument, but the help page doesn't seem to list an option for setting the folder. I think there's some stuff here that isn't implemented.

I'm not sure if the docker build will work yet, but I can test it if you'd like. (My prod isn't docker, so it's not just a drop-in replacement.)


I should note, it's also possible I'm doing something wrong, so feel free to correct anything. All I did to run was get deno and go (already had them), then run `make build`, then `./silverbullet` (as I wrote above).

zefhemel:
> It seems like I can't set the port for SB to run on. I do make build in the toplevel and then I run

Yes, I had also discovered this in the mean time, fixed this now.

Issue #1562, opened 2025-09-23
[BUG] Lua: compatibility issues


mjf:
I've found the following incompability with Lua 5.4.x in SilverBullet (edge build):

```markdown
Space Lua | Lua 5.4.x
-|-
${  1/ 0 ==  1/ 0 } | `true`
${  1/ 0 == -1/ 0 } | `false`
${  1/ 0 ==  1/-0 } | `true`
${  1/ 0 == -1/-0 } | `false`
${ -1/ 0 ==  1/ 0 } | `false`
${ -1/ 0 == -1/ 0 } | `true`
${ -1/ 0 ==  1/-0 } | `false`
${ -1/ 0 == -1/-0 } | `true`
${  1/-0 ==  1/ 0 } | `true`
${  1/-0 == -1/ 0 } | `false`
${  1/-0 ==  1/-0 } | `true`
${  1/-0 == -1/-0 } | `false`
${ -1/-0 ==  1/ 0 } | `false`
${ -1/-0 == -1/ 0 } | `true`
${ -1/-0 ==  1/-0 } | `false`
${ -1/-0 == -1/-0 } | `true`
```

([discussion](https://community.silverbullet.md/t/bug-broken-infinity-comparison-compatibility/3257))

zefhemel:
Some more compatibility issues:


# Lua 5.4 Compatibility Tests

## Numeric Model

- Numeric types (`math.type(n)`): #todo
- Modulo has divisor's sign: ${5 % -2 == -1 and -5 % 2 == 1}

- Coercion in arithmetic:
  ${{pcall(function() return "1" + 2 end)}}
  Expected:
  ${{true, 3}}

- `math.huge`: ${math.huge == 1/0} (`math.huge = 1/0` should go to the library)
- hex float literals: ${tonumber("0x1p2") == 4 and tonumber("0x1.8p1") == 3}

```lua
-- IEEE-754
do
  local nan1, nan2 = 0/0, 0/0
  assert(not (nan1 == nan1) and not (nan2 == nan2))

  local pz, nz = 0.0, -0.0
  assert(pz == nz)              -- 0 == -0
  assert(1/pz > 0 and 1/nz < 0) -- distinguish sign of zero

  local pinf, ninf, pinf2, ninf2 = 1/0, -1/0, 1/-0, -1/-0
  assert(pinf == pinf and ninf == ninf and pinf ~= ninf)
  assert(pinf == pinf2 and ninf == ninf2)
end
```

[...]

PR #1573, opened 2025-09-30
Fix arithmetic coercion with proper Lua semantic

mjf:
Tries to solve one part of the #1562. Also adds hexadecimal string numbers as supported by Lua (`tonumber()`). Not sure where else this should potentionaly apply.


mjf:
There are some simple tests for Space Lua expressions (they do not cover all possible combinations but that could be easily generated, if needed):

```
arithmetical operations:

${'1' + 1} ${'1' - 1} ${'1' * 1} ${'1' / 1} ${'1' // 1} ${'1' ^ 1} ${'1' % 1}
${1 + '1'} ${1 - '1'} ${1 * '1'} ${1 / '1'} ${1 // '1'} ${1 ^ '1'} ${1 % '1'}
${'1' + '1'} ${'1' - '1'} ${'1' * '1'} ${'1' / '1'} ${'1' // '1'} ${'1' ^ '1'} ${'1' % '1'}

${'+0x10' - 1} ${'-0x10' + 1} (signs work)

${'123 ' + 321} ${' 0xffff ' * 1} (trimming works)
```

**Note:** Any `${'123'}` still evaluates to `"string"`, which is correct as it should apply in arithmetical operations only (I used Lua 5.4 as reference implementation).

client/space_lua/eval.ts
```diff
@@ -435,13 +453,33 @@ const operatorsMetaMethods: Record<string, {
     sf: LuaStackFrame,
   ) => LuaValue;
 }> = {
-  "+": { metaMethod: "__add", nativeImplementation: (a, b) => a + b },
-  "-": { metaMethod: "__sub", nativeImplementation: (a, b) => a - b },
-  "*": { metaMethod: "__mul", nativeImplementation: (a, b) => a * b },
-  "/": { metaMethod: "__div", nativeImplementation: (a, b) => a / b },
+  "+": {
```
zefhemel, line 453: This may break some existing Lua code that people wrote because they don't realize the + operator in Lua is just for numbers, and `..` should be used to concatenate strings. I think this is ok though, you give an appropriate error.

zefhemel (changes requested):
Could you add some tests? I think the `language_core_test.lua` file would be the best place.

mjf:
@zefhemel I added some simple tests. But note that now only `Error` is thrown (because I didn't figured out how to get some `StackFrame` in the tests).

What's missing (issues pending):

- unary operator "-" should also call the coersion function (supported by Lua, the "+" operator isn't),

  ```lua
  > type(-'0x10')
  number
  
  > type(+'0x10')
  stdin:1: unexpected symbol near '+'
  ```

- the `tonumber()` isn't aligned with how the `tonumber()` in the native Lua works (needs discussion whether it should be replaced with some variant of the `luaToNumber()` provided in this patchset and unify everything on it)

-----

Funny note/message for others perhaps writting in `vi` or it's clones, remember `truncate -s -1 client/space_lua/language_core_test.lua`! :-D

client/space_lua/eval.ts
mjf: I see this as rather **very rare case** too personally. And the fix is easy enough (just replace "+" with "..", hopefuly). Should be OK, especially if proper error will be thrown to the users... I have no idea how to add negative tests though (to test error is really firing, for completeness). ;-)

zefhemel:
Nice so far, let me know if you want to still finish the remaining items in this PR or already merge this. I'm good with either.

zefhemel (approved):
Only minor nit that I have (and would otherwise fix myself): I like all my if-statements and else clauses to have curly braces, even if there's only one statement. Probably should make this a lint rule at some point.

client/space_lua/tonumber.ts
```diff
@@ -0,0 +1,57 @@
+/**
```
zefhemel, line 1: Maybe name this luaToNumber and put it in runtime.ts?

client/space_lua/eval.ts
```diff
@@ -30,6 +30,19 @@ import {
   ArrayQueryCollection,
   type LuaCollectionQuery,
 } from "./query_collection.ts";
+import {
+  toNumber,
+} from "./tonumber.ts";
+
+// Wrapper for arithmetical operators
+function luaCoerceToNumber(val: unknown): number {
```
zefhemel, line 38: And move this one to runtime.ts also (with the other luaTo* functions)

zefhemel:
So this one can be closed?

Issue #1574, opened 2025-09-30
Export to clipboard not working in Safari


noahlst:
On Safari, choosing export to clipboard causes a permissions error:

```
[Error] Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
	writeText
	editor.copyToClipboard (client.js:110:14786)
	syscall (client.js:62:4145)
	At (client.js:62:10487)
	(anonymous function) (client.js:62:16105)
	J1 (client.js:62:17476)
	(anonymous function) (client.js:62:15598)
```

<img width="1387" height="942" alt="Image" src="https://github.com/user-attachments/assets/56aee29e-36c4-48af-83b0-d907ad9ec6fb" />

Safari 26, macOS Tahoe 26, Silverbullet 2.1.5-0-g2b32ab6

Works in Chrome.

Safari is particularly strict about associating user gesture with Clipboard API calls - so I'm not sure what a path to a solution could be, but I just wanted to raise that this is happening.

zefhemel:
Yeah, not sure what to do here either... your PR at least communicates that it doesn't work. Maybe trigger this through a button press would be acceptable to Safari, but then you'd need buttons for everything that operates on the clip board...

PR #1575, opened 2025-09-30
Show an error alert on Clipboard API write exceptions

noahlst:
This PR adds an alert for errors that come from calling the Clipboard API `navigator.clipboard.write()` and `navigator.clipboard.writeText()`, such as when the API is not supported.

I've moved the success notification from the plug to the syscall, so that the notification only appears when there is no errors.

Helps with issue #1574.

### DIFF
--- a/client/plugos/syscalls/editor.ts
+++ b/client/plugos/syscalls/editor.ts
@@ -632,26 +632,18 @@ export function editorSyscalls(client: Client): SysCallMapping {
     "editor.openSearchPanel": () => {
       openSearchPanel(client.editorView);
     },
-    "editor.copyToClipboard": (_ctx, data: string | Blob) => {
-      if (typeof data === "string") {
-        navigator.clipboard.writeText(data)
-          .then(() => {
-            // Only show notification if the copy operation was successful
-            client.flashNotification("Copied Markdown to clipboard!")
-          })
-          .catch((e) => {
-            console.error(e);
-            window.alert(e)
-          })
-      } else {
-        navigator.clipboard.write([new ClipboardItem({[data.type]: data })])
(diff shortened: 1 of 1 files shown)


client/plugos/syscalls/editor.ts
```diff
@@ -634,9 +634,24 @@ export function editorSyscalls(client: Client): SysCallMapping {
     },
     "editor.copyToClipboard": (_ctx, data: string | Blob) => {
```
zefhemel: Stylistically it's nicer to now turn this syscall into an async one, you can just add `async` before `(_ctx, ...`).

Then you can use a proper try-catch to catch the errors.

client/plugos/syscalls/editor.ts
```diff
@@ -634,9 +634,24 @@ export function editorSyscalls(client: Client): SysCallMapping {
...
       if (typeof data === "string") {
-        navigator.clipboard.writeText(data);
+        navigator.clipboard.writeText(data)
+          .then(() => {
+            // Only show notification if the copy operation was successful
+            client.flashNotification("Copied Markdown to clipboard!")
+          })
+          .catch((e) => {
+            console.error(e);
+            window.alert(e)
```
zefhemel: This will probably fail the linter, just use `alert` (without window) or `client.flashNotification("your error", "error")`

zefhemel (changes requested):

client/plugos/syscalls/editor.ts
noahlst: Changed to flashNotification

zefhemel:
Very nice, thank you!

PR #1576, opened 2025-09-30
Fix arithmetic coercion - refactored and enhanced

mjf:
The previous implementation not correct in many ways:

- did not support `tonumber()`
- did not support unary minus
- could cause (minor) performance issues
- some tests failed (perhaps just my local messing, not sure)

Therefor I tried to completely refactor it. Please, review.


zefhemel:
Ah only noticed this one now, looks good but same comment of moving some code I added here applies here: #1573

Also, linting fails (run `make check` locally)

mjf:
All tests pass, linting also passes.

PR #1585, opened 2025-10-03
Space Lua: Remove unary plus 

mjf:
This removes the unary plus operation that is not supported by standard Lua.
Unary plus operation now yields parse error.

Simple script to run parser generator in TypeScript added and made it part
of the build process via new Deno task `gen-lua-parser`.

New generated Lua parser also added (although not really needed while now
it is part of the build process).


client/space_lua/gen_parser.ts
```diff
@@ -0,0 +1,16 @@
+const grammar = "client/space_lua/lua.grammar";
```
zefhemel: Honestly I don't think we need to do this, the grammar doesn't change that much and I think it's fine to run `make generate` when it happens and commit the result.

deno.json
```diff
@@ -2,15 +2,16 @@
   "name": "@silverbulletmd/silverbullet",
   "version": "2.1.6",
   "tasks": {
-    "build": "make clean && deno run -A build_plugs_libraries.ts && deno run -A build_client.ts",
+    "build": "make clean && deno task gen-lua-parser && deno run -A build_plugs_libraries.ts && deno run -A build_client.ts",
```
zefhemel: and as you can see this makes the current CI fail.

zefhemel (changes requested):
I think it's unnecessary to include the parser generator step in the standard build. The grammars are hardly ever updated.

Other than that I'm ok with this PR.

deno.json
mjf: Fixed.

client/space_lua/gen_parser.ts
mjf: Fixed, removed.

PR #1599, opened 2025-10-07
silverbullet: fix circular recent page deletion

okuvshynov:
Consider the scenario:
1. visit page 'foo'
2. visit page 'bar'
3. delete page 'bar' (and navigate to foo based on editor.getRecentlyOpenedPages)
4. delete page 'foo' (and navigate back to now empty 'bar', which is confusing)

This happens because deletion is not propagated to recentlyOpenPages.

In this fix we get the most recently opened page which still exists. Alternative way to fix it would be to push the deletion to the recentlyOpenPages.

### DIFF
--- a/plugs/editor/page.ts
+++ b/plugs/editor/page.ts
@@ -10,9 +10,12 @@ export async function deletePage() {
   }
   // Query for last
   const recentlyOpenedPages = await editor.getRecentlyOpenedPages();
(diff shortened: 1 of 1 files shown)


zefhemel:
I noticed this happening today actually. Nice one. Thanks!

okuvshynov:
Thank you!

PR #1600, opened 2025-10-08
Fix lua arithmetic model

mjf:
This PR aligns Space Lua arithmetic model with that of Lua 5.4.

### DIFF
--- a/client/space_lua/eval.ts
+++ b/client/space_lua/eval.ts
@@ -32,20 +32,14 @@ import {
   luaValueToJS,
   type NumKind,
   singleResult,
+  isPromiseLike,
+  hasCall,
 } from "./runtime.ts";
 import {
   ArrayQueryCollection,
   type LuaCollectionQuery,
 } from "./query_collection.ts";
 import { luaToNumber } from "./tonumber.ts";
-
-function isPromiseLike<T = unknown>(v: unknown): v is Promise<T> {
-  return typeof (v as any)?.then === "function";
-}
-
-function hasCall(v: unknown): v is ILuaFunction {
-  return !!v && typeof (v as any).call === "function";
-}
 
 interface LuaFunctionLike {
   body: LuaFunctionBody;
--- a/client/space_lua/runtime.ts
+++ b/client/space_lua/runtime.ts
@@ -16,11 +16,11 @@ export type LuaValue = any;
 export type JSValue = any;
 export type NumKind = "int" | "float" | "unknown";
 
-function isPromiseLike<T = unknown>(v: unknown): v is Promise<T> {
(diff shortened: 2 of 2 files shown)


zefhemel:
Ok, I looked through the code. Here is my thinking:

1. To me having exactly matching semantics between Space Lua and Lua is a nice to have. I don't see it as absolutely critical (but perhaps you see things that I don't or have a use case that makes this essential).  I don't expect people to do a lot of complicated math, or use a lot of off-the-shelf numeric heavy Lua libraries. I could be wrong. So far (except you) nobody brought up issues with how numbers are handled in Lua.
2. What you're doing here is technically impressive, but also adds a lot of complexity to the implementation.
3. I'm not sure if you're not making the whole evaluation slower. I haven't done any benchmarking or really gone into details, but if on every operator call you have to do deep analysis on what potential numeric kinds each branch could return, even following the call graph (which seems to be what you may be doing?), this sounds like potentially expensive to me. Which brings me back to (1) which is: is this all worth it?

mjf:
> 1. To me having exactly matching semantics between Space Lua and Lua is a nice to have. I don't see it as absolutely critical (but perhaps you see things that I don't or have a use case that makes this essential).  I don't expect people to do a lot of complicated math, or use a lot of off-the-shelf numeric heavy Lua libraries. I could be wrong. So far (except you) nobody brought up issues with how numbers are handled in Lua.

I am indeed doing some more advanced math in Space Lua where some of the flaws are appearing, at least when reaching edge-case assumptions. So the motivation is now clearer, I hope. I has been "bitten" several times by discrepancies between the two arithmetical models (as explained in the previous parser-only fix already applied).

The second think is purely pragmatic one: once user assumes she writes code in Lua, she also assumes that it has the Lua arithmetical model too. :-)

> 2. What you're doing here is technically impressive, but also adds a lot of complexity to the implementation.

I hope we can reduce the complexity to some minimum in the end...

There are also certain things that can be reverted to have less function calls in the end again, for example. These were not strictly neccessary for the actual arithmetical model alignment but were fixed to make the whole interpreter more correct (in the sense of types etc.).

> 3. I'm not sure if you're not making the whole evaluation slower. I haven't done any benchmarking or really gone into details, but if on every operator call you have to do deep analysis on what potential numeric kinds each branch could return, even following the call graph (which seems to be what you may be doing?), this sounds like potentially expensive to me. Which brings me back to (1) which is: is this all worth it?

I think it is worth of it. I/we can try to find different way of doing it. This is just a proof it can be done. I don't think the overall **code complexity** raised too much. The raised  executional complexity does not seem to hurt the overall performance, at least my eye cannot catch it. But that's of course no proof. I agree we definitely need some benchmarks to detect serious performance regressions. I will give a try to find some complex-enough one.

PR #1603, opened 2025-10-09
Align Lua arithmetic model

mjf:
- ensure Lua-like arithmetic semantics around zero, integer vs float, evaluation order
- evaluator can now distinguish integer results (where -0 should collapse to +0 and integer div/mod by zero raises an error) from float zeros (where -0 is preserved, allowing -Inf results like 1/(-0.0))
- new `luaToNumberDetailed` function for both the numeric value and its number type (the eval uses this once per string operand), the original function `luaToNumber` just wraps it now
- unary minus now collapses only integer zero, preserving float -0, and binary operators strictly evaluate left then right
- should have only negligable impact on performance (although performance tests are still missing)
- all tests and checks are passing

### DIFF
--- a/client/space_lua/runtime.ts
+++ b/client/space_lua/runtime.ts
@@ -389,14 +389,22 @@ export class LuaTable implements ILuaSettable, ILuaGettable {
       return value.then(() => this.rawSet(key, value));
     }
     if (typeof key === "string") {
-      this.stringKeys[key] = value;
+      if (value === null || value === undefined) {
+        delete this.stringKeys[key];
+      } else {
+        this.stringKeys[key] = value;
+      }
     } else if (Number.isInteger(key) && key >= 1) {
       this.arrayPart[key - 1] = value;
     } else {
       if (!this.otherKeys) {
         this.otherKeys = new Map();
       }
-      this.otherKeys.set(key, value);
+      if (value === null || value === undefined) {
+        this.otherKeys.delete(key);
+      } else {
+        this.otherKeys.set(key, value);
+      }
     }
   }
 
--- a/client/space_lua/stdlib/table.ts
+++ b/client/space_lua/stdlib/table.ts
@@ -1,6 +1,7 @@
 import {
   type ILuaFunction,
   LuaBuiltinFunction,
+  luaCall,
   type LuaEnv,
   luaEquals,
   luaGet,
@@ -175,7 +176,7 @@ export const tableApi = new LuaTable({
       const startIndex = fromIndex < 1 ? 1 : fromIndex;
       for (let i = startIndex; i <= tbl.length; i++) {
         const val = await luaGet(tbl, i, sf);
-        if (await criteriaFn.call(sf, val)) {
+        if (await luaCall(criteriaFn, [val], sf.astCtx!, sf)) {
           return new LuaMultiRes([i, val]);
         }
       }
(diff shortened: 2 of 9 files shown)


client/space_lua/arithmetic_test.lua
```diff
@@ -1,91 +1,178 @@
...
-assert_eq(  1/-0.0 ==  1/ 0,  false, 'float-int:  1/-0.0 ==  1/ 0')
-assert_eq(  1/-0.0 == -1/ 0,  true,  'float-int:  1/-0.0 == -1/ 0')
-assert_eq(  1/-0.0 ==  1/-0,  false, 'float-int:  1/-0.0 ==  1/-0')
-assert_eq(  1/-0.0 == -1/-0,  true,  'float-int:  1/-0.0 == -1/-0')
-
-assert_eq( -1/-0.0 ==  1/ 0,  true,  'float-int: -1/-0.0 ==  1/ 0')
-assert_eq( -1/-0.0 == -1/ 0,  false, 'float-int: -1/-0.0 == -1/ 0')
-assert_eq( -1/-0.0 ==  1/-0,  true,  'float-int: -1/-0.0 ==  1/-0')
-assert_eq( -1/-0.0 == -1/-0,  false, 'float-int: -1/-0.0 == -1/-0')
+-- 1. Integer vs float zero divisors
```
zefhemel, line 7: I'm just going to assume these tests are correct, you probably pulled them from some test suite somewhere?

client/space_lua/eval.ts
```diff
@@ -30,20 +30,130 @@ import {
...
+// Decide if unary minus zero should collapse
+function collapseUnaryMinus(
+  argExpr: LuaExpression,
+  value: number,
+  env: LuaEnv,
+): boolean {
+  if (value !== 0) {
+    return false;
+  }
+  if (argExpr.type === "String") {
```
zefhemel, line 130: Minor: this looks like it could be a `switch`

client/space_lua/eval.ts
```diff
@@ -309,6 +413,41 @@ export function evalExpression(
   }
 }
 
+function finishBinary(
```
zefhemel, line 416: This may need a better name (I don't understand what this mean) or at least a doc string

client/space_lua/runtime.ts
```diff
@@ -20,6 +20,8 @@ export interface ILuaFunction {
   call(sf: LuaStackFrame, ...args: LuaValue[]): Promise<LuaValue> | LuaValue;
 
   asString(): string;
+  // Static numeric return type tags
+  returnNumericTypes?: ("int" | "float" | null)[];
```
zefhemel, line 24: Lot of replication of "int" | "float" can probably be extracted into a type?

client/space_lua/runtime.ts
```diff
@@ -158,11 +177,41 @@ export function singleResult(value: any): any {
   }
 }
 
+// Static numeric tagging (for unary minus zero collapsing)
+function analyzeReturnNumericTypes(
+  body: LuaFunctionBody,
+): ("int" | "float" | null)[] | null {
+  const stmts = body.block.statements;
+  if (stmts.length !== 1 || stmts[0].type !== "Return") {
```
zefhemel, line 185: So this will only work for single statement function bodies, with just a `return`?

PR #1611, opened 2025-10-10
Space Lua: Align arithmetic model with standard Lua

mjf:
This patchset aligns Space Lua arithmetic model with standard Lua. It also extends arithmetic test suite for Space Lua.

### DIFF
--- a/client/space_lua/tonumber.ts
+++ b/client/space_lua/tonumber.ts
@@ -18,7 +18,7 @@ function skipSpace(s: string, i: number): number {
   const n = s.length;
   while (i < n) {
     const c = s.charCodeAt(i);
-    if (c === 32 || (c >= 9 && c <= 13)) {
+    if (c === 32 || (c >= 9 && c <= 13)) { // SP, HT, LF, VT, FF, CR
       i++;
     } else {
       break;
@@ -28,21 +28,21 @@ function skipSpace(s: string, i: number): number {
 }
 
 function charToDigitBase(c: number, base: number): number {
-  if (c >= 48 && c <= 57) {
+  if (c >= 48 && c <= 57) { // '0'..'9'
     const v = c - 48;
     if (v < base) {
       return v;
     } else {
       return -1;
     }
-  } else if (c >= 65 && c <= 90) {
+  } else if (c >= 65 && c <= 90) { // 'A'..'Z'
     const v = 10 + (c - 65);
     if (v < base) {
       return v;
     } else {
       return -1;
     }
-  } else if (c >= 97 && c <= 122) {
+  } else if (c >= 97 && c <= 122) { // 'a'..'z'
     const v = 10 + (c - 97);
     if (v < base) {
       return v;
@@ -66,10 +66,10 @@ function parseIntWithBase(
   }
 
   let sign = 1;
-  if (s.charCodeAt(i) === 45) {
+  if (s.charCodeAt(i) === 45) { // '-'
     sign = -1;
     i++;
-  } else if (s.charCodeAt(i) === 43) {
+  } else if (s.charCodeAt(i) === 43) { // '+'
     i++;
   }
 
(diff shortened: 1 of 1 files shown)


client/space_lua/eval.ts
```diff
@@ -30,18 +36,130 @@ import {
...
+  throw new LuaRuntimeError(
+    `attempt to compare ${ta} with ${tb}`,
+    sf.withCtx(ctx),
+  );
+}
+
+function luaUnaryMinus(
+  v: any,
+): number {
+  const { ax, aZeroKind } = ((): {
```
zefhemel: Why go through this self invoking function gymnastics?

client/space_lua/tonumber.ts
```diff
@@ -120,10 +121,10 @@ function parseInt(s: string): { ok: boolean; value: number } {
   }
 
   let neg = false;
-  if (s.charCodeAt(i) === 45) { // '-'
```
zefhemel, line 123: Why remove these comments?

client/space_lua/tonumber.ts
```diff
@@ -120,10 +121,10 @@ function parseInt(s: string): { ok: boolean; value: number } {
...
+  if (s.charCodeAt(i) === 45) {
     neg = true;
     i++;
-  } else if (s.charCodeAt(i) === 43) { // '+'
```
zefhemel, line 126: Why remove these comments?

client/space_lua/tonumber.ts
mjf: This comments were forgotten relicts (not speaking that all the ASCII values should have comments if we want to comment them, I can add them easily)... Or, if you want this cleaner/nicer, I can also make some `Enum` with numeric values for this. But I think to lookup ASCII value using e.g., `ascii(1)` command is so easy that I removed all these comments at some point and these two were left by accident only... Just tell me what to do with it - add comments everywhere / add `Enum` / keep it without comments completely... ?

client/space_lua/tonumber.ts
mjf: ditto

mjf:
We only add distinction for zero and only when needed in order to align with Lua. Number itself does not preserve int vs float but we do, via a Symbol tag, on boxed +0 and by recomputing "integerness" for non-zero values.

client/space_lua/eval.ts
mjf: This even seems to add some additional overhead. Fixed.

mjf:
The code seems to be working as expected (as I hope). There is some code left that **may** be deduplicated into helper wrappers but I am uncertain about this step for I do not know what the impact of extra function calls to these wrappers would be in terms of performance. It won't be hard to make this step later...

It does not break anything **for me** and also passes all checks and tests. I also do not see any performance impact **personally**. Both does not mean there is not some issue left, or some performance regression will occure.

This branch also adds missing `__unm` and `__bnot` metamethods. The implementation is still missing the `__len` and correct `#`. As I realized the `#` operator is misbehaving now (as well as the accompanied unimplemented `__len` metamethod) which should be fixed. But not here. I'd rather create separate FR for it (and perhaps take over it) for it is not strictly speaking an arithmetical operation and it will definitely need some extra work to make it aligned with Lua.

> `__len`: the length (`#`) operation. If the object is not a string, Lua will try its metamethod. If there is a metamethod, Lua calls it with the object as argument, and the result of the call (always adjusted to one value) is the result of the operation. If there is no metamethod but the object is a table, then Lua uses the table length operation (see [§3.4.7](https://www.lua.org/manual/5.4/manual.html#3.4.7)). Otherwise, Lua raises an error. -- [docs](https://www.lua.org/manual/5.4/manual.html#2.4)

client/space_lua/tonumber.ts
mjf: OK, just added them back and also added comments for all the ASCII numeric constants at places where it makes sense. Fixed.

client/space_lua/tonumber.ts
mjf: Resolved, fixed.

zefhemel:
Alright, this works for me. I mostly understand what's happening here. Thanks for the effort!

PR #1630, opened 2025-10-15
Space Lua: Fix truthiness semantic deviation

### LINKED ISSUE #1620: v2: Space Lua: Fix truthiness semantic deviation
I have just found critical semantic deviation in Space Lua.

In standard Lua (5.4) we see:

```lua
{} and 'foo' or 'bar'
```
```
foo
```

In Lua, any non-`nil` and non-`false` value is `true`, including empty tables (*sic!*), but in Space Lua we see:

```lua
${{} and 'foo' or 'bar'}
```
```
bar
```

This is a **major** semantic deviation that affects `if`, `while`, `and` and `or`.

mjf:
Fixes #1620.

### DIFF
--- a/client/space_lua/lua.test.ts
+++ b/client/space_lua/lua.test.ts
@@ -9,7 +9,11 @@ Deno.test("[Lua] Core language", async () => {
   await runLuaTest("./language_core_test.lua");
 });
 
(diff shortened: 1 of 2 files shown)


zefhemel:
Very thorough, thank you!

PR #1631, opened 2025-10-15
Space Lua: Add `load` function

### LINKED ISSUE #1619: v2: Space Lua: Add `load` builtin
Add `load` builtin that compiles a chunk of Space Lua code and returns a function. On syntax error it returns `nil` and an error string, somewhat similar to how standard Lua implements it:

```lua
a = load('return 1 + 2')
a()
```
```
3
```

And with, e.g., syntax error:

```lua
a, e = load('return +"1"') -- there is no unary + in Lua
e
```
```
[string "return +"1""]:1: unexpected symbol near '+'
```

The function returned by the `load` builtin will run in the global (`_GLOBAL`) environment captured from thread or in a new `LuaEnv()`.

mjf:
Fixes #1619.

### DIFF
--- a/client/space_lua/stdlib/load.ts
+++ b/client/space_lua/stdlib/load.ts
@@ -9,12 +9,22 @@ import { parse } from "../parse.ts";
 import { evalStatement } from "../eval.ts";
 
 // Returns a function (callable chunk) or (nil, "error message") pair.
-export function luaLoad(sf: LuaStackFrame, code: LuaValue): LuaValue {
+export function luaLoad(code: LuaValue, sf: LuaStackFrame): LuaValue {
   const s = typeof code === "string" ? code : String(code);
 
   try {
     const block = parse(s, sf.astCtx || {});
-    const globalEnv: LuaEnv = sf.threadLocal.get("_GLOBAL") || new LuaEnv();
+    const globalEnvMaybe = sf.threadLocal.get("_GLOBAL");
+
+    // Be vocal when no _GLOBAL is set
+    if (!globalEnvMaybe) {
+      console.warn(
+        "load() called without _GLOBAL in thread-local environment",
+      );
+      return new LuaMultiRes([null, "Global environment not set"]);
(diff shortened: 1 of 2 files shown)


client/space_lua/stdlib/load.ts
```diff
@@ -0,0 +1,33 @@
...
+  LuaEnv,
+  LuaMultiRes,
+  type LuaStackFrame,
+  type LuaValue,
+} from "../runtime.ts";
+import { parse } from "../parse.ts";
+import { evalStatement } from "../eval.ts";
+
+// Returns a function (callable chunk) or (nil, "error message") pair.
+export function luaLoad(sf: LuaStackFrame, code: LuaValue): LuaValue {
```
zefhemel: Minor thing: I tend to pass in the stack frame argument last, so I'd prefer flipping these arguments.

client/space_lua/stdlib/load.ts
```diff
@@ -0,0 +1,33 @@
...
+  const s = typeof code === "string" ? code : String(code);
+
+  try {
+    const block = parse(s, sf.astCtx || {});
+    const globalEnv: LuaEnv = sf.threadLocal.get("_GLOBAL") || new LuaEnv();
```
zefhemel: If there is no _GLOBAL set I think we need to be more vocal about it than to quietly fall back to an empty environment, either just error out or at least print a warning message (I've done this elsewhere).

client/space_lua/stdlib/load.ts
mjf: Fixed.

client/space_lua/stdlib/load.ts
mjf: Fixed (and you're right, I found something like that somewhere in `runtime.ts`). Do you want to change anything else? (`load_test.lua` still passing)

zefhemel:
Thank you!

PR #1632, opened 2025-10-15
Space Lua: Add `huge` constant and `type()` function to the `math` API

mjf:
To align better with the `math` Lua API this patch adds:
- `math.huge` constant, and
- `math.type()` function (to determine if a number is integer or float).


mjf:
This won't fit well into the #1611 and won't collide with other work, so I made it separate PR.

PR #1635, opened 2025-10-15
Expose codemirror's toggleComment command

onespaceman:


zefhemel:
Nice, now all this needs is a default command and key binding... Sadly Cmd-/ is taken :(

PR #1637, opened 2025-10-16
Space Lua: Align length features with standard Lua

mjf:
The overall behaviour of
- `#` operator and
- `__len` metamethod.

Both should be aligned with standard Lua now.

Once at it we also add
- `rawlen` function.

This function was missing.

We also add many tests.

### DIFF
--- a/client/space_lua/len_test.lua
+++ b/client/space_lua/len_test.lua
@@ -37,7 +37,45 @@ do
   assert_eq(#t, 42, "__len should override raw table length")
 end
 
--- 3. Non-strings/non-tables (must error)
+-- 2.1. Metamethod result handling
+do
+  local t = {}
+
+  setmetatable(t,
+  {
+    __len = function(_)
+      return 42, 1, 2, 3
+    end,
+  })
+
+  assert_eq(#t, 42, "__len must use only first return value")
+end
+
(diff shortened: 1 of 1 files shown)


client/space_lua/eval.ts
```diff
@@ -893,6 +894,45 @@ function luaOp(
...
+    return val.length;
+  }
+
+  // JS arrays
+  if (Array.isArray(val)) {
+    return val.length;
+  }
+
+  // Allow metatable `__len` if present for other values
+  const mt = getMetatable(val, sf);
```
zefhemel: Is there a case like this where `val` is not a LuaTable (already handled above)?

mjf: The second metatable code supports `__len` on non-tables values and it would only be redundant if the `setmetatable` doesn't allow metatables on non-table value... OTOH, if you think it's safe to be removed, I can remove it. It's there just to be sure.

zefhemel: Maybe there should be a check on setmetatable, but let's remove it here at least.

mjf: I moved the JS branch after the generic `__len` path now and keep (sic!) it. It's explained in the commit message why.

PR #1640, opened 2025-10-16
Optional client encryption

zefhemel:
## The use case
If you login to SB from a browser or computer you don't control, this is nice to enable. (Note to self: so probably it's an option you select at the login screen, not enable server-wide like now). On launch it will prompt you for an encryption key (you can pick whatever you like) and it will use this to encrypt all data kept locally (both files and the full index). This means that if somebody else gets access to your browser, they cannot read anything from IndexedDB without knowing this key. This would make it safe to use SB on public computers, which now is now a bad idea without resetting the client after use (since all content is kept in the browser).

## What it does
Enables optional client-side encryption (at the IndexedDB level, so encrypting both keys and values) of both the synced files as well as the index. Everything becomes garbelygook when you look at IndexedDB with this enabled.

The encryption key is only kept in memory on the service worker side, to allow multiple tabs and windows (as well as the sync engine) to share it without constant reprompting. The moment all clients disconnect (tabs close), the service worker within 5s will flush the key.

## UX
The UX of this isn't amazing because right now you have to enter a password every time you launch SilverBullet "from the cold" (no previous tabs open). I considered integrating this into the auth screen and use the same password as you enter there, but the persistence of the client encryption key is in memory only, where as the JWT authentication token is kept in a cookie and may live up to a week. Of course the encryption token could be pulled from the auth token somehow, but that would defeat the purpose of this safety measure.

Thinking about this now, it's probably best to indeed use the auth screen for this every time rather than the browser-level prompt, just to ask to reauthenticate and when auth succeeds use the entered user/password as encryption key.

## Cost
This will make everything slower, every index and file interaction now requires encryption and decryption. In basic testing this still seemed fine, but not sure how it will perform with large spaces.

## Configuration
This feature is enabled by setting the SB_ENCRYPTION_SALT variable:

```
export SB_ENCRYPTION_SALT=$(openssl rand -base64 16)
```

This needs to change into a checkbox on the auth page I think. Maybe labeled "Logging in from a public computer" or similar.

## Is this E2EE?
This is not end-to-end-encryption, it only encrypts data locally. Data on the server is still kept unencrypted. This would a separate effort (that I'm also thinking about, optionally).

## To do
* [x] Put this as an option on the auth screen, and rely on this auth screen as a source of username/password, which can also serve as encryption key.
* [x] Generate the salt on the server side, without requiring the SB_ENCRYPTION_SALT variable

### DIFF
--- a/client/plugos/syscalls/editor.ts
+++ b/client/plugos/syscalls/editor.ts
@@ -13,8 +13,8 @@ import {
   moveLineDown,
   moveLineUp,
   redo,
+  toggleComment,
   undo,
-  toggleComment
 } from "@codemirror/commands";
 import type { Transaction } from "@codemirror/state";
 import { EditorView } from "@codemirror/view";
@@ -600,7 +600,7 @@ export function editorSyscalls(client: Client): SysCallMapping {
       return toggleComment({
         state: client.editorView.state,
         dispatch: client.editorView.dispatch,
-      })
+      });
     },
     "editor.moveLineUp": () => {
       return moveLineUp({
--- a/client/service_worker.ts
+++ b/client/service_worker.ts
@@ -229,8 +229,6 @@ self.addEventListener("message", async (event: any) => {
           console.log("Enabled client-side encryption for synced files");
         }
 
-        console.log("Encryption key", encryptionKeyMemoryStore);
-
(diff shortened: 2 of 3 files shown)


Anachron:
Very nice, appreciated!

PR #1649, opened 2025-10-23
Add benchmarking support for Space Lua

mjf:
- add `bench/lua.test.ts` (almost perfect clone of the `client/space_lua/lua.test.ts` reworked for `Deno.bench` 100x)
- add `make bench` to the `Makefile`
- add task `bench` to the Deno JSON config
- fix `rawset` test to not print out


mjf:
We can use GitHub CI to run the `make bench` on PR and push and store the output of the `make bench` in PR comment.

@zefhemel It would be at least something and not nothing... :grimacing:

PR #1666, opened 2025-11-05
Space Lua: Perf improvements

mjf:
* Add result-or-promise (RP) exec model to interpreter (async-first).
* Add centralized table property access with fast misses.
* Update table string keys to be object map.
* Update error contexts to be more consistent and precise.
* Update _evaluator_ to return either values or _Promises_.
* Update func calls, unary and bin operations, and expression lists.
* Update _stdlib_ to the changes and improve errors.
* Add tests to validate async control flow and error contexts.
* Add dedicated benches that isolate the _runtime_ from parsing.


client/space_lua/ast_narrow.ts
```diff
@@ -0,0 +1,81 @@
+// Typed narrowers for AST nodes to avoid `(node as any)` in consumers.
+
+import type { ASTCtx, LuaExpression, LuaLValue, LuaStatement } from "./ast.ts";
+
+// Extract by `type` discriminant
+type NarrowByType<U, K extends U extends { type: infer T } ? T : never> =
```
zefhemel, line 6: Wow, I have no idea what this does 😆

zefhemel: But I'll accept it as useful.

client/space_lua/break_async.test.ts
```diff
@@ -0,0 +1,173 @@
...
+Deno.test("Nested for loops: inner break (sync body)", async () => {
+  const env = new LuaEnv(luaBuildStandardEnv());
+  await evalBlock(
+    `
+      local t = {1,2,3}
+      found = 0
+      for i = 1, #t do
+        for j = i + 1, #t do
+          found = found + 1
+          break
```
zefhemel, line 27: worth adding an addition statement after `break` (if Lua allows it) to make sure you never get there? Can be an `error("Shoulnd't happen")` type call? Same for test below.

client/space_lua/break_async.test.ts
```diff
@@ -0,0 +1,173 @@
...
+      until true
+    `,
+    env,
+  );
+  assertEquals(env.get("n"), 1);
+});
+
+Deno.test("For-in: custom iterator, Promise in body before break", async () => {
+  const env = new LuaEnv(luaBuildStandardEnv());
+  env.set(
```
zefhemel, line 113: Minor: every test in this file seems to have some type of "forced async" call, maybe extract it to the top module level to build an `env` with various useful utility functions for test purposes? Not critical, just an idea.

client/space_lua/eval.ts
```diff
@@ -451,84 +498,125 @@ function evalPrefixExpression(
...
-          key = singleResult(key);
-
-          return luaGet(table, key, sf.withCtx(e.ctx));
-        });
-      } else {
-        const table = singleResult(values[0]);
-        const key = singleResult(values[1]);
-        return luaGet(table, singleResult(key), sf.withCtx(e.ctx));
+      const ta = asTableAccess(e);
+      // Sync-first: evaluate object and key without allocating Promise when both are sync.
```
zefhemel, line 516: 🎉

client/space_lua/rp.ts
```diff
@@ -0,0 +1,33 @@
...
+export function isP<T>(v: RP<T>): v is Promise<T> {
+  return v !== null && typeof v === "object" &&
+    typeof (v as any).then === "function";
+}
+
+export function rpThen<A, B>(v: RP<A>, f: (a: A) => RP<B>): RP<B> {
+  return isP(v) ? (v as Promise<A>).then(f) : f(v as A);
+}
+
+export function rpAll<T>(arr: RP<T>[]): RP<T[]> {
```
zefhemel, line 19: Maybe a quick doc string on what this does?

client/space_lua/rp.ts
```diff
@@ -0,0 +1,33 @@
...
+      hasPromise = true;
+      break;
+    }
+  }
+  if (!hasPromise) {
+    // All sync values: return as-is
+    return arr as T[];
+  }
+
+  // At least one Promise: allocate a new array
```
zefhemel, line 32: Out of interest, are you doing this because of performance reasons? The "lazy" approach would simply be `return Promise.all([...out])` I suppose? If you feel you need to make a copy at all, can you not just do `return Promise.all(out)`?

client/space_lua/rp.ts
```diff
@@ -0,0 +1,33 @@
+// Result-or-Promise helpers
+
+export type RP<T> = T | Promise<T>;
+
+export function isP<T>(v: RP<T>): v is Promise<T> {
```
zefhemel: Would a simple `v instanceof Promise` not be sufficient (and maybe faster?) are there cases where "fake" promises are returned?

zefhemel: If we want to keep this function I'd prefer longer, more descriptive names like `isPromise`. We don't have to worry about code length, this will be minified anyway.

client/space_lua/eval.ts
```diff
@@ -451,84 +498,125 @@ function evalPrefixExpression(
...
-        const args = evalExpressions(e.args, env, sf);
-
-        if (args instanceof Promise) {
-          return args.then((args) =>
-            luaCall(prefixValue, [...selfArgs, ...args], e.ctx, sf)
-          );
+        const parts = fc.args.map((arg: LuaExpression) =>
+          evalExpression(arg, env, sf)
+        );
+        const argsVal = rpAll(parts);
```
zefhemel, line 580: Fun fact that still holds (I think): all arguments are evaluated (potentially async) in parallel, so this is a way to get parallel execution in general. Not actionable, just something I thought of when I originally implemented this too.

client/space_lua/eval.ts
```diff
@@ -451,84 +498,125 @@ function evalPrefixExpression(
...
+
+        const thenCall = (args: LuaValue[]) =>
+          luaCall(calleeVal, [...selfArgs, ...args], fc.ctx, sf);
+
+        if (isP(argsVal)) {
```
zefhemel: Not super in love with the code duplication in both paths here, but maybe there's performance reasons not to extract this into a (local) function?

client/space_lua/eval.ts
```diff
@@ -988,56 +1059,94 @@ function evalExpressions(
...
  * 2. with `returnOnReturn` set to `false` will throw a LuaReturn exception if a return statement is encountered
  */
-export async function evalStatement(
+export function evalStatement(
   s: LuaStatement,
   env: LuaEnv,
   sf: LuaStackFrame,
   returnOnReturn = false,
-): Promise<void | LuaValue[]> {
+): void | LuaValue[] | Promise<void | LuaValue[]> {
```
zefhemel, line 1083: It's weird to get excited about a return type change, but screw it: I'm very excited about this return type change and everything it embodies 🤣

client/space_lua/eval.ts
```diff
@@ -988,56 +1059,94 @@ function evalExpressions(
...
+      const apply = (values: LuaValue[], lvalues: { env: any; key: any }[]) => {
+        const ps: Promise<any>[] = [];
+        for (let i = 0; i < lvalues.length; i++) {
+          const r = luaSet(
+            lvalues[i].env,
+            lvalues[i].key,
+            values[i],
+            sf.withCtx(a.ctx),
+          );
+          if (r instanceof Promise) {
```
zefhemel: Calling back to my other comment on `isP` I think it would be good to standardize on one approach, either an immediate instanceof check, or isP check everywhere, no?

client/space_lua/eval.ts
```diff
@@ -988,56 +1059,94 @@ function evalExpressions(
...
+          }
+        }
+        if (ps.length) {
+          return Promise.all(ps).then(() => undefined);
+        }
+        return;
+      };
+
+      if (!(valuesRP instanceof Promise) && !(lvaluesRP instanceof Promise)) {
+        return apply(valuesRP, lvaluesRP as LuaLValueContainer[] as any);
```
zefhemel: does casting it to a LuaLValueContainer[] in between buy you anything?

client/space_lua/eval.ts
```diff
@@ -1047,38 +1156,155 @@ export async function evalStatement(
...
-        );
-        // Will only happen with `return` statement
-        if (result !== undefined) {
-          return result;
+      const stmts = b.statements;
+
+      const processFrom = (
+        idx: number,
+      ): void | LuaValue[] | Promise<void | LuaValue[]> => {
+        for (let i = idx; i < stmts.length; i++) {
```
zefhemel, line 1179: Interesting approach and maybe worth documenting a comment: as I understand it this defaults to simply running over the statement in a for loop (sync case) but then jumps to async mode when an async statement is encountered, continuing the statements from there. Clever.

client/space_lua/eval.ts
```diff
@@ -1047,38 +1156,155 @@ export async function evalStatement(
...
-        await evalStatement(s.elseBlock, env, sf);
-        return;
-      }
-      break;
+        const cv = evalExpression(conds[i].condition, env, sf);
+        if (cv instanceof Promise) {
+          return (cv as Promise<any>).then((val) => {
+            if (luaTruthy(val)) {
+              const r = evalStatement(conds[i].block, env, sf, returnOnReturn);
+              return r instanceof Promise ? r : r;
```
zefhemel: Ehm? Why not return `r` immediately (like in the `else` branch below)?

client/space_lua/eval.ts
```diff
@@ -1047,38 +1156,155 @@ export async function evalStatement(
...
+          return runFrom(i + 1);
+        }
+      };
+
+      return runFrom(0);
     }
     case "While": {
-      while (luaTruthy(await evalExpression(s.condition, env, sf))) {
+      const w = asWhile(s);
+      const runAsync = async () => {
```
zefhemel, line 1239: My recurring gripe of code duplication. I know the original implementation also had this, but having a way to afford it (without a performance hit) would be really nice, this is asking for issues down the line where one branch is updated, but the other isn't.

client/space_lua/eval.ts
```diff
@@ -1087,146 +1313,381 @@ export async function evalStatement(
...
+            ) {
+              break;
+            }
+            const localEnv = new LuaEnv(env);
+            for (let i = 0; i < fi.names.length; i++) {
+              localEnv.setLocal(fi.names[i], iterResult.values[i]);
+            }
+            try {
+              const r = evalStatement(fi.block, localEnv, sf, returnOnReturn);
+              if (r instanceof Promise) {
```
zefhemel: Something to maybe check: what if you do a straight up `await evalStatement` instead, would that have performance implications vs the current (more code) approach?

client/space_lua/rp.bench.ts
```diff
@@ -0,0 +1,270 @@
+// Benchmark suite for Space Lua RP (Result-or-Promise) optimizations
```
zefhemel, line 1: Great!

client/space_lua/runtime.ts
```diff
@@ -165,7 +189,7 @@ export class LuaFunction implements ILuaFunction {
     this.capturedEnv = closure;
   }
 
-  async call(sf: LuaStackFrame, ...args: LuaValue[]): Promise<LuaValue> {
+  call(sf: LuaStackFrame, ...args: LuaValue[]): Promise<LuaValue> | LuaValue {
```
zefhemel, line 195: Once again: exciting times due to a return type change!

client/space_lua/runtime.ts
```diff
@@ -281,17 +314,30 @@ export class LuaBuiltinFunction implements ILuaFunction {
...
+
   // When tables are used as maps, the common case is that they are string keys, so we use a simple object for that
   private stringKeys: Record<string, any>;
   // Other keys we can support using a Map as a fallback
   private otherKeys: Map<any, any> | null;
   // When tables are used as arrays, we use a native JavaScript array for that
   private arrayPart: any[];
 
+  // Metamethod presence flags
+  private _hasIndexMM = false;
```
zefhemel: why the `_` prefix all of a sudden? If you need enforce privacy (I don't think we do and we're not using elsewhere) there's also `#` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements

client/space_lua/runtime.ts
```diff
@@ -281,17 +314,30 @@ export class LuaBuiltinFunction implements ILuaFunction {
...
+  private _hasNewIndexMM = false;
+  private _hasLenMM = false;
+
   constructor(init?: any[] | Record<string, any>) {
     // For efficiency and performance reasons we pre-allocate these (modern JS engines are very good at optimizing this)
     this.arrayPart = Array.isArray(init) ? init : [];
-    this.stringKeys = init && !Array.isArray(init) ? init : {};
+    this.stringKeys = Object.create(null);
```
zefhemel: You're doing this for performance or correctness reasons?

client/space_lua/runtime.ts
```diff
@@ -304,20 +350,30 @@ export class LuaTable implements ILuaSettable, ILuaGettable {
     return (key instanceof Number ? Number(key) : key) - 1;
   }
 
-  get length(): number {
-    return this.arrayPart.length;
+  private refreshMetaFlags(): void {
```
zefhemel: This looks good, but I don't see it being called anywhere?

client/space_lua/runtime.ts
```diff
@@ -372,17 +445,35 @@ export class LuaTable implements ILuaSettable, ILuaGettable {
...
-    if (metatable && metatable.has("__newindex") && !this.has(key)) {
-      // Invoke the meta table!
-      const metaValue = metatable.get("__newindex", sf);
-      if (metaValue.then) {
+    const mt = this.metatable;
+    if (!mt) {
+      return this.rawSet(key, value);
+    }
+    // Refresh flags when used
+    this._hasNewIndexMM = mt.rawGet("__newindex") != null;
```
zefhemel: Nit: to me the `!=` operator doesn't exist in JavaScript, there's only `!==`

client/space_lua/runtime.ts
```diff
@@ -493,42 +584,53 @@ export function luaIndexValue(
...
+      return null;
+    }
   }
   // If not, let's see if the value has a metatable and if it has a __index metamethod
   const metatable = getMetatable(value, sf);
   if (metatable && metatable.has("__index")) {
     // Invoke the meta table
     const metaValue = metatable.get("__index", sf);
-    if (metaValue.then) {
+    if ((metaValue as any)?.then) {
```
zefhemel: definitely either use `isP` or instanceof Promise here

client/space_lua/runtime.ts
```diff
@@ -493,42 +584,53 @@ export function luaIndexValue(
...
       // Got a promise, we need to wait for it
-      return metaValue.then((metaValue: any) => {
-        if (metaValue.call) {
-          return metaValue.call(sf, value, key);
-        } else if (metaValue instanceof LuaTable) {
-          return metaValue.get(key, sf);
+      return (metaValue as Promise<any>).then((mv: any) => {
+        if (mv?.call) {
```
zefhemel, line 592: I know this is legacy code, but as you're changing it anyway this should probably use luaCall instead?

client/space_lua/runtime.ts
```diff
@@ -873,6 +986,41 @@ export function luaToString(
...
 
+export function getMetatable(
+  value: LuaValue,
+  sf?: LuaStackFrame,
+): LuaValue | null {
+  if (value === null || value === undefined) {
+    return null;
+  }
+  if (typeof value === "string") {
+    // Prefer per-thread cached string metatable if `_GLOBAL` available
```
zefhemel, line 989: Nice, so far I've been feeling very uncomfortable calling string methods via the `:` syntax because I know how inefficient this is (and now it gets better so 🎉 )

client/space_lua/runtime.ts
```diff
@@ -30,6 +31,29 @@ export interface ILuaGettable {
   get(key: LuaValue, sf?: LuaStackFrame): LuaValue | undefined;
 }
 
+// Small helpers for type safety/readability
+export function isILuaFunction(v: unknown): v is ILuaFunction {
+  return !!v && typeof (v as any).call === "function";
+}
+
+export function isLuaTable(v: unknown): v is LuaTable {
```
zefhemel, line 39: Now that this exists (and I assume there's no significant performance impact on using this rather than inlining this expression) perhaps do a quick codebase scan for all `instanceof LuaTable` checks and replace it with this (or if there's perf impact: the reverse to inline it)

client/space_lua/stdlib.ts
```diff
@@ -155,7 +159,7 @@ const setmetatableFunction = new LuaBuiltinFunction(
     if (!metatable) {
       throw new LuaRuntimeError("metatable cannot be set to nil", sf);
     }
-    table.metatable = metatable;
+    (table as any).metatable = metatable;
```
zefhemel: You do all this fancy typescript wizardry and then I see this 🤣  is this necessary, can this not be better type? I'm worried about refactoring later (renaming `metatable` to something else) and this not being caught due to all the `as any ` shenanigans.

zefhemel (changes requested):
A few things to start:

1. This is _absolutely fantastic_ I almost have tears in my eyes. There's so much in here that I would want to do some day but never got around to, and you're doing it. Absolutely awesome work!
2.  That's it, let's wrap this up and merge it ASAP 🤣 

A few thoughts, some potentially useful tests to add:

* Now that there's more cases where we can do execution fully synchronously, perhaps it makes sense to have unit test to make sure the implementation actually results in an ssync execution? Tests that evaluate a complicated program that _should_ be executable fully sync, and then make sure that if you eval it doesn't return a promise. That sort of thing.

Some general questions (also mention in individual comments):

* There's a bunch of parts of the implementations with a lot of code duplication between the sync and async path it would be really nice for everybody to find a way to remove this duplication, but not if it comes at a significant performance cost and I think with your benchmarking setup we now finally have a way to check this.
* Inconsistent use of isP and `instanceof Primise` checking, would be good to unify, similar story for the LuaTable instance checks.

Again, thank you so much for this, this is super duper awesome!

Implementing `goto` next? 🤣

client/space_lua/ast_narrow.ts
mjf: The idea was to refactor code a little to be both more maintainable in the future (easier refactors if AST evolves) and much cleaner and especially to elimitate all the `(foo as any).bar`s across `eval.ts`, achieve better type safery.  This approach can be extended far more further. In this patchset, only what I was digging in is generally touched by this, but there are many other places where these can be also used. I plan to refactor rest of the code in `client/space_lua` soon (not strictly needed right now, the helpers are de facto pure casts only).

client/space_lua/break_async.test.ts
mjf: Good idea, I will look into this.

client/space_lua/eval.ts
mjf: :rocket:

client/space_lua/rp.ts
mjf: Will do.

client/space_lua/rp.ts
mjf: WIll do.

client/space_lua/eval.ts
mjf: Fixed on several places.

client/space_lua/eval.ts
mjf: Tried to add it everywhere. But this is for some further review of the whole `client/space_lua` code base. There are more things that should be refactored...

client/space_lua/eval.ts
mjf: Documented.

client/space_lua/rp.ts
mjf: Documented.

client/space_lua/rp.ts
mjf: Done.

client/space_lua/eval.ts
mjf: Unconditionally awaiting will always create/resolve a Promise and incur a microtask hop even for synchronous results which we want to avoid.

client/space_lua/eval.ts
mjf: Leaving code duplication to some further refactoring steps I plan to do in near future (jotting down to my todo list in SB). :)

client/space_lua/rp.ts
mjf: The `as T[]` is just a type assertion.

client/space_lua/eval.ts
mjf: Yeah. Let's leave that for further refactoring right now. There are more such places. I think I will PR something like "Space Lua: Dedup dups" or something like that soon too.

client/space_lua/runtime.ts
mjf: Correctness. I reverted it.

client/space_lua/runtime.ts
mjf: Yeah! That's right! :-D Typos fixed.

client/space_lua/runtime.ts
mjf: Fixed as non-priv.

client/space_lua/runtime.ts
mjf: Argh! The original idea was to cache metamethod presence (`__index`, `__newindex`, `__len`) on the table to fast‑path lookups/sets without repeated `rawGet` checks. With the current code recomputing on demand (and checking metatable directly), it isn’t needed, so it’s safe to remove. Therefor removing it. :+1:

client/space_lua/runtime.ts
mjf: That would not be as easy. I already tried that but something went terribly wrong. -> TODO list.

client/space_lua/break_async.test.ts
mjf: Won't fix.

mjf:
> Implementing goto next?

Sure! That's high on my todo list.

client/space_lua/break_async.test.ts
mjf: There are many such places over all the test suite. Needs to be implemented consistently, I think. -> TODO

client/space_lua/stdlib.ts
mjf: Weird quirk fixed.

client/space_lua/runtime.ts
mjf: fixed

mjf: updated

client/space_lua/runtime.ts
mjf: Fixed on both places.

client/space_lua/break_async.test.ts
mjf: Done.

client/space_lua/rp.ts
```diff
@@ -2,19 +2,24 @@
 
 export type RP<T> = T | Promise<T>;
 
-export function isP<T>(v: RP<T>): v is Promise<T> {
+// Returns true when v is a Promise or a has a then function.
+export function isPromise<T>(v: RP<T>): v is Promise<T> {
```
zefhemel, line 6: Not a blocker for merging but I'm still wondering what the scenarios are where v is not an instance of Promise, but _does_ have a .then method, and if those cases don't exist if simply doing an instanceof check inline doesn't make more sense (and is slightly more readable)?

client/space_lua/runtime.ts
```diff
@@ -323,14 +323,15 @@ export class LuaTable implements ILuaSettable, ILuaGettable {
   private arrayPart: any[];
 
   // Metamethod presence flags
-  private _hasIndexMM = false;
-  private _hasNewIndexMM = false;
-  private _hasLenMM = false;
+  hasIndexMM = false;
```
zefhemel, line 329: I'd still keep them `private` to be honest unless you need to access them externally, just to communicate intent.

zefhemel:
Amazing, very excited about this!

PR #1675, opened 2025-11-08
More robust "offline" handling when using proxies

lutzky:
Identify more HTTP statuses as indicating "offline", specifically ones delivered by proxies such as cloudflare.


zefhemel:
Nice one, thanks

PR #1678, opened 2025-11-10
ui.touch.bindings

jdiercks:
Replace hard-coded 2/3-finger touch gestures with new config.set options to map multi-touches to any command (or disable).

### DIFF
--- a/client/editor_ui.tsx
+++ b/client/editor_ui.tsx
@@ -456,14 +456,6 @@ export class MainUI {
       }
     }
   }
-
-  destroy() {
-    try {
-      this._touch?.dispose();
-    } catch {
-      // ignore
-    }
-  }
 }
 
 // TODO: Parking this here for now, this is very similar to the definition in top_bar.tsx
--- a/client/input/touch_registry.ts
+++ b/client/input/touch_registry.ts
@@ -3,9 +3,9 @@
 // Later sources override earlier ones per finger count.
 
 export type TouchBinding = {
-  fingers: number;                 // e.g. 2 or 3
-  command: string;                 // e.g. "Navigate: Page Picker" | "Command: Open Palette" | other command name
-  preventDefault?: boolean;        // default true
+  fingers: number; // e.g. 2 or 3
+  command: string; // e.g. "Navigate: Page Picker" | "Command: Open Palette" | other command name
+  preventDefault?: boolean; // default true
 };
 
 export type TouchMapEntry = { command: string; preventDefault: boolean };
@@ -22,7 +22,7 @@ export const defaultBindings: TouchBinding[] = [
 function merge(list: TouchBinding[], into: Map<number, TouchMapEntry>) {
   for (const b of list) {
     if (!b) continue;
-    const pd = (b.preventDefault ?? true);
+    const pd = b.preventDefault ?? true;
     into.set(b.fingers, { command: b.command, preventDefault: pd });
   }
 }
--- a/client/input/touch_router.ts
+++ b/client/input/touch_router.ts
@@ -3,53 +3,52 @@
 
 import type { TouchMapEntry } from "./touch_registry.ts";
 import { buildTouchMap } from "./touch_registry.ts";
+import type { Client } from "../client.ts";
 
-type ClientLike = {
-  config: { get<T>(path: string, def: T): T };
-  startPageNavigate: (mode: "page" | "meta" | "document" | "all") => void;
-  startCommandPalette: () => void;
-  runCommandByName: (name: string, args?: any[]) => Promise<void>;
-  clientSystem?: { commandHook?: { on?: (h: { commandsUpdated: () => void }) => void } };
-};
-
-export function setupTouchRouter(client: ClientLike) {
+export function setupTouchRouter(client: Client) {
   if ((navigator as any).maxTouchPoints <= 0) {
     return { dispose: () => {}, refresh: () => {} }; // noop
   }
 
-  function readConfigBindings(): { fingers: number; command: string; preventDefault?: boolean }[] {
-    return client.config.get("ui.touch.bindings", [] as any[]);
+  function readConfigBindings(): {
+    fingers: number;
+    command: string;
+    preventDefault?: boolean;
+  }[] {
+    return client.config.get("ui.touch.bindings", []);
   }
 
   function compile(): Map<number, TouchMapEntry> {
     const cfg = readConfigBindings();
-    return buildTouchMap(cfg);
+    try {
+      map = buildTouchMap(cfg);
+    } catch {
+      client.flashNotification(
+        "unexpected errur in touch_router:compile()",
+        "error",
+      );
+    }
+    return map;
   }
 
   let map: Map<number, TouchMapEntry>;
 
-  try {
-    client.clientSystem?.commandHook?.on?.({
-      commandsUpdated: () => {
-        try { map = compile(); } catch {}
-      },
-    });
-  } catch {}
+  client.clientSystem?.commandHook?.on?.({
+    commandsUpdated: () => {
+      map = compile();
+    },
+  });
   // Optional: one-shot delayed refresh for slow boots
-  setTimeout(() => { try { map = compile(); } catch {} }, 1500);
-
-  try {
-    map = compile();
-  } catch {
-    map = buildTouchMap(readConfigBindings());
-  }
+  // setTimeout(() => { map = compile(); }, 1500);
 
   const onTouchStart = (ev: TouchEvent) => {
-    const n = (ev.touches?.length ?? 0);
+    const n = ev.touches?.length ?? 0;
     if (!n) return;
     let binding = map.get(n);
     if (!binding) {
-      try { map = compile(); } catch { /* ignore */ }
+      try {
+        map = compile();
+      } catch { /* ignore */ }
       binding = map.get(n);
       if (!binding) return;
     }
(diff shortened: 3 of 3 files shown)


zefhemel:
Oh this is cool. I'll have a look soon.

jdiercks:
A bonus feature that I forgot to document:
4/5-finger taps are also supported! Example:

```lua
config.set("ui.touch.bindings", {
  -- disable the default 2-finger tap mapping, preserve browser pinch-zoom
  { fingers = 2, command = "",                           preventDefault = false },
  -- add custom mappings for 3-5 fingers
  { fingers = 3, command = "Editor: Toggle Reader Mode", preventDefault = true }, 
  { fingers = 4, command = "Navigate: Page Picker",      preventDefault = true },
  { fingers = 5, command = "Open Command Palette",       preventDefault = true }, 
})
```

client/editor_ui.tsx
```diff
@@ -466,6 +456,14 @@ export class MainUI {
       }
     }
   }
+
+  destroy() {
```
zefhemel: This looks nice, but this is never called, nor does it really need to be so I think this can be removed.

client/input/touch_router.ts
```diff
@@ -0,0 +1,91 @@
+// client/input/touch_router.ts
+// A single non-passive touchstart listener that dispatches actions based on finger-count.
+
+import type { TouchMapEntry } from "./touch_registry.ts";
+import { buildTouchMap } from "./touch_registry.ts";
+
+type ClientLike = {
```
zefhemel: I don't see why you need this, why not simply import type { Client }?

client/input/touch_router.ts
```diff
@@ -0,0 +1,91 @@
...
+  clientSystem?: { commandHook?: { on?: (h: { commandsUpdated: () => void }) => void } };
+};
+
+export function setupTouchRouter(client: ClientLike) {
+  if ((navigator as any).maxTouchPoints <= 0) {
+    return { dispose: () => {}, refresh: () => {} }; // noop
+  }
+
+  function readConfigBindings(): { fingers: number; command: string; preventDefault?: boolean }[] {
+    return client.config.get("ui.touch.bindings", [] as any[]);
```
zefhemel: `as any[]` not required here (just checked)

client/input/touch_router.ts
```diff
@@ -0,0 +1,91 @@
...
+      },
+    });
+  } catch {}
+  // Optional: one-shot delayed refresh for slow boots
+  setTimeout(() => { try { map = compile(); } catch {} }, 1500);
+
+  try {
+    map = compile();
+  } catch {
+    map = buildTouchMap(readConfigBindings());
```
zefhemel: looking at what `compile` does (= exactly what's in this line, just longer?), doesn't this effectively run compile twice hoping for a different result?

client/input/touch_router.ts
```diff
@@ -0,0 +1,91 @@
...
+    const cfg = readConfigBindings();
+    return buildTouchMap(cfg);
+  }
+
+  let map: Map<number, TouchMapEntry>;
+
+  try {
+    client.clientSystem?.commandHook?.on?.({
+      commandsUpdated: () => {
+        try { map = compile(); } catch {}
```
zefhemel: Quite uncomfortable with the silent ignoring of exceptions here (and other places, see lint errors). Can we actually expect them, and if so, shouldn't we report them?

client/input/touch_router.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    // Disabled if empty or "none"
+    if (!name || name === "none") return;
+
+    if (binding.preventDefault) {
+      ev.preventDefault();
+      ev.stopPropagation();
+    }
+
+    if (name === "Navigate: Page Picker") {
```
zefhemel: All these `if` checks seem redundant because the `else` handles them all?

client/input/touch_router.ts
```diff
@@ -0,0 +1,91 @@
...
+      client.startPageNavigate("page");
+    } else if (name === "Command: Open Palette") {
+      client.startCommandPalette();
+    } else {
+      // ✅ Run any command by name (no palette fallback unless you want one)
+      try {
+        // fire-and-forget is fine; or await if you prefer
+        void client.runCommandByName(name);
+      } catch (_e) {
```
zefhemel: Don't like failing quietly here

zefhemel (changes requested):
Overall really like this feature, but a few things:

1. Please run `make fmt` to format the code
2. Please run `make check` to run the linters (they fail)
3. Have a look at my comments, some weird stuff in there (do I smell AI coding agents?)

jdiercks:
> Overall really like this feature, but a few things:
>     1. Please run `make fmt` to format the code
>     2. Please run `make check` to run the linters (they fail)
>     3. Have a look at my comments, some weird stuff in there (do I smell AI coding agents?)

Zef, you're absolutely right, I relied heavily on ChatGPT in putting this PR together. Your analysis reveals that I am even more ignorant than I thought about TypeScript, and you've exposed my relative inexperience at robust programming practices in general. All of your comments make sense, and I will work on tightening this up. Learning opportunity! :-)

Glad you like the concept; I'm definitely scratching my own itch here, but without the ability to unmap 2-finger tap, SB would be almost unusable for me on mobile, because I'm constantly pinch-zooming. Thanks.

jdiercks:
> All of your comments make sense, and I will work on tightening this up. Learning opportunity! :-)

Think I've addressed all the issues you raised, and ran make fmt/clean which passes now. See commit v1.3 just pushed.

client/input/touch_registry.ts
```diff
@@ -0,0 +1,49 @@
...
+export type TouchBinding = {
+  fingers: number; // e.g. 2 or 3
+  command: string; // e.g. "Navigate: Page Picker" | "Command: Open Palette" | other command name
+  preventDefault?: boolean; // default true
+};
+
+export type TouchMapEntry = { command: string; preventDefault: boolean };
+
+/** Built-in fallbacks that mirror current behavior */
+export const defaultBindings: TouchBinding[] = [
```
zefhemel, line 14: I think it would be better to move this to Config.md. Have a look at the "Default values" header there where actual default configuration settings are set. This is more transparent than "hiding" it here.

docs/Touch Bindings.md
```diff
@@ -0,0 +1,24 @@
+# Touch Bindings
```
zefhemel, line 1: Rather than introducing a new `docs` folder I'd rather this to be somewhere in `website`

libraries/Library/Std/Config.md
```diff
@@ -384,3 +384,41 @@ config.set {
...
+Default (mirroring current behavior):
+
+```lua
+config.set("ui.touch.bindings", {
+  { fingers = 2, command = "Navigate: Page Picker", preventDefault = true },
+  { fingers = 3, command = "Command: Open Palette", preventDefault = true },
+})
+```
+
+Schema (reference):
```
zefhemel, line 403: Why not actually add it to the schema with `config.define` together with all the other schema definitions earlier in this file? Then there's value in that these configuration values are actually checked. Also see my comment on defaults earlier.

client/input/touch_registry.ts
```diff
@@ -0,0 +1,49 @@
...
+  { fingers: 2, command: "Navigate: Page Picker", preventDefault: true },
+  { fingers: 3, command: "Command: Open Palette", preventDefault: true },
+];
+
+/**
+ * Merge a list of TouchBinding into a map, overriding existing entries for the same fingers.
+ */
+function merge(list: TouchBinding[], into: Map<number, TouchMapEntry>) {
+  for (const b of list) {
+    if (!b) continue;
```
zefhemel, line 24: Not super important but this check seems redundant. Schema checking will ensure that touch bindings follow the right format (including if they're not `null`)

client/input/touch_registry.ts
```diff
@@ -0,0 +1,49 @@
...
+export function buildTouchMap(
+  configBindings: TouchBinding[] = [],
+): Map<number, TouchMapEntry> {
+  const map = new Map<number, TouchMapEntry>();
+
+  // 1) Built-in defaults
+  merge(defaultBindings, map);
+
+  // 2) Config-provided bindings (if any)
+  if (Array.isArray(configBindings)) {
```
zefhemel, line 44: Unless we don't trust the runtime, I think we can be sure that `configBinding` is actually an array (since that is its type), so no need for this check.

client/input/touch_router.ts
```diff
@@ -0,0 +1,85 @@
...
+    return client.config.get("ui.touch.bindings", []);
+  }
+
+  function compile(): Map<number, TouchMapEntry> {
+    const cfg = readConfigBindings();
+    try {
+      map = buildTouchMap(cfg);
+    } catch {
+      client.flashNotification(
+        "unexpected errur in touch_router:compile()",
```
zefhemel, line 27: -> `error`

client/input/touch_router.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (binding.preventDefault) {
+      ev.preventDefault();
+      ev.stopPropagation();
+    }
+
+    try {
+      client.runCommandByName(name);
+    } catch {
+      client.flashNotification(
+        "unexpected errur in touch_router:onTouchStart()",
```
zefhemel, line 70: `error`

PR #1686, opened 2025-11-13
Space Lua: Add `goto` and labels support

mjf:
- Add `goto` and labels support to Space Lua with proper standard Lua 5.4 semantics.
- Add extensive test suite, including some take from the original Lua 5.4 test suite (all existing tests are pass ok).
- The performance impact is negligable (certain paths seem to gain even better performance).


client/space_lua/eval.ts
```diff
@@ -1157,49 +1203,130 @@ export function evalStatement(
...
-        for (let i = idx; i < stmts.length; i++) {
-          const result = evalStatement(
-            stmts[i],
-            newEnv,
-            sf,
-            returnOnReturn,
+
+      // Parses flags
+      type BlockFlags = LuaBlock & {
+        hasGoto?: boolean;
```
zefhemel: these attributes are already in LuaBlock, no?

zefhemel:
Ok, as usual, first of all: thank you for this effort.

Thinking out loud: this adds 1k+ lines of code for a feature that I honestly hope few people will actually use. So not wild about that, but still it can be worth and nice for Lua compatibility.

I see a lot of this

```
if (isGotoSignal(res)) {
  return res;
}
```

I know why and it may be performant, but it's also very redundant and ugly. I think a `goto` is very similar to a `return` in this sense and for that I use throw a `LuaReturn` as an exception to "escalate" a return from any level of code without needing all the return shenanigans that you're introducing here. Throwing an exception is obviously expensive and I'm not wild about it, but it keeps the code reasonably readable. I expect the use of `goto` be uncommon enough and not to happen in "hot" code paths, so perhaps using a `LuaGoto` as an exception could clean this up also?

mjf:
You can never know how will people use any language feature once you introduce it. But what's clear are the expectations. I wouldn't expect goto to cause any overhead personally. And it does, so I tried hard to get rid of it. What you are suggesting will bring some overhead back in case hot paths are frequent in some code.

I also struggle between the two points of view: On one hand I am tempted to avoid any additional overhead everywhere, which is indeed doable but will lead to exactly this effect - the repetitive code everywhere. It might seem less readable in the sense of the TypeScript idiomatics but it's also more low-level language like C-like which is sort of expected in case of a language interpreter. So I struggle to choose between the two coding styles, but not as much as you do.

What's clear and proven now is that the approach you suggest may lead to unexpected performance loss. Consider, e.g., somebody crazy enough wrote parser in C-like style, perhaps just rewrote one generated by Ragel or similar tool in C to Space Lua. That's exactly the case where the performance impact will be huge, if she used the goto-style of automaton.

Therefor, and for other reasons, I would beg for the Space Lua interpreter be treated as something that, although fully integrated in the SilverBullet, can divert from the overall coding style, to divert from idiomatic TS. Once most of the hidden corners reveal, there won't be a need to touch it for a long period of time for we chosen to support relatively recent version of Lua (5.4). The C-like "linearity" of the code seems as better and more performant approach from the long term perspective to me.

Now, the real issue with the code I see is that it needs to be refactored one or the other way. It uses both of the approaches and this unintended (sic!) mixture of both hurts the comprehension and readability and feels "weird". Either way it will be refactored, it will gain better comprehensibility in the end (and although the C-like style may be less accessible to the TS developers, it will bring the best possible and expectable performance without side effects and also V8 will benefit from it - better optimizations, less stack use, et cetera).

mjf:
@zefhemel The code changes in evaluation to achieve what you would probably like to see are pretty invasive and non-trivial because every place that currently returns, checks or forwards a GotoSignal (sync or using promise) must be updated to use try/catch for LuaGoto (including Promise branches .then/.catch) so that goto exceptions are consumed locally when a label exists and re-thrown othewise.

zefhemel:
Ok, you convinced me. I do think we should implement the same for `return` then in time, actually using exceptions to implemented "nested" returns has mentally been bugging me for a long time and this is actually something that I know is happening in "hot paths". Doesn't need to be part of this PR of course.

client/space_lua/ast.ts
```diff
@@ -12,6 +12,12 @@ export type ASTCtx = {
 export type LuaBlock = {
   type: "Block";
   statements: LuaStatement[];
+  hasLabel?: boolean;
+  hasGoto?: boolean;
+  // the first duplicated label is recorded here for later eval
+  dupLabelError?: { name: string; ctx: ASTCtx };
+  // this block declares top-level locals and therefore needs a new env
+  needsEnv?: boolean;
```
zefhemel, line 20: This looks like a nice optimization opportunity! ❤️

client/space_lua/eval.ts
```diff
@@ -71,6 +74,43 @@ import {
...
+): number | any | undefined {
+  if (res === undefined) {
+    return undefined;
+  }
+  if (isGotoSignal(res)) {
+    const labelIdx = labels.get(res.target);
+    if (labelIdx !== undefined) {
+      return labelIdx + 1; // next statement
+    }
+    return res; // upwards
```
zefhemel: This seems redundant

client/space_lua/eval.ts
```diff
@@ -71,6 +74,43 @@ import {
...
+}
+
+function blockMetaOrThrow(
+  block: LuaBlock,
+  sf: LuaStackFrame,
+): ReturnType<typeof getBlockGotoMeta> {
+  try {
+    return getBlockGotoMeta(block);
+  } catch (e: any) {
+    if (e && typeof e === "object" && "astCtx" in e) {
```
zefhemel, line 107: I do wonder what is the scenario where this holds true and it's not already a LuaRuntimeErrror? Have you encountered such a case or trying to be safe here?

client/space_lua/eval.ts
```diff
@@ -1157,49 +1203,142 @@ export function evalStatement(
...
+      } else {
+        if (hasGotoFlag || hasLabelFlag) {
+          meta = blockMetaOrThrow(b, sf);
+          if (curFn) {
+            curFn.funcHasGotos = !!meta?.funcHasGotos;
+          }
+        }
+      }
+
+      if (!meta || !meta.funcHasGotos) {
```
zefhemel: To check, do we really need to track this, if this is validated in the parser can the parser not simply blow up before we even get to evaluation?

client/space_lua/goto_test.lua
```diff
@@ -0,0 +1,451 @@
...
+    error("Assertion failed: " .. tostring(a)
+      .. " != " .. tostring(b) .. " " .. (message or ""))
+  end
+end
+
+local function contains(hay, needle)
+  return string.find(tostring(hay), needle, 1, true) ~= nil
+end
+
+local function expect_error(fn, needle, msg)
```
zefhemel: I will soon declare camel casing to be The Way for Space Lua, but at least don't mix snake_case (here) with camelCase (assertEquals) in one file, it makes my eyes bleed ;)

client/space_lua/goto_test.lua
```diff
@@ -0,0 +1,451 @@
...
+  assert(y == nil)
+  y = true
+  if x == nil then
+    x = 1
+    goto l1
+  else
+    x = x + 1
+  end
+  assert(x == 2 and y == true)
+end
```
zefhemel, line 451: Very thorough, as usual 👏🏻

client/space_lua/labels.ts
```diff
@@ -0,0 +1,401 @@
...
+  LuaWhileStatement,
+} from "./ast.ts";
+import { asBlock } from "./ast_narrow.ts";
+
+type BlockGotoMeta = {
+  labels: Map<string, number>; // in this block
+};
+
+type FunctionMeta = {
+  blockMeta: WeakMap<LuaBlock, BlockGotoMeta>;
```
zefhemel, line 24: A comment on why a WeakMap here may be useful

client/space_lua/parse.ts
```diff
@@ -72,7 +73,89 @@ function parseBlock(t: ParseTree, ctx: ASTCtx): LuaBlock {
...
+  }
+
+  if (hasLabel) {
+    block.hasLabel = true;
+  }
+  if (hasGoto) {
+    block.hasGoto = true;
+  }
+  if (dup) {
+    block.dupLabelError = dup;
```
zefhemel, line 152: Would it not be simpler to simply throw an error at parse time here? I think we can consider duplicate labels a parse-time error. Then you don't have to maintain this information and check for it during eval time?

client/space_lua/goto_test.lua
mjf: `s/expect_error/expectError/g` ;)

client/space_lua/parse.ts
mjf: I tried but it broke tests that expect the error to surface when the function executes. Therefor I had to revert that change (at least for now). But the parse now computes `hasLabelHere` and preserves `dupLabelError`. :)

client/space_lua/labels.ts
mjf: Added.

client/space_lua/eval.ts
mjf: Scenario **might** occur when invalid forward goto or scope violation was detected.

client/space_lua/eval.ts
mjf: `BlockFlags` is just a local intersection type (`LuaBlock & { … }`) so eval can read those optional fields without changing the global AST type. `LuaBlock` already make all the flags optional, so this could have been removed. It has zero runtime cost, so it was kept.

mjf: I removed the `BlockFlags` indirection completely (apart from some minor code tweaks in parse). I think this is resolved now.

client/space_lua/eval.ts
mjf: This check enables fast path execution when a function or block has no goto and thus avoiding computing metadata unless really needed. If the parser validated this earlier (but the parser does not fully resolve gotos and labels as of now) we still would want this guard at least to preserve the performance. (Won't fix now.)

client/space_lua/eval.ts
mjf: Removed.

mjf:
> I think a similar mechanism could probably work to implement returns in a more efficient way than (often) throwing errors, this would likely result in a significant performance boost

I also assume this to be very helpful and will try in another round in the near future (noting it down in my SilverBullet instance for now). :-)

> To think about: would there be benefit in having a (serializable) parse cache with all analysis included in there for faster start-up time, to avoid parsing analyzing every time for the common case where code doesn't change between runs?

We could certainly:

- Cache the fully parsed AST (our custom `LuaBlock` tree) along with block flags (the `hasGoto`, `hasLabel`, `hasLabelHere`, `needsEnv` and  `dupLabelError`).
- Precompute goto/label analysis for each block, labels map (name to statement index) and function flag `funcHasGotos` etc.
- Serialize this to permanent storage and load if code block did not changed...

Some serialized parse and analysis cache can can significantly cut cold start time, **when code hasn’t changed**. And this is what may be tricky and I have no special idea about this. We can for example store by key by content hash of the original source (Space Lua code blocks, in fact?). Then we need to store the serialized object (in JSON?) in the IndexedDB and FS etc. and restore if needed. Is the effort worth it right now? The idea is indeed tempting, and **may** work as intended if implemented properly, but I do not see straightforward implementation path *atm.*

zefhemel:
Looks good to merge to me. Thank you!

As to a parse/analysis cache: yeah I wouldn't prioritize this right now, there may be little gain and a lot of complexity added. Just a thought I had. The return optimization will likely have more impact.

Thank you!

Issue #1588, opened 2025-10-03
Fix linked mentions multi-line behavior


zefhemel:
https://community.silverbullet.md/t/linked-mentions-display-multiple-lines-as-one/3322

zefhemel:
This logic has now been completely rewritten, should be better

PR #1688, opened 2025-11-14
Fix issue #1588: Multiline Linked Mentions

FoxtrotFaux:
[This issue](https://github.com/silverbulletmd/silverbullet/issues/1588) is due to that newline characters not being included in the multiline snippet. 

Simplified snippet logic and **temporarily** removed multiline snippets. Please see below.

This is what I have been using locally to make the Linked Mentions look more familiar as a user coming from Logseq.

I will readd multiline logic to this PR depending on what you think.

I feel that for most mentions, adding context lines arbitrarily makes the snippet visually noisy in most cases.

I have the following possibly unwelcome ideas to improve this. I am willing to add them to this PR or separate PRs if you are interested.

Style the context lines with a slightly darker text color or something similar to make the line with the link stand out more. Could hide the context lines behind some sort of caret. 

Include the preceding outline parent (see below) while skipping any lines in between and maybe adding an elipsis to indicate the skipped lines.

```
- Parent
  - Child
  - Child with [[link]]
  - Child
```

Include some succeeding outline children for cases where the link is used as a title for those outline children like below

```
- [[link]]
  - Child
  - Child
  - Child
```

### DIFF
--- a/plugs/index/snippet_extractor.ts
+++ b/plugs/index/snippet_extractor.ts
@@ -11,7 +11,7 @@ export function extractSnippetAroundIndex(
 
   // Find which line contains the index
   const textBeforeIndex = text.slice(0,index);
-  const linesBeforeIndex = textBeforeIndex.matchAll(new RegExp("/\n/","g")).toArray();
+  const linesBeforeIndex = textBeforeIndex.matchAll(new RegExp("\n","g")).toArray();
   targetLineIndex = linesBeforeIndex.length;
 
   if (targetLineIndex === -1) {


zefhemel:
Awesome, thank you for doing this. It's been bugging me forever but never enough to actually act 😆 

Small request is that you run `make check` on the repo, you'll see there's a typecheck error now (see failing tests in this PR also).

The improvements you suggest sound interesting, happy to have a look at them!

FoxtrotFaux:
Yup, this was very much thrown together. I used the wrong regex function as well. Can use match() instead of matchAll() since we're not using capture groups. Don't have to convert to an array that way as well. 

I was expecting you were going to make me reimplement the feature in some way like in my examples.

Will submit the fixes to pass when I get off work. I should be able to set up a new PR for the 2 multiline cases as well. 

I haven't looked at anything else, so not sure how I would go about changing the styling or adding a caret at the moment.

FoxtrotFaux:
Ok that should be good. Ignore what I said about match vs matchAll. I've been doing this on mobile lol

Will get on that other PR

Edit: no, that wasn't good. I will no longer be doing PRs at work dinners on mobile 2 cocktails in. desktop only from now on.

zefhemel:
Ok, I suppose I forgot to mention to also run `make test` to run unit tests, which now fail (because you changed the expected snippets). See build output.

PR #1689, opened 2025-11-15
Added search to the action buttons

zeus-web:
Especially on a mobile device it is a bit cumbersome to get to the search function the silverbullet site.  But search is essential for findeing the needed docs...

### DIFF
--- a/website/CONFIG.md
+++ b/website/CONFIG.md
@@ -49,6 +49,13 @@ config.set {
       end
     },
     {
+      icon = "search",
+      description = "Search",
+      run = function()
+        editor.invokeCommand("Search Space")
+      end
+    },
+    {
       icon = "terminal",
       description = "Run command",
       run = function()


zefhemel:
I agree, but actually think we should enable Silversearch instead. The built in search is quite bad. 

The Dockerfile.website shows how excalidraw is currently included. We can do the same for Silversearch. I can also do it you don't manage.

zeus-web:
I agree! Including silversearch would be even better, i use it here, too.

However, i had no idea about using it on the Silverbullet.md website. Would be great, if you could integrate it, along with a button '-)

Thanks!

zefhemel:
Ok, did this here: https://github.com/silverbulletmd/silverbullet/commit/79df63bd797dd2ebb691788c695e93f99da4f6b0

Thanks for the idea!

PR #1696, opened 2025-11-20
Follow the system theme automatically when dark mode toggle was not set before

LelouchHe:
This pr is to make the app follow the system theme change dynamically. When users haven't toggled dark mode before, the theme will be the same with the system automatically.

The current status is that users have to reload the app after changing the system theme. It might be fine for desktop usage but kind of uneasy on mobile.

https://github.com/user-attachments/assets/654d8de7-6440-4614-9cb6-ed4e52ec6ffd


LelouchHe:
there was some discussion [here](https://github.com/silverbulletmd/silverbullet/pull/1294) on clearing the "darkMode" setting. Once it's toggled, users can't reset it unless it's deleted from IndexedDB manually, which is quite difficult on mobile.

zefhemel:
Thanks!

lutzky:
> there was some discussion [here](https://github.com/silverbulletmd/silverbullet/pull/1294) on clearing the "darkMode" setting. Once it's toggled, users can't reset it unless it's deleted from IndexedDB manually, which is quite difficult on mobile.

This can be made easier using lua.

```lua
clientStore.delete("darkMode")
```

I added that to my `CONFIG`, but you could e.g. make a command of it.

PR #1709, opened 2025-11-26
Fix HEIC/HEIF image MIME type detection

### LINKED ISSUE #1671: Wrong Mime Type for HEIC images
HEIC images reported as application/octet-stream. 

The bug seems to be in server/util.go that in turns calls a go standard mime. 

Based on the go documentation https://cs.opensource.google/go/go/+/go1.25.4:src/mime/type.go;l=104 
The built-in table is small but on unix it is augmented by the local system's MIME-info database or mime.types file(s) if available under one or more of these names:

So probably it is enough to include a good table in the right location in the distribution docker file.

Sadly I am unable to create a pull request :-(

gitlon:
Fix HEIC/HEIF image MIME type detection

Closes #1671.

HEIC and HEIF image files were being reported as `application/octet-stream` instead of proper image MIME type.

To fix, I updated the existing function in `server/util.go`, to include explicit handling for `.heic` and `.heif` extensions.

I also changed this function to make case-insensitive comparisons; eg for the existing markdown detection `.MD` now works like `.md`.

I added new unit tests for this function in `server/util_test.go`.

### DIFF
--- a/server/util.go
+++ b/server/util.go
@@ -6,22 +6,23 @@ import (
 	"strings"
 )
 
+// Custom MIME type mappings for file extensions not in the system database
+var mimeTypeExtMapping = map[string]string{
+	".md":   "text/markdown",
+	".heic": "image/heic",
+	".heif": "image/heic",
+}
+
 // lookupContentTypeFromPath determines MIME type based on file extension
 func LookupContentTypeFromPath(path string) string {
-	ext := filepath.Ext(path)
+	ext := strings.ToLower(filepath.Ext(path))
 
-	// Handle specific file types that may not be in the system MIME database
-	// or need special handling
-	if strings.EqualFold(ext, ".md") {
-		return "text/markdown"
-	}
-	if strings.EqualFold(ext, ".heic") {
-		return "image/heic"
-	}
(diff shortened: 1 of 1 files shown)


server/util.go
```diff
@@ -3,14 +3,25 @@ package server
...
+	// or need special handling
+	if strings.EqualFold(ext, ".md") {
 		return "text/markdown"
 	}
+	if strings.EqualFold(ext, ".heic") {
+		return "image/heic"
+	}
+	if strings.EqualFold(ext, ".heif") {
+		return "image/heic"
+	}
```
zefhemel, line 23: I see where this is going in the future, so perhaps we use this opportunity to refactor this slightly and turn this into a (global) map?

Something along the lines of a global variable:

```go
var mimeTypeExtMapping = map[string]string {
	".md": "text/markdown",
	".heic": "image/heic",
	".heif": "image/heic",
}
```
Combine that with a strings.ToLower(ext) and a lookup. What do you think?

gitlon: Updated per the feedback.

zefhemel:
Awesome, thanks!

PR #1710, opened 2025-11-27
Fix document editors on safari

### LINKED ISSUE #1627: Document editors don't work on safari
I've made the switch in #1526 to modifying the iframe from the outside, which reduced a lot of weird api layers. This works fine in firefox and chrome, but in safari, shortly before the iframe is loaded, the `contentWindow` is somehow replaced and thus all our augmented data/functions are lost. Unsure what the best approach is on fixing this.

Props to Antares in the forum for reporting this.

MrMugame:
Fixes #1627 by basically reverting to the old model.

One thing that annoys me here is that the APIs between codeWidgets and document editors are very different (Which I'm kind of to blame for).

```js
// Document editors
globalThis.silverbullet.syscall(...)
globalThis.silverbullet.sendMessage(...) // For communicating saves etc., basically scoped events
globalThis.silverbullet.addEventListeners(...) // For receiving messages
// Syscall editor.sendMessage(), can be used to send message from the outside (This could just be done using events)

// Panel HTML + Code Widgets
globalThis.syscall(...)
loadJsByUrl(...) // I really don't think anyone is using this
updateHeight() // The height check is also performed for panels
// No way to send data from the outside (no events, etc.)
```

In an ideal world we should remove codeWidgets, update the panel API, remove `editor.sendMessage` and allow users to listen to events. This is probably impossible tho, as people don't really seem to update stuff, both the excalidraw plug and the tree view plug throw a bunch of deprecation warnings ....

### DIFF
--- a/client/components/panel.tsx
+++ b/client/components/panel.tsx
@@ -49,9 +49,6 @@ export function Panel({
         return;
       }
       switch (data.type) {
-        case "event":
-          editor.dispatchAppEvent(data.name, ...data.args);
-          break;
         case "syscall": {
           const { id, name, args } = data;
           editor.clientSystem.localSyscall(name, args).then(
--- a/client/components/panel_html.ts
+++ b/client/components/panel_html.ts
@@ -8,11 +8,11 @@ export const panelHtml = `<!DOCTYPE html>
 const pendingRequests = new Map();
 let syscallReqId = 0;
 
-self.syscall = async (name, ...args) => {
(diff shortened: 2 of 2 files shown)


zefhemel:
Seems to work as advertised. 

Regarding your comments, I agree that codeWidgets should be phased out. Problem is that I do not have a clear picture of who's using them (outside the core). I know excalidraw is one, here we can offer help (I have created PRs to third-party plugs before).

PR #1711, opened 2025-11-27
Improve links in frontmatter

tgorochowik:
Without this, only a single quoted link would be rendered properly inside frontmatter, i.e:
```yaml
a: "[[link]]"
b: "https://silverbullet.md"
c: "[[link]] asdas"          # wouldn't work
d: "[[link]] [[das]]"        # wouldn't work - would render a link to: link]] [[das
e: '[[link]]'                # wouldn't work
```

With this change, the following works as expected:
```yaml
a: "[[aaaa]]"
b: "[[aaa]a]]"
c: "[[aaaa]] sadasd"
d: "[[aaaa]] [[bbbb]]"
e: "[[aaaa]] @ [[bbbb]]"
f: [[bbbb]]                  # this will NOT be rendered as a link as it's a valid yaml array (in array)
g: "https://silverbullet.md"
h: "https://silverbullet.md https://github.com"
i: "asd ads https://silverbullet.md"
j: '[[asdas]]'
```

Note that the links are rendered only when the cursor is outside the whole frontmatter block.
I am not sure what is preferred - before this patch wiki links would be rendered when the cursor was outside the whole block, but regular urls would be rendered when the cursor was outside the url itself, so it was kind of inconsistent.

I am also attaching a few screenshots.

Before this change:

With this change, with the cursor outside of the frontmatter block:

With this change, with the cursor inside the frontmatter block:

Thougts?

### DIFF
--- a/client/codemirror/frontmatter.ts
+++ b/client/codemirror/frontmatter.ts
@@ -9,6 +9,7 @@ import {
 } from "./util.ts";
 import type { Client } from "../client.ts";
 import {
+  frontmatterQuotesRegex,
   frontmatterUrlRegex,
   frontmatterWikiLinkRegex,
 } from "../markdown_parser/constants.ts";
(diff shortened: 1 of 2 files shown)


zefhemel:
Yeah it's not entirely clear if this live preview should be enabled at a granular or more broader level. I played with your solution and I kind of like it. So let's go with this for now. Thank you!

PR #1712, opened 2025-11-28
feat: Add linkName attribute support for wiki link autocomplete

mweichert:
## Summary

- Add support for `linkName` frontmatter attribute that controls the alias text inserted during wiki link autocomplete
- Precedence order: `linkName` > `displayName` > page name
- Allows users to have separate link text from display name

## Use Case

For a page named "Randy Mario Poffo" with frontmatter:
```yaml
---
linkName: Randy Savage
displayName: Macho Man
---
```

Selecting this page from autocomplete inserts: `[[Randy Mario Poffo|Randy Savage]]`

This separates the concept of "how this page appears in links" (`linkName`) from "how this page is displayed in the UI" (`displayName`).

## Changes

- Modified `plugs/editor/complete.ts` to check for `linkName` before falling back to `displayName`
- Updated detail text in autocomplete to indicate which attribute is being used

## Test plan

- [ ] Create a page with `linkName` in frontmatter
- [ ] Verify autocomplete inserts `[[pagename|linkName]]`
- [ ] Create a page with only `displayName` - verify existing behavior unchanged
- [ ] Create a page with both - verify `linkName` takes precedence

### DIFF
--- a/plugs/editor/complete.ts
+++ b/plugs/editor/complete.ts
@@ -96,20 +96,23 @@ export async function pageComplete(completeEvent: CompleteEvent) {
 
       if (isWikilink) {
         // A [[wikilink]]
-        if (pageMeta.displayName) {
-          const decoratedName = namePrefix + pageMeta.displayName;
+        const linkAlias = pageMeta.linkName || pageMeta.displayName;
+        if (linkAlias) {
+          const decoratedName = namePrefix + linkAlias;
           let boost = new Date(pageMeta.lastModified).getTime();
           if (pageMeta._isAspiring) {
             boost = -Infinity;
           }
           completions.push({
-            label: pageMeta.displayName,
+            label: linkAlias,
             displayLabel: decoratedName,
             boost,
             apply: pageMeta.tag === "template"
               ? pageMeta.name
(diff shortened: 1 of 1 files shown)


mweichert:
Btw, absolutely love what you've created with Silverbullet! Keep up the great work!

zefhemel:
I understand what this does, but what is the scenario where you need this level of control? Is displayName not sufficient?

mweichert:
Sorry for the pause - travelling at the moment.

My particular use case is when I reference my wife, son, and close family members. I have People pages like `People/William Weichert`. When I view that page, the displayName is "William "Hayes" Weichert. Hayes is William's Nickname.

When I'm journalling, I like to write something like `Went for a walk with Hayes today`. So when I write `Went for a walk with [[` and then use the arrow keys to select which page, I want to select `People/William Weichert|Hayes`. But when I look at the `People/William Weichert` page, I want the `displayName` to be `William "Hayes" Weichert`.

I publish my people pages to external sources to use as contact information. So having the displayName as "Hayes" isn't ideal to my liking.

zefhemel:
Alright, fair enough. We should really have both of these documented, I don't think this is currently explained anywhere except in the code 😆

PR #1715, opened 2025-12-02
Space Lua: Add `<const>` attribute

mjf:
- First commit shows the general idea of implementing `<const>` attribute on top of current Space Lua codebase.
- The second one shows some more effort to move validation to parse time and allows for new attributes in the future.
- Also adds comprehensive test suite for the `<const>` attribute.


zefhemel:
Very nice. No notes.

PR #1717, opened 2025-12-03
Include sourcemaps in production builds

### LINKED ISSUE #1699: Please provide sourcemaps in production builds
Feature request: providing sourcemaps in production builds would make debugging of any JS issue much easier.

I'm trying to debug a small issue I'm having with Silverbullet (hosted using the Docker image), and only seeing `client.js` in the browser devtools isn't great. Providing sourcemaps would be very nice. The impact would be minimal as they are only loaded on-demand, and since the typical Silverbullet user is probably quite technical, I think it would make a lot of sense for a lot of users to have them available when needed.

Would it be possible to add that, please?

Schnouki:
This makes it easier to debug issues, but also to understand SilverBullet internals while developing new features in Space Lua.

Fixes #1699.

### DIFF
--- a/build_client.ts
+++ b/build_client.ts
@@ -70,7 +70,7 @@ async function buildCopyBundleAssets() {
     absWorkingDir: Deno.cwd(),
     bundle: true,
     treeShaking: true,
-    sourcemap: Deno.args[0] === "--production" ? undefined : "linked",
+    sourcemap: "linked",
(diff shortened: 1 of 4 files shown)


zefhemel:
I feared this would dramatically increase the `silverbullet` binary size, but it only seems to add ~1MB so that's ok. Thanks!

PR #1721, opened 2025-12-11
Fix missing x-proxy-headers- from non-lua fetch calls

justyns:
buildProxyHeaders was being used for calls coming from Lua, but not the fetch calls from a js plug.  So auth headers were getting silently dropped even though the request was still being proxied.

I also added `host.docker.internal` to the regex for urls that use http instead of https.   That way if ollama is running on the same machine as silverbullet, we can still use `http://host.docker.internal:11434/v1` as the url.

### DIFF
--- a/server/proxy.go
+++ b/server/proxy.go
@@ -13,7 +13,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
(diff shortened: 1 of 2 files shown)


zefhemel:
Thanks!

PR #1734, opened 2025-12-25
Fix auth initialization race condition

dkoosis:
## Summary

Fixes data race in auth initialization using `sync.Once` pattern.

## Motivation

- **Issue:** Classic double-checked locking race in `InitAuth()`
- **Problem:** Reads `JwtIssuer` outside mutex, writes inside → data race
- **Impact:** Multiple concurrent requests can trigger duplicate initialization
- **UX Impact:** None (correctness fix)

### types.go:53
- Added `authOnce sync.Once` field to SpaceConfig

### auth.go:160-177
- Replaced mutex-based lazy initialization with `sync.Once.Do()`
- Guarantees exactly-once execution under concurrent access
- Eliminates unsynchronized read of `JwtIssuer`

## Before (Race Condition)
```go
func (spaceConfig *SpaceConfig) InitAuth() error {
    if spaceConfig.JwtIssuer == nil {  // ← Unsynchronized read
        spaceConfig.authMutex.Lock()
        defer spaceConfig.authMutex.Unlock()
        // ... initialization
    }
    return nil
}
```

## After (Race-Free)
```go
func (spaceConfig *SpaceConfig) InitAuth() error {
    var initErr error
    spaceConfig.authOnce.Do(func() {
        // ... initialization runs exactly once
    })
    return initErr
}
```

## Testing

- All auth tests pass (20/20)
- Overall coverage: 68.5% (maintained)
- No behavioral changes for valid usage

## Context

Part of focused hardening series. Each PR is independent and cherry-pickable.

### DIFF
--- a/server/auth.go
+++ b/server/auth.go
@@ -158,15 +158,12 @@ func addAuthEndpoints(r chi.Router, config *ServerConfig) {
 }
 
 func (spaceConfig *SpaceConfig) InitAuth() error {
-	if spaceConfig.JwtIssuer == nil {
-		spaceConfig.authMutex.Lock()
-		defer spaceConfig.authMutex.Unlock()
-
-		var err error
+	var initErr error
+	spaceConfig.authOnce.Do(func() {
 		// Need to do some initialization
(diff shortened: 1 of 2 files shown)


dkoosis:
Closing in favor of #1743 which uses the cleaner sync.Once pattern for auth initialization.

PR #1741, opened 2025-12-26
Fix Windows self-upgrade by renaming instead of removing binary

### LINKED ISSUE #1628: Upgrade failed with "text file busy"
I have the `silverbullet` binary in a directory `sb`. I try to self-upgrade it (which used to work in the past):

```console
mihaimaruseac@ankh:~/sb$ ./silverbullet upgrade-edge
Now going to attempt an upgrade...
Install dir: ~/sb
Downloading from https://github.com/silverbulletmd/silverbullet/releases/download/edge/silverbullet-server-linux-x86_64.zip
Now going to replace the existing silverbullet binary in ~/sb
2025/10/14 11:25:04 failed to extract zip: open ~/sb/silverbullet: text file busy
```

According to https://stackoverflow.com/a/52427512 this looks like a missing `unlink`. If I get some time, I could dig in to place it.

dkoosis:
## Summary

Fixes #1628

The previous approach used `os.Remove()` which fails on Windows because running executables cannot be deleted (they are locked by the OS).

## Changes

This PR changes the upgrade process to use `os.Rename()` instead of `os.Remove()`. This approach:

- ✅ Works on both Unix and Windows
- ✅ Creates a new inode for the new binary  
- ✅ Preserves the old binary as a backup (`.old` file) in case upgrade fails
- ✅ Matches the pattern used by other self-update tools

## Testing

Added comprehensive tests that verify:
- Existing files are properly renamed to `.old`
- New files are extracted correctly
- The approach works when no existing file is present

## Notes

While #1628 was closed after PR #1642, that fix only addressed Linux. On Windows, `os.Remove()` still fails for running executables. The `os.Rename()` approach is the standard cross-platform pattern used by projects like [go-update](https://github.com/inconshreveable/go-update) and similar self-update libraries.

Tried on Windows 11 Home Edition.

### DIFF
--- a/server/cmd/upgrade.go
+++ b/server/cmd/upgrade.go
@@ -144,10 +144,14 @@ func extractZip(src, dest string) error {
 			return err
 		}
 
-		// First, attempt to remove the file to prevent "text file busy" error
-		err = os.Remove(path)
+		// Rename existing file to .old instead of removing it
+		// This works on Windows where running executables cannot be deleted
+		oldPath := path + ".old"
+		err = os.Rename(path, oldPath)
 		if err != nil && !errors.Is(err, fs.ErrNotExist) {
-			return err
+			// If rename fails for reasons other than file not existing, continue anyway
+			// The new file will be created with a new inode
+			log.Printf("Warning: could not rename existing file: %v\n", err)
 		}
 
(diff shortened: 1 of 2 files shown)


dkoosis:

PR #1745, opened 2025-12-26
Document verbose error messages as intentional design choice

dkoosis:
## Summary

Documents why filesystem errors are exposed to clients with full detail, explaining this is an intentional design choice for a personal tool.

## Context
Note: when using static analysis, I found a number of security-related no-no's.
On closer inspection I think these are OK for a personal tool, but I want to note my rationale to make sure I'm in harmonly with the Silverbullet way of thinking and @zefhemel's intention.

In enterprise applications, exposing detailed error messages is considered a security issue:
- Can leak system internals (paths, permissions, etc.)
- Generic errors ("An error occurred") are safer
- Detailed errors go to server logs for operator review

However, SilverBullet has a different threat model.

## Personal Tool Philosophy

SilverBullet is a **personal tool** where the user is both the operator and the end user:

**Benefits of verbose errors:**
- ✅ User immediately knows why an operation failed
- ✅ No need to dig through server logs for basic issues
- ✅ "Permission denied" vs "Disk full" vs "File locked" are actionable
- ✅ Reduces friction in day-to-day use

**Example scenario:**  
A note fails to save. Current behavior shows "ENOSPC: no space left on device". Generic error would just say "Write failed" and force the user to check logs, wasting time when they own the entire stack.

## Changes

Added comments explaining:
1. Why full error messages are intentionally exposed
2. The personal tool context that makes this the right choice
3. That enterprise deployments should wrap errors generically

**No behavioral changes** - this is pure documentation.

## Code

```go
// Expose full error message to client for debuggability
// Design choice: SilverBullet is a personal tool where the owner
// benefits from detailed errors (e.g., "permission denied", "disk full")
// Enterprise deployments should wrap errors generically to avoid leaking internals
http.Error(w, err.Error(), http.StatusInternalServerError)
```

## Philosophy

This PR acknowledges the security tradeoff while explaining why **transparency beats security-through-obscurity** for personal tools where the user owns the entire stack.

The comment serves as documentation for:
- Future contributors who might flag this as a "security issue"
- Users deploying in enterprise contexts (guidance to wrap errors)
- Maintainers making design decisions about error handling

-- dk & claude

### DIFF
--- a/server/fs.go
+++ b/server/fs.go
@@ -62,6 +62,10 @@ func handleFsGet(w http.ResponseWriter, r *http.Request) {
 			if err == ErrNotFound {
 				http.NotFound(w, r)
 			} else {
+				// Expose full error message to client for debuggability
+				// Design choice: SilverBullet is a personal tool where the owner
+				// benefits from detailed errors (e.g., "permission denied", "disk full")
+				// Enterprise deployments should wrap errors generically to avoid leaking internals
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 			}
 			return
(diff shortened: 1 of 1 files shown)


dkoosis:

PR #1750, opened 2025-12-26
Fix HTTP 500 errors locking client outside offline mode

### LINKED ISSUE #1724: HTTP error 500 locks client outside offline mode
If the remote server responds with HTTP 500 to the `.config` request, the [offline‑mode failsafes](https://github.com/silverbulletmd/silverbullet/blob/b9bbb32051283c7501efcdf8be88bc58b85cae31/client/boot.ts#L283) are not triggered, and the client [caches the error page’s body](https://github.com/silverbulletmd/silverbullet/blob/b9bbb32051283c7501efcdf8be88bc58b85cae31/client/boot.ts#L313) (for example, an Nginx 500 HTML page). 

During boot, that cached `.config` is then [parsed as JSON](https://github.com/silverbulletmd/silverbullet/blob/b9bbb32051283c7501efcdf8be88bc58b85cae31/client/boot.ts#L45), which fails when the content is unrelated HTML. As a result, the client cannot boot and shows “Could not process config and no cached copy, please connect to the Internet” on a blank page. 

Would it make sense for Silverbullet to [enter offline mode for a broader set of HTTP status codes than the ones currently listed](https://github.com/silverbulletmd/silverbullet/blob/b9bbb32051283c7501efcdf8be88bc58b85cae31/plug-api/constants.ts#L24-L30)? Maybe any 5xx error?

dkoosis:
## Problem

Fixes #1724

When the server returns HTTP 500 (or other 5xx errors), the client was caching the error HTML page (e.g., Nginx error pages) and attempting to parse it as JSON during boot. This caused a catastrophic lockout with the error message:

```
Could not process config and no cached copy, please connect to the Internet
[Client] Failed to process config Unexpected token '<', "<html> <h"... is not valid JSON
```

The root cause is that only specific error codes (502, 503, 504, 530) triggered offline mode, while 500 and other 5xx errors were treated as successful responses and cached.

## Solution

Add all standard 5xx server error codes to `offlineStatusCodes`:
- 500: Internal Server Error ⭐ (the main issue from #1724)
- 501: Not Implemented
- 505-511: Other server errors

When the server returns any 5xx error, the client now:
1. Detects it as an offline condition
2. Falls back to cached data
3. Does NOT cache the error HTML response

## Testing

**Before:** Server returning 500 → client caches error HTML → JSON parse fails → lockout

**After:** Server returning 500 → client detects offline → uses cached config → continues working

## Impact

- Prevents catastrophic client lockout from server errors
- Enables true offline-first behavior for all server failures
- User can continue working with cached data during server maintenance/errors

### DIFF
--- a/plug-api/constants.ts
+++ b/plug-api/constants.ts
@@ -20,11 +20,23 @@ export const pingInterval: number = 5000;
  *
  * This enables SilverBullet to work in a true "offline-first" manner, falling back to cached
  * content when the backend is unavailable through no fault of the user's network connection.
+ *
+ * All 5xx server errors are included to prevent the client from caching error HTML pages
+ * (e.g., Nginx 500 error pages) which would prevent the client from booting in offline mode.
  */
 export const offlineStatusCodes = {
+  500: "Internal Server Error", // Server encountered an unexpected condition
+  501: "Not Implemented", // Server does not support the functionality required
   502: "Bad Gateway", // Proxy server received invalid response from upstream server
   503: "Service Unavailable", // Server is temporarily unable to handle the request
   504: "Gateway Timeout", // Proxy server did not receive a timely response from upstream server
+  505: "HTTP Version Not Supported", // Server does not support the HTTP version
+  506: "Variant Also Negotiates", // Server has an internal configuration error
+  507: "Insufficient Storage", // Server is unable to store the representation
+  508: "Loop Detected", // Server detected an infinite loop while processing
+  509: "Bandwidth Limit Exceeded", // Server bandwidth limit has been exceeded
+  510: "Not Extended", // Further extensions to the request are required
+  511: "Network Authentication Required", // Client needs to authenticate to gain network access
 
   530: "Unable to resolve origin hostname", // Served when cloudflared is down on the host
 } as const;


zefhemel:
While to some degree I appreciate the effort, you have to see that you're effectively denial-of-servicing my project here. 

While your faith in the code quality of an AI agent may be high, mine is not, based on lived experience. This means that I have to thoroughly evaluate every single line. If you create 17 PRs in a single day with thousands of lines of code added (sure, many are tests, but those deserve scrutiny too) this means I will likely have to spend multiple days reviewing them. I doubt you spent that much time generating them, so this is not a fair "fight". Therefore, I can do no different than simply close them. I'm sorry, I don't have this type of time. Also, I'm not super convinced many of the problems these claim fix are necessary real (again, based on experience using these type of AI tools).

dkoosis:
Hello Zef:

I completely understand and I'm very sorry for spamming the project.

FWIW - I don't trust AI generated code blindly either; I was QA'ing the Go
server for my own deployment and found several issues that looked clear, so
I directed AI, reviewed, tested. I should have been much more selective.
Then I split up everything into a dozen PRs thinking it would be easier for
you to cherry-pick what would be helpful -- but neglecting to think about
all the added review work.

A few of the PRs are very small, focussed, fixes that may be of value
without consuming much of your time:

PR #1750 is 12 lines of code that addresses issue #1724 ***@***.***) -
HTTP 500 errors were locking clients out of offline mode. This adds 5xx
codes to the offline status array. I reproduced the lockout and verified it
works in my environment.

PR #1731 is 15 lines and removes a redundant syscall on every file save (I
measured 3-15% faster on macos environment).

Again, forgive me for for creating noise when I meant to be contributing
value.

I appreciate the work you've done.

/dk
==
David Koosis
646.481.4141
fractional CIO/CTO for mission driven organizations | *schedule a call
<https://calendar.app.google/rnZB34XRugxK9uMW8>*

zefhemel:
Cherry picked the commit with the additional status codes. Thanks

zefhemel:
Also looked at #1731 but it introduces a regression by setting the creation time wrong (which was the reason for the syscall).

Issue #1729, opened 2025-12-24
_CTX async loading and no re-evaluation


kesor:
I wanted to use functions to place widgets on a page, where the functions read from the frontmatter some data to show. For example,

```lua
function dateUtils.isReady(ctx)
  return ctx and ctx.currentPage and ctx.currentPage.date
end

function dateUtils.time(ctx)
  if not dateUtils.isReady(ctx) then return "" end
  if (not ctx.currentPage) or (not ctx.currentPage.date) then
    return "ERR"
  end
  local year, month, day = string.match(ctx.currentPage.date, "^(%d+)%-(%d+)%-(%d+)")
  if not year then return "ERR" end
  return os.time{year=tonumber(year), month=tonumber(month), day=tonumber(day)}
end
```

The problem is that when a page is shown, the context is 'nil'. While some async process is loading the data in. And unless I alt-click the widget, the function is not going to be reevaluated with the "new" data. So when visiting the page the widget is in a broken state until clicking to refresh it.

kesor:
This still happens in v2.4.1

Also started getting this error --

```
[Client] Failed to parse frontmatter: TypeError: globalThis.syscall is not a function
    at Ke (document_editor_js.ts:1:16)
    at Object.hnt (document_editor_js.ts:1:16)
    at dQ (boot.ts:351:1)
    at Ad.loadPage (boot.ts:351:1)
    at async boot.ts:351:1
    at async space.ts:144:48
```

This is how it looks like when opening a new page from template, or using SPA navigation to reach a page that uses `_CTX` as a param to its widgets to read the `date` from the frontmatter:

<img width="703" height="498" alt="Image" src="https://github.com/user-attachments/assets/2a31ca50-c9e0-4798-8f47-31f12935eee3" />


And this is how it should look like. I had to Alt-click each widget to "refresh" it and make it render:

<img width="1178" height="386" alt="Image" src="https://github.com/user-attachments/assets/374d8bc0-0266-4dc4-83d0-11715b4b92b4" />

For example, the title is a simple:

```
${dateUtils.titleWithDay(_CTX)}
```

And the implementation is also fairly simple, the `isReady` was obviously an attempt at a fix, that failed:

```lua
dateUtils = dateUtils or {}

function dateUtils.isReady(ctx)
  if ctx and ctx.currentPage and ctx.currentPage.date
    then return true
  end
  return false
end

function dateUtils.time(ctx)
[...]

PR #1759, opened 2026-01-03
Fix widgets not re-rendering when page metadata loads

kesor:
Trigger editor re-render after page metadata is loaded to ensure widgets re-evaluate with the correct context. This fixes issue [#1729](https://github.com/silverbulletmd/silverbullet/issues/1729) where widgets would continue showing broken state on initial page load because `ctx.currentPage` would be `nil` until the metadata finished loading asynchronously.

The fix dispatches an empty transaction after updating the page metadata, which triggers all decorator state fields (including widgets) to re-render with the updated metadata.

### DIFF
--- a/client/client.ts
+++ b/client/client.ts
@@ -543,6 +543,9 @@ export class Client {
                     type: "update-current-page-meta",
                     meta: enrichedMeta,
                   });
+
+                  // Trigger editor re-render to update Lua widgets with the new metadata
+                  this.editorView.dispatch({});
                 }
               })
               .catch((e) => {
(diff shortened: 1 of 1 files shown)


zefhemel:
Nice, thanks!

kesor:
@zefhemel This actually didn't resolve the issue. I went diving into it more, it is difficult to test with all the caching involved at various levels.

The issue occurs when using SPA navigation. Moving to a new page, loads the minimal metadata, and later asynchronously enriches the metadata with frontmatter. The widgets are already rendered, and are not re-rendered when enrichment happens. When visiting the page directly (not via SPA) it does refresh the widgets. When turning on/off 'vim mode', it also refreshes widgets.

Often when coming back to a page (using SPA navigation) the cached widgets are shown (not all of them) and they render differently than normal widgets, for example they include a newline while the regular ones don't.

I spend a couple of days digging into this, but couldn't resolve all of these issues :(

PR #1776, opened 2026-01-14
fix: add null checks to prevent 'cannot read children' crash

loweaxerium:
## Problem
When rendering Lua widget output containing tasks inside markdown tables, SilverBullet crashes with:
```
TypeError: Cannot read properties of undefined (reading 'children')
```

The error occurs in the `preprocess()` function in `markdown_render.ts` where `node.parent.children` is accessed without first checking if `node.parent` exists.

## Cause
This was likely introduced during the inline content refactor (commit fb3dd51b). When certain markdown nodes are processed in isolation (e.g., within table cells or widget outputs), they may not have a parent pointer set.

## Solution
Added defensive null checks:
- **markdown_render.ts**: Guard `node.parent?.children` before accessing in `preprocess()`
- **inline.ts**: Guard `n.children` and `n.parent?.from` in task rewriting
- **item.ts**: Use optional chaining for task state extraction

## Testing
Tested locally with a Lua widget that renders tasks inside a markdown table. The crash no longer occurs and tasks render correctly.

### DIFF
--- a/client/markdown_renderer/inline.ts
+++ b/client/markdown_renderer/inline.ts
@@ -122,8 +122,9 @@ export async function expandMarkdown(
     } else if (n.type === "Task" && options.rewriteTasks !== false) {
       // Add a task reference to this based on the current page name if there's not one already
       const existingLink = findNodeOfType(n, "WikiLink");
-      if (!existingLink) {
-        n.children!.splice(1, 0, {
+      // Only rewrite if we have children, parent, and parent position
+      if (!existingLink && n.children && n.parent?.from !== undefined) {
+        n.children.splice(1, 0, {
           "text": " ",
         }, {
           "type": "WikiLink",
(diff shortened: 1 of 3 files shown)


zefhemel:
While I appreciate the attempt, this actually just covers up a more real issue: a naive "optimization" in addParentPointers, which I now removed: https://github.com/silverbulletmd/silverbullet/commit/062c1ea44d406d5c48152d479f15c8a830de70f8#diff-ef0ad8048a0a6a7317c497a25075cd0428a153f7f1cc6ef44ee91a8b08e3e45c

This more properly should resolve this issue.

**Side note:** I suspect you used some AI coding agent to help you. This type of fix is very typical for an AI agent: it doesn't _really_ understand a code base and its fixes tend to be superficial, which just buries actual issues, leading to more technical debt.

loweaxerium:
Thanks a lot, Zef! I couldn’t find a better way to report this issue since I’m not familiar with the codebase, so I tried a quick-and-dirty approach to get my space running locally again. I did get some helpful pointers—and a slightly silly patch—from an agent, and I thought it might at least point you in the right direction. You might also consider adding issue and PR templates to improve guidance in these cases.

PR #1782, opened 2026-01-21
 fix(remove-completed): correctly remove completed tasks in nested li…

### LINKED ISSUE #1722: "Remove Completed" command leaves empty space in nested lists
Hi folks!

I use the "Remove Completed" command to automatically delete my completed tasks, and found (what I assume) is a bug when handling tasks in nested lists:

https://github.com/user-attachments/assets/3782845d-00b2-486d-bff1-756a7c540212

Single-level lists of tasks will correctly remove the whole line with the task, but nested lists will leave an empty space in-between tasks. This persists after a reload, so I assume it's not just a UI issue.

dmalkinski:
…sts (fixes #1722)

Description: This changes removeCompletedTasksCommand to stop mutating the Markdown AST in-place with brittle index assumptions. Instead it:

Collects completed Task list items first.
Removes items bottom-up (by offset) to avoid invalidating parent indices while iterating. Removes adjacent newline runs robustly.
Prunes now-empty List / ListItem containers by walking parents after removal. Uses findNodeOfType to locate TaskState instead of numeric child indexing. Re-runs addParentPointers after modifications and renders the cleaned tree. This fixes leftover blank lines when removing completed items from nested lists (issue #1722) and makes the remove routine more robust with a minimal surface-area change.Add missing newline at end of file


dmalkinski:
yeah nah

PR #1791, opened 2026-01-27
Space Lua: Add `<close>` attribute

mjf:
Add `<close>` attribute along with extensive test suite.


client/space_lua/eval.ts
```diff
@@ -334,47 +337,8 @@ export function evalExpression(
...
-                throw new LuaRuntimeError(
-                  `Unknown unary operator ${u.operator}`,
-                  sf.withCtx(u.ctx),
-                );
-              }
-            }
-          });
-        } else {
+
+        const apply = (value: LuaValue) => {
```
zefhemel, line 341: Nice reduction in duplication 👍🏻

client/space_lua/eval.ts
```diff
@@ -1109,9 +1036,206 @@ function evalExpressions(
...
+    idx: number,
+  ): void | LuaValue[] | Promise<void | LuaValue[]> => {
+    for (let i = idx; i < stmts.length; i++) {
+      const result = evalStatement(
+        stmts[i],
+        execEnv,
+        sf,
+        returnOnReturn,
+      );
+      if (isPromise(result)) {
```
zefhemel, line 1065: As a side note: you may want to have a look at https://www.typescriptlang.org/docs/handbook/advanced-types.html it's possible to change `isPromise` to guarantee that its argument is a Promise if successful, in which case you don't have to do the `as Promise<any>` magic below. I think.

PR #1793, opened 2026-01-28
Space Lua: Align arithmetical error messages with Lua 5.4

mjf:
Provides:
- important fixes,
- error propagation for coercion,
- proper Lua 5.4 errors for arithmetics and coercion,
- new negative tests.

### DIFF
--- a/client/space_lua/numeric.ts
+++ b/client/space_lua/numeric.ts
@@ -26,11 +26,9 @@ export type OpHints = {
 };
 
 // Marker error used to let the evaluator decide the final message.
-export class LuaStringCoercionError extends Error {
-  constructor() {
-    super("LuaStringCoercionError");
-  }
-}
+export const luaStringCoercionError: Error = new Error(
+  "LuaStringCoercionError",
+);
 
 export function coerceNumeric(
   val: unknown,
@@ -84,7 +82,7 @@ export function coerceNumeric(
     const det = luaToNumberDetailed(val);
     if (!det) {
       // Let evaluator produce the final error message
-      throw new LuaStringCoercionError();
+      throw luaStringCoercionError;
     }
 
     const n = det.value;
--- a/client/space_lua/eval.ts
+++ b/client/space_lua/eval.ts
@@ -47,7 +47,7 @@ import {
   boxZero,
   coerceNumeric,
   coerceNumericPair,
-  LuaStringCoercionError,
+  luaStringCoercionError,
   type OpHints,
 } from "./numeric.ts";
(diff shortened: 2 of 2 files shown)


client/space_lua/numeric.ts
```diff
@@ -24,6 +25,13 @@ export type OpHints = {
   rightKind?: NumericType;
 };
 
+// Marker error used to let the evaluator decide the final message.
+export class LuaStringCoercionError extends Error {
```
zefhemel: While I'm not married to this idea, in other places I've done this (plug-api/constants.ts):

```
export const offlineError: Error = new Error("Offline");
export const notFoundError: Error = new Error("Not found");
export const notAuthenticatedError: Error = new Error("Unauthenticated");
```

Then I just throw the constant and can do an `===` check rather than instanceof. For consistency maybe adopt the same approach, except in cases where the error message is configurable?

client/space_lua/runtime.ts
```diff
@@ -64,6 +64,39 @@ type LuaThreadState = {
   closeStack?: LuaCloseEntry[];
 };
 
+export function luaTypeName(val: any): LuaType {
+  const t = luaTypeOf(val);
```
zefhemel, line 68: doesn't this always return a (promise of) a string, why replicate this logic in this function?

client/space_lua/tonumber_test.lua
```diff
@@ -5,6 +5,24 @@ local function assert_eq(a, b, message)
     end
 end
 
+local function assert_throws(msg_substr, fn)
```
zefhemel: Note to self (or you): let's use camelCase for all functions, also in tests. Seeing a mix of snake_case and camelCase freaks me out. I know I started this myself.

mjf: Fixed.

client/space_lua/runtime.ts
mjf: `luaTypeName()` is a sync helper intended for hot paths / error formatting where we must not return a `Promise`, thus **the duplication is intentional**. It is basically: `luaTypeOf` if sync; otherwise best-effort sync classification. If we removed it and used `luaTypeOf()` directly, we would have to make a bunch of code async just to format an error message. Won't fix?

client/space_lua/numeric.ts
mjf: Ok, so you incline to treat certain errors as *sentinel* values, not as "real" exception types. Ok, fixed.

PR #1794, opened 2026-01-28
Space Lua: Refactor control flow signals

mjf:


client/space_lua/eval.ts
```diff
@@ -1482,33 +1486,20 @@ export function evalStatement(
...
-              }
-            } else if (isGotoSignal(r)) {
-              return r;
-            } else if (r !== undefined) {
-              return r;
-            }
-          } catch (e: any) {
-            if (e instanceof LuaBreak) {
+          const r = evalStatement(w.block, env, sf, returnOnReturn);
+          const res = isPromise(r) ? await r : r;
```
zefhemel, line 1497: Fun fact (I think) which may or may not have performance impact, you can also `await` a non-promise, which will be a no-op. I'm pretty sure. If that's the case, this can just become `const res = await r;`

client/space_lua/eval.ts
```diff
@@ -1525,102 +1516,65 @@ export function evalStatement(
...
                 if (r !== undefined) {
+                  if (isBreakSignal(r)) {
+                    return;
+                  }
                   return r;
                 }
                 return runAsync();
               }
             } catch (e: any) {
-              if (e instanceof LuaBreak) {
```
zefhemel, line 1551: I think I added this to fix this: https://community.silverbullet.md/t/break-is-broken/3739 you're deliberate in removing it again, right?

client/space_lua/eval.ts
mjf: Sole `await r` behavior is indeed equivalent (as long as you are already in an async function and `r` is intended to be `T | Promise<T>`). But the sole `await r` will always introduce an async boundary (it resumes on a later task) even when `r` is not a promise! The ternary version used avoids that boundary in the sync case. Also performance could **slightly** worsen in *hot loops* because async suspension is forced even for sync values which in tight interpreter loops (like this one Lua evaluator) **CAN** matter.

client/space_lua/eval.ts
mjf: It’s ok provided we have fully removed the exception-based control flow mechanism (`LuaBreak`/`LuaReturn`) from the execution path. All break/return propagation is now done via the control signals. So I think the assumption holds and we can keep it. Also, the regression test introduced in cb0c74146d5b7934188c58d893e3127f125baf3f passes ok. I see no issue, thus removing the unused `LuaBreak`/`LuaReturn` in the next commit.

mjf: Removed `LuaBreak`/`LuaReturn`.

client/space_lua/runtime.ts
```diff
@@ -1009,16 +1026,6 @@ export function luaTypeOf(val: any): LuaType | Promise<LuaType> {
   }
 }
 
-// Both `break` and `return` are implemented by exception throwing
-export class LuaBreak extends Error {
-}
-
-export class LuaReturn extends Error {
```
zefhemel, line 1016: You are making me very happy

zefhemel:
Love this. Thanks so much.

Issue #1795, opened 2026-01-28
Configure remember me session duration


strboul:
Administrators should be able to set a session duration that fits their security requirements, and users should know how long they'll stay logged in. From https://silverbullet.md/Authentication

PR #1796, opened 2026-01-28
Configure remember me option

strboul:
Solves #1795

- New env var: `SB_REMEMBER_ME_HOURS`
- Login page shows: "Remember me (X days)"
- Also expiring JWT based on the remember me duration instead of no expiry


zefhemel:
This is great, one request: could you add this new env variable to https://github.com/silverbulletmd/silverbullet/blob/main/website/Install/Configuration.md as well as part of this?

PR #1803, opened 2026-01-30
Space Lua: Improve numeric types semantics

mjf:
Add numeric result normalization for arithmetic operations:

* Integer operations preserve int type, collapse -0 to +0
* Float operations preserve float type, maintain -0.0
* Division always produces float results

Fix for-loop variable type tracking:

* Loop variable type determined by start/step parameters
* Integer loops produce integer variables
* Float/mixed loops produce float-tagged variables

Implement table key normalization:

* Integer-valued floats (1.0, 2.0) normalize to integer keys
* All zero variants (0, 0.0, -0.0) map to same key
* Non-integer floats (1.5) remain distinct keys

Add tests for:

* Arithmetic type preservation across operators
* For-loop type modes (int/float/mixed)
* Table key equivalence and normalization
* Edge cases with zero, -0.0, and special values
* Standard library `type` and `math.type`

### DIFF
--- a/client/space_lua/eval.ts
+++ b/client/space_lua/eval.ts
@@ -926,12 +926,11 @@ const operatorsMetaMethods: Record<string, {
   "-": {
     metaMethod: "__sub",
     nativeImplementation: (a, b, _ctx, _sf, hints) => {
-      const { ax, bx, bothInt, aZeroKind: _aZeroKind, bZeroKind } =
-        coerceNumericPair(
-          a,
-          b,
-          hints,
-        );
+      const { ax, bx, bothInt, bZeroKind } = coerceNumericPair(
+        a,
+        b,
+        hints,
+      );
       const result = ax - bx;
 
       if (result === 0 && isNegativeZero(result) && !bothInt) {
@@ -1872,6 +1871,16 @@ export function evalStatement(
         return i;
       };
 
+      // Extract the common iteration logic
+      const executeIteration = (
+        i: number,
+        mode: "int" | "float",
+      ): void | ControlSignal | Promise<void | ControlSignal> => {
+        const localEnv = new LuaEnv(env);
+        localEnv.setLocal(fr.name, wrapLoopVar(i, mode));
+        return evalStatement(fr.block, localEnv, sf, returnOnReturn);
+      };
+
       const runAsync = async (
         _start: any,
         end: any,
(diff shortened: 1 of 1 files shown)


client/space_lua/eval.ts
```diff
@@ -855,52 +920,52 @@ const operatorsMetaMethods: Record<string, {
...
-      return r;
+      return normalizeArithmeticResult(ax + bx, bothInt);
     },
   },
   "-": {
     metaMethod: "__sub",
     nativeImplementation: (a, b, _ctx, _sf, hints) => {
-      const { ax, bx, bothInt } = coerceNumericPair(a, b, hints);
-      const r = ax - bx;
+      const { ax, bx, bothInt, aZeroKind: _aZeroKind, bZeroKind } =
```
zefhemel: FYI: I think you can just leave out any keys you don't care about (so `aZeroKind` in this case)

client/space_lua/eval.ts
```diff
@@ -1775,21 +1843,74 @@ export function evalStatement(
...
+            const r = evalStatement(fr.block, localEnv, sf, returnOnReturn);
+            const res = isPromise(r) ? await r : r;
+            if (res !== undefined) {
+              if (isBreakSignal(res)) {
+                break;
+              }
+              return res;
+            }
+          }
+        } else {
```
zefhemel: The `if` and `else` bodies are 99% the same (except for the second `for` condition is that right?), can this not be merged to reduce code duplication?

`std0 > 0 ? i  <= e0 : i >= e0` or something similar?

client/space_lua/eval.ts
```diff
@@ -1799,50 +1920,82 @@ export function evalStatement(
...
               }
-              return runAsync(i + step, end, step);
-            });
-          } else if (r !== undefined) {
-            if (isBreakSignal(r)) {
-              break;
+              return r;
+            }
+          }
+        } else {
```
zefhemel: Same story on code duplication as above?

client/space_lua/eval.ts
mjf: Removed/fixed.

client/space_lua/eval.ts
mjf: Done.

client/space_lua/eval.ts
mjf: Yeah. Deduplicated.

PR #1805, opened 2026-01-31
Render attribute decorations only when the cursor is not in range

LogeshG5:
To ensure easy editing, attributes are no longer decorated when the cursor is within range. Now that attributes are decorated for user styling, this helps.


redstreet:
This is an awesome change. Love the behavior, and it opens up a whole bunch of nice customizations, thank you!

Found a bug when using this with vim, filed #1814.

PR #1806, opened 2026-01-31
Fix docker image helthcheck

lalyos:
With a compose.yaml:
```
  silver:
    image: zefhemel/silverbullet
    environment:
    - SB_PORT=8080
    - SB_HOSTNAME=0.0.0.0
    ports:
    - 8080
```
it will show the container unhealthy as it check on the port 3000.
The healthchek was already using the `SB_URL_PREFIX` but `SB_PORT` was missing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ VOLUME /space
 
 RUN apk add --no-cache git curl bash tini
 
-HEALTHCHECK CMD curl --fail http://localhost:3000$SB_URL_PREFIX/.ping || exit 1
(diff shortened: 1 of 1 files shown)


MrMugame:
This breaks if you don't define SB_PORT (and then traefik won't detect you container hehe). I'm no bash expert, but something like `${SB_PORT:-3000}` would probably fix this issue @zefhemel

zefhemel:
Should be addressed with https://github.com/silverbulletmd/silverbullet/pull/1812

PR #1812, opened 2026-02-02
fix: undefined SB_PORT in healthcheck

lalyos:
fixes https://github.com/silverbulletmd/silverbullet/pull/1806#issuecomment-3836993483


zefhemel:
Thanks for the quick fix

lalyos:
finally github notification was useful

PR #1819, opened 2026-02-05
Panels - strings to elements

onespaceman:
This does work, but is a breaking change. It returns a shadow dom instead of an iframe, allowing to pass elements with their own attached methods/onclicks.
I tried to make it backwards compatible but wasn't able to.
1. I could re-implemented the events and listeners for syscalls like in the iframe, but this exposes the syscall function globally.
2. I tried returning an iframe or shadow dom based on if a string or element was passed. This kind of worked, but opening panels of one kind after the other causes errors. I don't understand react and useRef enough to fix this.

Idk where to go from here. If you or someone more knowledgeable could fix it, that would be great, otherwise feel free to reject this.

A simple example to see it in action:
```space-lua
command.define {
  name = "Test Panel: Open",
  run = function()
    local link = dom.p {
      "Link to Index",
      onclick = function() editor.navigate("index", false, false) end
    }
    local cmd = dom.p {
      "Open Command Palette",
      onclick = function() editor.openCommandPalette() end
    }
  
    local div = dom.div {
      id = "test-panel",
      link,
      cmd,
    }
    
    editor.showPanel("lhs", 0.8, div)
  end
}
```

### DIFF
--- a/client/components/panel.tsx
+++ b/client/components/panel.tsx
@@ -10,6 +10,118 @@ export function Panel({
   config: PanelConfig;
   editor: Client;
 }) {
+  switch (typeof config.html) {
+    case "string":
+      return <IFramePanel config={config} editor={editor} />;
+    case "object":
+      return <ShadowPanel config={config} />;
+    default:
+      return null;
+  }
+}
+
+function IFramePanel({
+  config,
+  editor,
+}: {
+  config: PanelConfig;
+  editor: Client;
+}) {
+  const iFrameRef = useRef<HTMLIFrameElement>(null);
+
+  const html = useMemo(() => {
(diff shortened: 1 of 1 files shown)


zefhemel:
I'd really like to try to keep existing APIs stable and not regress. So how about another angle, what if I ask you to design a new API "from scratch" so without having to worry what's already there. We can have it be available in parallel and slowly deprecate the old one (maybe). What would it look like? Maybe it doesn't rely on iframes at all, for instance? What are the use cases?

onespaceman:
After a nice break, was able to come back and fix it! Tested it with the treeview panel and Document explorer and no errors, this time!

zefhemel:
Ok, so is this still in draft then?

onespaceman:
oops, forgot I can mark it ready.

Issue #1323, opened 2025-03-29
Vim Mode interferes with some keyboard shortcuts


marad:
For some reason on MacOS when Vim mode is enabled the `Option-H` (open index page) doesn't work. The add quicknote (Shift-Option-N) doesn't work as well. 

When I disable Vim mode - it works as expected.

LangdalP:
I am similarly experiencing that the command palette, which I have bound to Shift-Alt-P, only works in insert mode. This is on MacOS in Firefox.

PR #1821, opened 2026-02-06
Ensure command keybindings work reliably in any vim-mode

rktjmp:
When the vim extension is enabled and in "normal" mode, some default bindings collide with Silverbullets commands (eg: `ctrl-p` - Share, `ctrl-o` - Open Document, `ctrl-e` - Export, `ctrl-g h` - Home). These vim bindings are not always functional (they may appear to do nothing depending on vims state). This gives the impression that Silverbullet is bugged and wont trigger commands. In "insert" mode, most bindings are passed through correctly with some exceptions (eg: `ctrl-o`).

By splitting the command keybindings into their own keymap extension and loading it before the vim extension, we can ensure Silverbullet commands always trigger.

This does mean some vim bindings are overshadowed, but if a binding triggers a Silverbullet command, its generally obvious that *something* happened and the user should be able to recognise why the vim binding failed and then rebind the command or vim mapping.

Closes https://github.com/silverbulletmd/silverbullet/issues/1323

### DIFF
--- a/client/client.ts
+++ b/client/client.ts
@@ -121,7 +121,7 @@ export class Client {
 
   // CodeMirror editor
   editorView!: EditorView;
-  keyHandlerCompartment?: Compartment;
+  commandKeyHandlerCompartment?: Compartment;
   indentUnitCompartment?: Compartment;
   undoHistoryCompartment?: Compartment;
 
--- a/client/client_system.ts
+++ b/client/client_system.ts
@@ -33,7 +33,7 @@ import { languageSyscalls } from "./plugos/syscalls/language.ts";
 import { codeWidgetSyscalls } from "./plugos/syscalls/code_widget.ts";
 import { clientCodeWidgetSyscalls } from "./plugos/syscalls/client_code_widget.ts";
 import { KVPrimitivesManifestCache } from "./plugos/manifest_cache.ts";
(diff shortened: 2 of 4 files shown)


rktjmp:
I think its a better experience that if the documentation says "`ctrl-g h` goes home", it should work and override anything vim-mode may be trying to do, especially considering the number of default binds that are non-functional in normal-mode. Otherwise Silverbullet looks broken.

Assuming that Silverbullets default commands will always be `somemod-somekey` and not raw keys (eg: `a`) then I think this should not interfere with vim mode in general.

I also think it feels more natural to default to Silverbullet, and allow vim-modes users config to act an an overlay on that (unbind, rebind, whatever) rather than having the binding behaviour be inconsistent between no-vim,
vim-normal and vim-insert.

For example, currently if you make a command that happens to collide with a vim map you must also update the config.vim table with a mapping & command to proxy through, if you change the command you must also change those too. It also makes sharing things less plug and play.

Also this is amazing software.

zefhemel:
Agreed, thanks for this. It wasn't quite clear to me why certain keys would work in one case and not in another, this at least makes this consistent. I think there's a case to be made to revisit keybindings for vim specifically (to have more "vim-y" keybindings rather than the default ones inherited from the non-vim mode), but that's another project.

PR #1822, opened 2026-02-06
Make any frontmatter <scheme>:// URI's clickable

### LINKED ISSUE #1818: URIs in frontmatter should be clickable
Currently, it is possible to create a note with clickable URLs in the frontmatter, for example:

```markdown
---
something: "https://example.com"
---
```

In this case, the URL is recognized and rendered as a clickable link. However, custom URI schemes are not recognized and are rendered as plain strings, for example:
```markdown
---
something: "application://resource"
---
```

It would be useful if non-HTTP URI schemes were also recognized as clickable links. This would make it easier to associate notes in the knowledge base with resources in external applications.

rktjmp:
Per https://www.rfc-editor.org/rfc/rfc3986#section-3.1, the scheme must follow the pattern:

    scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )

Closes #1818

### DIFF
--- a/client/codemirror/frontmatter.ts
+++ b/client/codemirror/frontmatter.ts
@@ -12,6 +12,7 @@ import {
   frontmatterQuotesRegex,
   frontmatterUrlRegex,
   frontmatterWikiLinkRegex,
+  frontmatterMailtoRegex,
 } from "../markdown_parser/constants.ts";
 import { processWikiLink, type WikiLinkMatch } from "./wiki_link_processor.ts";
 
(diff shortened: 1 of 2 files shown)


rktjmp:
I made mailto links target self, not blank, as they either pop out to an application or at least here on linux + firefox, open a dialogue to select a handler.

zefhemel:
Thanks! This whole frontmatter link handling part is slightly messy and needs to be clean up, but that's not your doing. This works for now.

PR #1823, opened 2026-02-06
Space Lua: Align numeric/table semantics with Lua, align number formating, optimize loops allocations

mjf:
This change improves Space Lua compatibility with standard Lua 5.4, focusing on numeric subtypes and table behavior. The test suite is extended to lock in the expected semantics and should pass under both Space Lua and a Lua interpreter.

SUMMARY OF CHANGES
------------------

Tighten Lua compatibility across evaluator and runtime:

- correct metamethod dispatch (`__index`, `__newindex`, `__call`, comparison metamethods),

- loop limits,

- raw metamethod lookups.

Rework numeric semantics to preserve Lua **integer** vs **float** behavior:

- $0$ vs $0.0$ and $-0.0$,

- explicit zero kind representation, and

- updates arithmetic/bitwise coercions accordingly.

Improve parser correctness by rejecting **unary plus** with aligned Lua errors and better parsing errors reporting.

Fix `stdlib` behavior to match that of Lua:

- `table` function `concat`, `insert`, `remove`, `sort` and `unpack` gain metamethod awareness and enforce Lua errors,

- `ipairs` iteration updated to stop on first nil and honor `__index`,

- `tonumber` updated to Luae conversion using `luaToNumberDetailed`,

- `math.modf` return corrected,

- `math.type` accuracy improvements for float/integer and $-0.0$,

- `math.pi` added.

Fix numeric subtypes for `/` and `^` operators so `math.type` matches Lua results using tagging as well as unary `-`.

Expand test coverage:

- new `metamethods_test.lua` for Lua metamethod/operator semantics,

- extend arithmetic and length tests for zero-kind propagation and `rawlen` vs `__len` metamethod,

- Extend `math` test suite to test proper Lua alignment (`math.type` and more), and

- update context error expectations for Lua error messages.

RATIONALE
---------

Lua differs from JavaScript by having two numeric subtypes: **integer** and **float**. Operators depend on the subtype: `+`, `-`, `*`, `//` and `%` use integer mode when both operands are integers and float mode otherwise. Bitwise operators require integers and `math.type(x)` reports "integer" or "float". JavaScript has one numeric primitive type (`number`) so a plain number value cannot record whether Lua considers a value to be a float when the value has no fractional part (for example $2.0$).

Lua also differs from "everything is IEEE 754 double" because the rules are defined in terms of integer and float subtypes. Float operations preserve IEEE 754 behavior including `NaN`, infinities and signed zero ($-0.0$) which affects results like $1/0.0$ versus $1/-0.0$. Integer arithmetic does not preserve $-0$ and collapses it to $0$. Lua integer arithmetic is exact within its integer range while JavaScript `number` cannot exactly represent all integers in that range.

Lua numbers are integers or floats. Numeric strings coerce to integer or float based on _lexical_ form. Each arithmetic operator selects the result subtype from the operator rules and operand subtypes.  Integer only operators (bitwise and `//` as integer division) require integer representability. Mixed arithmetic promotes to float as needed.  Two operators are **always float** typed: division (`/`) and exponentiation (`^`) produce floats even if both operands are integers and even if the numeric value has no fractional part. `math.type` reports that internal subtype.

Tables are associative arrays and assigning `nil` removes a key. The length operator `#` uses `__len` metamethod if present otherwise it uses the raw length rule. `rawlen(table)` ignores `__len`. Without `__len` Lua defines `#` as some boundary `N` such that `table[N]` is not `nil` and `table[N+1]` is `nil`. If the table has holes (missing or `nil` entries in the positive integer key sequence) the boundary may be non unique so `#` is stable only for proper sequences without holes.

PERFORMANCE NOTES
-----------------

Numeric changes add small checks to preserve Lua integer and float subtype semantics and avoid allocations except when the subtype would otherwise be lost.

Some table operations may be slower due to stricter Lua 5.4 behavior especially around length and sequence boundary handling which currently requires extra metadata tracking and scans and cannot be avoided without a completely different internal table representation.

TL;DR
-----

The original patchset evolved into serie of additional patchsets, one over the preceeding, all of them somewhat related
one to each other or at least elaborating on the previous ones. See commit messages for further explanation of these
additional patchsets.

### DIFF
--- a/client/space_lua/eval.ts
+++ b/client/space_lua/eval.ts
@@ -89,26 +89,19 @@ import { getBlockGotoMeta } from "./labels.ts";
 const astNumberKindCache = new WeakMap<LuaExpression, NumericType>();
 
 function astNumberKind(e: LuaExpression | undefined): NumericType | undefined {
-  if (!e) {
-    return undefined;
-  }
+  if (!e) return undefined;
 
   const cached = astNumberKindCache.get(e);
-  if (cached) {
-    return cached;
-  }
-
-  let result: NumericType | undefined;
+  if (cached) return cached;
 
   let unwrapped = e;
   while (unwrapped.type === "Parenthesized") {
     unwrapped = unwrapped.expression;
   }
 
-  if (
-    unwrapped.type === "Unary" &&
-    unwrapped.operator === "-"
-  ) {
+  let result: NumericType | undefined;
+
+  if (unwrapped.type === "Unary" && unwrapped.operator === "-") {
     result = astNumberKind(unwrapped.argument);
   } else if (unwrapped.type === "Number") {
     result = unwrapped.numericType === "int" ? "int" : "float";
@@ -126,13 +119,13 @@ function astNumberKind(e: LuaExpression | undefined): NumericType | undefined {
       } else if (lk === "int" && rk === "int") {
         result = "int";
       } else {
-        result = undefined;
+        return undefined;
       }
     } else {
-      result = undefined;
+      return undefined;
     }
   } else {
-    result = undefined;
+    return undefined;
   }
 
   if (result !== undefined) {


client/plugos/plug_compile.ts
```diff
@@ -97,7 +97,11 @@ setupMessageListener(functionMapping, manifest, self.postMessage);
   const outFile = `${destPath}/${manifest.name}.plug.js`;
   await Deno.writeTextFile(inFile, jsFile);
 
-  // console.log("JS FILE", inFile);
+  try {
+    await Deno.remove("deno.lock");
```
zefhemel: Is this still necessary? Not part of a merge conflict resolve?

client/space_lua/stdlib/math_test.lua
```diff
@@ -1,48 +1,182 @@
...
+  assert_close(f2, -0.5, 1e-12, "modf(-3.5) frac")
+end
+
+-- sqrt/exp/log/pow
+do
+  assert_eq(math.exp(0), 1, "exp(0)")
+  assert_close(math.log(math.exp(1)), 1, 1e-12, "log(exp(1))")
+  assert_close(math.log(8, 2), 3, 1e-12, "log base 2 of 8")
+
+  if has_math("pow") then
```
zefhemel: Do we need to check this type of stuff? If not, then something's off, no?

zefhemel: I'd question this whole has_math function in general. Is this copied from some external test suite?

client/space_lua/stdlib/string.ts
```diff
@@ -112,6 +113,10 @@ export const stringApi = new LuaTable({
     },
   ),
   format: new LuaBuiltinFunction((_sf, format: string, ...args: any[]) => {
+    // Unwrap tagged floats so printf sees plain numbers
```
zefhemel: I'd rather you use a `.map` as in other places

client/space_lua/stdlib/format.ts
```diff
@@ -0,0 +1,770 @@
+// Supported specifiers set: [diuoxXaAfeEgGcspq%]
+// Supported flags set: [-+0# ]
+// Width and precision via digits or `*`
+
+interface FormatSpec {
```
zefhemel: We use `type` everywhere else, why `interface` here?

client/space_lua/eval.ts
```diff
@@ -87,6 +86,62 @@ import {
...
+
+    if (numericOp) {
+      const lk = astNumberKind(unwrapped.left);
+      const rk = astNumberKind(unwrapped.right);
+
+      if (lk === "float" || rk === "float") {
+        result = "float";
+      } else if (lk === "int" && rk === "int") {
+        result = "int";
+      } else {
```
zefhemel, line 121: Not a huge deal, but why not `return undefined` (or I suppose simply `return`) here immediately? Then you probably also don't need the `if` undefined check at the end there?

client/space_lua/parse.ts
```diff
@@ -83,6 +84,455 @@ function hasCloseLocal(names: LuaAttName[] | undefined): boolean {
   return false;
 }
 
+function expressionHasFunctionDef(e: LuaExpression): boolean {
```
zefhemel, line 87: How about we move all these analysis functions into a separate module? parse.ts is getting extremely long

deno.json
```diff
@@ -2,16 +2,16 @@
   "name": "@silverbulletmd/silverbullet",
   "version": "2.4.1",
   "tasks": {
-    "build": "make clean && deno run -A build_plugs_libraries.ts && deno run -A build_client.ts",
+    "build": "make clean && deno run -A --no-lock build_plugs_libraries.ts && deno run -A --no-lock build_client.ts",
```
zefhemel: This is probably also from a merge conflict, should be reverted (shouldn't have --no-lock)

mjf: It is from the merge conflict for sure.

mjf: Should I add the new `deno.lock` make process creates or can't we put it to the `.gitignore`?

client/space_lua/parse.ts
mjf: Yeah, why not? Suggest name You would like for it. TBH, I am very tempted to refactor the file hierarchy of the whole Space Lua codebase! Have something like:

- `tests/lua/*.lua` with many small files, perhaps even `tests/lua/stdlib/*.lua` or similar (test suites are of great interest to work with somehow better, also to comment special tests in the small files for good documentation etc. which can also suite as nice study material and intro to the Lua semantics),
- `tests/js/deno/*.test.ts` with Deno-based tests (should never be Lua tests, except few very special cases of interop, for example).

and then also

- `runtime/*.ts`
- `eval/*.ts`
- `parse/*.ts`
- `stdlib/*.ts`

also broken into smaller functional areas.

The `eval.ts` is horrorrous file, I must say, and quite a mess. Some files are mixing in a way they should not (something you would expect in runtime is in eval and vice versa etc.).

I hope I will find some good way to clean this up one day. So, for now, I'd let the `parse.ts` be as is?

client/space_lua/stdlib/format.ts
mjf: Resolved.

deno.json
mjf: Reverted the merge mess.

client/space_lua/stdlib/math_test.lua
mjf: Relict. Resolved.

client/space_lua/stdlib/string.ts
mjf: Ok. But consider that

- `map` always allocates a new array, even if no element changes (!),
- the `for` loop mutates args in place meaning zero allocation.

I follow the following rule of thumb...

> Keep the `for` loop if you care about allocation in hot paths; use `map` if you prefer brevity and the call is not performance-critical.

In this particular case, and given that the `format` is not called in tight loop with many arguments, the impact on performance is therefor negligable (unmeasurable) and I replaced it as requested.

client/plugos/plug_compile.ts
mjf: I cannot know, this is something I haven't touched. It's weird. I, again, suggest to place `deno.lock` in the `.gitignore`. ;)

mjf: Resolved.

client/space_lua/eval.ts
mjf: Since undefined results are never cached anyway early return is actually fine (the cache store only runs when `result !== undefined`) - refactored.

deno.json
zefhemel: deno.lock should be in the repo

client/space_lua/stdlib/string.ts
zefhemel: ok, if this is for performance reasons then let's stick with for loops

PR #1831, opened 2026-02-11
Honor `openIfExists: true` when using `Page: From Template` command

rktjmp:
When using `Page: From Template` the value of `openIfExists` was ignored, which gave different behaviour compared to using a command key to create a page from a template.

`editor.prompt` can also return `""` if the user enters nothing but presses `ok` instead of `cancel`, so tighten the check on the prompt return value.

### DIFF
--- a/libraries/Library/Std/Infrastructure/Page Templates.md
+++ b/libraries/Library/Std/Infrastructure/Page Templates.md
@@ -79,11 +79,16 @@ This is my quick note version
 ```space-lua
 -- priority: 10
 
-local function createPageFromTemplate(templatePage, pageName)
+local function createPageFromTemplate(templatePage, pageName, openIfExists)
   -- Won't override an existing page
   if space.pageExists(pageName) then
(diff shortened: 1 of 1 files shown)


libraries/Library/Std/Infrastructure/Page Templates.md
```diff
@@ -156,11 +159,12 @@ command.define {
     if selected.suggestedName then
       pageName = (template.new(selected.suggestedName))()
     end
-    pageName = editor.prompt("Page name", pageName)
-    if not pageName then
+    pageName = string.trim(some(editor.prompt("Page name", pageName)) or "")
+    if pageName == "" then
+      editor.flashNotification("No page name given for template '" .. cleanName(selected.fullName) .. "', unable to continue.", "error")
```
rktjmp, line 164: Had to do `cleanName(selected.fullName)` instead of `selected.name` to avoid a `lintObjects` error.

zefhemel:
Nice, thank you, once again!

PR #1832, opened 2026-02-11
Alter editor.prompt handling of empty strings

rktjmp:
When working on #1831 I noticed that it was mishandling empty strings. `editor.prompt` could be expanded to do the  behaviour it was expecting, return null when no input is given. (note that in #1831, we're running in lua, where `""` is not falsey, in js these checks are ok aside from accepting `   ` as input.)

`client/window.prompt()` can return `null` (cancel), `""` (no input + ok), or whatever input the user entered. I think it probably makes a lot of sense to trim this input in most cases, eg filenames, titles, even input lines probably want to be trimmed most of the time, but sometimes a user may *want* to insert just whitespace (eg: indentation string for a function call?). In that case they could pass `trim = false`.

Since I felt like you may want untrimmed input (eg: `  space-around-my-string-for-visual-reasons  `), but would still want to consider `""` as "nothing", I separated out the checks for "entered nothing" and "trim or not".

This is an incomplete PR as it probably needs a number of checks to existing `editor.prompt` calls to check they act as expected, if the appetite for the change is there. It's also potentially breaking to downstream.

### DIFF
--- a/client/plugos/syscalls/editor.ts
+++ b/client/plugos/syscalls/editor.ts
@@ -484,8 +484,24 @@ export function editorSyscalls(client: Client): SysCallMapping {
       _ctx,
       message: string,
       defaultValue = "",
+      options: { acceptEmpty?: boolean, trim?: boolean } = { acceptEmpty: false, trim: true }
     ): Promise<string | undefined> => {
-      return client.prompt(message, defaultValue);
+      const { acceptEmpty: acceptEmpty, trim: trim } = options;
+      // prompt() can return the input string, "" if the user just presses ok, or
+      // null if they press cancel.
+      let input = client.prompt(message, defaultValue);
+      if (input == null) {
+        return null;
+      }
+      if (trim) {
+        // input = trim && input.trim() || input; will shortcut incorrectly because "" == falsey
+        input = input.trim()
+      }
+      // Trim again in this check incase trim == false
+      if (!acceptEmpty && input.trim() == "") {
+        return null;
+      }
+      return input;
     },
     "editor.confirm": (_ctx, message: string): Promise<boolean> => {
       return client.confirm(message);
--- a/plug-api/syscalls/editor.ts
+++ b/plug-api/syscalls/editor.ts
@@ -351,6 +351,7 @@ export function dispatch(change: any): Promise<void> {
 export function prompt(
   message: string,
   defaultValue = "",
+  // TODO: also here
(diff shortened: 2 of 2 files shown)


rktjmp:
Ex, these all probably should trim their input values:

https://github.com/silverbulletmd/silverbullet/blob/4fdc8c1af5cdcde0af45eec5b431a0c64673f42d/plugs/index/refactor.ts#L36-L40

https://github.com/silverbulletmd/silverbullet/blob/4fdc8c1af5cdcde0af45eec5b431a0c64673f42d/plugs/index/refactor.ts#L69-L72

https://github.com/silverbulletmd/silverbullet/blob/4fdc8c1af5cdcde0af45eec5b431a0c64673f42d/plugs/index/refactor.ts#L88-L91

This will fail on an empty `""` + ok

https://github.com/silverbulletmd/silverbullet/blob/4fdc8c1af5cdcde0af45eec5b431a0c64673f42d/libraries/Library/Std/Infrastructure/Library.md?plain=1#L43-L46

If you'd rather keep the existing `prompt` behaviour, I dont mind making a separate PR that patches the calls instead.

zefhemel:
I agree that many uses of prompt should trim (all of the examples you give) but not sure that means it needs to be included in the API itself? Is it not ok to simply always trim depending on the case?

> client/window.prompt() can return null (cancel), "" (no input + ok), or whatever input the user entered.

This is what I would expect indeed. This is how it works now, right?

rktjmp:
>>  client/window.prompt() can return null (cancel), "" (no input + ok), or whatever input the user entered.

> This is what I would expect indeed. This is how it works now, right?

Yep, that comment is documenting the browser behaviour, so the current implementation passes that through.

> I agree that many uses of prompt should trim (all of the examples you give) but not sure that means it needs to be included in the API itself? Is it not ok to simply always trim depending on the case?

I think you probably nearly always want to trim, so it would make sense to reduce the "forgot to trim" surface area. I can appreciate keeping it aligned with the default browser behaviour though. As I said, kind of happy to go either way on a PR, there isn't a huge number of prompt calls to check.

I think I have changed my mind on converting `"" -> null`, and instead we should just treat those failure cases with an notification, eg:

`cancel` => silently halt operation
`empty` => "cannot X, you must provide Y/got empty Y"
`value` => do X.

Maybe lets ditch this and just update with trim and notifications where needed?

Issue #1830, opened 2026-02-10
Images are silently overwritten


AngryJKirk:
Let's say you have a page A with some embedded image.png that you uploaded previously.
Then you create a page B, completely unrelated and you also upload an image called image.png.

Both "image.png" files are completely unrelated to each other, but the because they all stored in the same folder the older image.png is getting overwritten without even saying a word to the user.

zeus-web:
I can confirm that behaviour on SB-Version: 2.4.1-94-g4fdc8c1a-2026-02-08T14-23-01Z :-(

rktjmp:
Is the correct behaviour here to 

1. generate a non-conflicting name if a conflict-occurs
    -  `image.png + upload(image.png) = image.png xyzabc-image.png`
       - This is probably the worst option because the behaviour is unpredictable, sometimes your file is what you think sometimes it's not.
    -  `image.png + upload(image.png) = image.png image.1.png`
       - At least most of the filename is what you think, and a sequential integer makes it obvious it was a collision?
2. generate a (probably non-conflicting) prefix for all uploads
    - `upload(image.png) = xyzabc-image.png`
      - makes it harder to link anything
3. check if the file exists after getting the user filename and repeating the prompt on collision
    - `image.png + upload(image.png) = another_name? = another_name? = ...`
    - `image.png + upload(image.png) = collision, new-name or keep existing to replace` 
      - must have non naive behaviour, if the new name also collides, repeat

I think 3 is probably the best option? Supports replacing, does not clobber, user will know what it ended up being called?

https://github.com/silverbulletmd/silverbullet/blob/4fdc8c1af5cdcde0af45eec5b431a0c64673f42d/plugs/editor/upload.ts#L32-L42

zefhemel:
I'd vote for solution 3 yes: keep asking until the user comes up with an unused file name.

PR #1833, opened 2026-02-11
Dont clobber existing files when uploading

rktjmp:
Fix for #1830, 

I updated the `plug/editor/upload.ts` process to prompt the user to confirm the filename to replace, or give another name. This works fine, using the `Upload: File` command.

I tried to call the same implementation in `client/code_mirror/editor_paste.ts` but it wont work because you can't actually call that imported code and the existing syscall does not support passing a callback to handle the different page modification code for each context.

Need guidance on best way to solve this, adding the callback to the syscall or some other interface?

### DIFF
--- a/plugs/editor/upload.ts
+++ b/plugs/editor/upload.ts
@@ -53,7 +53,7 @@ export async function saveFile(file: UploadFile) {
   }
   desiredFilePath = desiredFilePath.trim();
   if (!isValidPath(desiredFilePath)) {
-    await editor.flashNotification(invalidPathMessage, "error");
+    editor.flashNotification(invalidPathMessage, "error");
     return;
   }
 
@@ -80,7 +80,7 @@ export async function saveFile(file: UploadFile) {
       }
       confirmedFilePath = confirmedFilePath.trim();
       if (!isValidPath(confirmedFilePath)) {
-        await editor.flashNotification(invalidPathMessage, "error");
+        editor.flashNotification(invalidPathMessage, "error");
         return;
       }
       if (desiredFilePath === confirmedFilePath) {


client/codemirror/editor_paste.ts
```diff
@@ -202,46 +203,25 @@ export function documentExtension(editor: Client) {
...
-    if (!finalFilePath || !isValidName(finalFilePath)) {
-      return;
-    }
-
-    await editor.space.writeDocument(finalFilePath, file.content);
-    let documentMarkdown = `[[${finalFilePath}]]`;
-    if (file.contentType.startsWith("image/")) {
-      documentMarkdown = "!" + documentMarkdown;
+  async function saveFileAndInsertLink(file: UploadFile) {
+    const uploadedFile = await saveFile(file)
```
rktjmp: Non functional, `TypeError: globalThis.syscall is not a function`.

plug-api/types/client.ts
```diff
@@ -45,6 +45,7 @@ export type UploadFile = {
   name: string;
   contentType: string;
   content: Uint8Array;
+  path?: string;
```
rktjmp: Could be a separate `UploadedFile` type.

zefhemel:
Ok, an in important thing to understand in the SilverBullet codebase is that the while both "client" and "plugs" are written in TypeScript, they should be viewed as being mostly living in isolation (also at runtime). All code part of plugs run in web workers in isolation, and cannot talk directly to the code that lives as part of the client. For plug code to talk to the client it uses syscalls (effectively `syscall` calls, but wrapped in typed interfaces). And the client can call into "plug space" via APIs like `invokeFunction` (you can find some uses in the code base).

While it is technically possible (as you try) to simply import code across these boundaries liberally, you're likely setting yourself up for a mistake (as you see with the errors you're getting).

This means that code reuse between these two areas is basically not happening (with the exception of "plugs-api/lib"), except for some types and constants. Either we go with some version of code duplication (as you're trying to eliminate now), or indeed "cross boundary" API calling (syscalls in one way, invokeFunction in another). In this case, I'd probably go with accepting that this logic needs to be duplicated...

client/codemirror/editor_paste.ts
```diff
@@ -214,22 +214,67 @@ export function documentExtension(editor: Client) {
...
     }
 
+    // Check the given desired file path wont clobber an existing file. If it
+    // would, ask the user to confirm or provide another filename. Repeat this
+    // check for every new filename they give.
+    // Note: duplicate any modifications here to client/code_mirror/editor_paste.ts
+    let finalFilePath = null;
+    while(finalFilePath == null) {
+      console.log(editor.space)
+      // How to call fileExists?
```
rktjmp: @zefhemel What is the right way to call fileExists from the client? I guess this might be an invokeFunction call?

zefhemel: It would be nice to have an API for it directly, right now the right way to do it is this (https://github.com/silverbulletmd/silverbullet/blob/main/client/plugos/syscalls/space.ts#L93):

```
try {
        await client.space.spacePrimitives.getFileMeta(name);
        return true;
      } catch {
        // 404
        return false;
      }
```

client/codemirror/editor_paste.ts
```diff
@@ -214,22 +214,84 @@ export function documentExtension(editor: Client) {
...
+        ensureFilename(file.name),
       ),
     );
-    if (!finalFilePath || !isValidName(finalFilePath)) {
+    if (desiredFilePath === undefined) {
+      // User hit cancel, so they know why we stopped and dont need an notification.
+      return;
+    }
+    desiredFilePath = desiredFilePath.trim();
+    if (!isValidName(desiredFilePath)) {
```
rktjmp, line 253: Note this editor_paste.ts uses isValidName here, where as upload.ts uses isValidPath. They both call isValidPath when checking the filename.

rktjmp:
I think this is set, it does include the change from https://github.com/silverbulletmd/silverbullet/pull/1847, and there is one maybe historical artefact, see comment above.

client/codemirror/editor_paste.ts
rktjmp: As in, the old code had this asymmetry between upload.ts and editor_paste.ts, which I have left, but I'm not sure why they are different, perhaps its a bug.

client/codemirror/editor_paste.ts
```diff
@@ -214,22 +214,84 @@ export function documentExtension(editor: Client) {
...
+    }
+    const notifyInvalidPath = () => {
+      editor.flashNotification(
+        "Unable to upload file, invalid target filename or path",
+        "error",
+      );
+    }
+    const doesFileExist = async (filePath: string) => {
+      try {
+        await client.space.spacePrimitives.getFileMeta(filePath);
```
zefhemel: I know this works (magic global `client`) but please use `editor` here (scoped), which actually should be renamed to `client`, but that's another matter.

client/codemirror/editor_paste.ts
```diff
@@ -214,22 +214,84 @@ export function documentExtension(editor: Client) {
       );
       return;
     }
+    const ensureFilename = (filename: string) => {
```
zefhemel: While I will not claim this code is a work of art, still stylistically I'd prefer to use regular function declarations rather than the `const name = () => ...` format (here and elsewhere):

```
function ensureFilename(filename: string){
   ...
}
```

zefhemel:
Thank you!

PR #1836, opened 2026-02-13
[Space Lua] Extend `os`

mjf:
* Add `os.getenv`
* Add `os.clock`

See detailed notes in commit messages.


zefhemel:
Space Lua only ever runs in the client (except for tests), you know this, right? I don't really see how the environment variable API could really be a usefully implemented unless we start to export server-side environment variables to the client, but not sure we should. Security reasons etc.

mjf:
Ok, I though some use case will be found. I will remove it for now...

PR #1843, opened 2026-02-17
[LIQ] Extend with `group by` and `having` with aggregators

mjf:
Also update documentation with extensive examples and add some parser tests.

Note: `make generate` JS files included.

### DIFF
--- a/website/Space Lua/Lua Integrated Query/Aggregating.md
+++ b/website/Space Lua/Lua Integrated Query/Aggregating.md
@@ -1,3 +1,5 @@
+#maturity/experimental
+
 The `group by` and `having` clauses of [[Space Lua/Lua Integrated Query]] support aggregate functions for grouped analysis, following SQL-style semantics.
 
 After `group by`, each result row contains:
--- a/website/Space Lua/Lua Integrated Query/Grouping.md
+++ b/website/Space Lua/Lua Integrated Query/Grouping.md
@@ -1,3 +1,5 @@
+#maturity/experimental
+
 The `group by` and `having` clauses extend [[Space Lua/Lua Integrated Query]] with SQL-style grouping and aggregate filtering.
 
 After `group by`, each result row has two fields:


client/space_lua/stdlib/string.ts
```diff
@@ -18,7 +18,7 @@ import {
 } from "./pattern.ts";
 
 function capturesToLua(caps: CaptureResult[]): any {
-  if (caps.length === 0) {return null;}
```
zefhemel, line 22: I should add a code style file, or once we migrate to node.js add a proper linter that checks this, but I really dislike if statements without curly braces, because its' so easy to add one more statement and then having to add braces afterwards.

```
if(caps.length === 0) {
   return null;
}
```

is my preferred format

website/Space Lua/Lua Integrated Query/Aggregating.md
```diff
@@ -0,0 +1,196 @@
...
+  select {
+    tag = key,
+    total = count(name),
+    min_size = min(size),
+    max_size = max(size),
+    avg_size = avg(size)
+  }
+]]}
+
+**With binding variable:**
```
zefhemel, line 54: Honestly I'd rather just skip examples WITHOUT binding a variable altogether, the version with binding is so much clearer and unambiguous (because no globals are going to be overwritten).

And with some new "magic" variables being introduced like `key` and aggregator functions, this problem only gets worse. I'd remove support for queries without binding to a variable altogether, but it's too late.

website/Space Lua/Lua Integrated Query/Aggregating.md
```diff
@@ -0,0 +1,196 @@
...
+    tags[1],
+    tags[2]
+  select {
+    first = key[1],
+    second = key[2],
+    count = count(name)
+  }
+]]}
+
+## Using `_` as the item variable
```
zefhemel: I'd also omit this one, for similar reasons. People should just get used to using queries with bound variables.

website/Space Lua/Lua Integrated Query/Aggregating.md
```diff
@@ -0,0 +1,196 @@
...
+  select {
+    tag = key,
+    first_page = name,
+    n = count(name)
+  }
+]]}
+
+## Custom aggregators
+
+Custom aggregator functions may be defined by the user. See [[website/CustomAggregates.md]] for details and usage examples.
```
zefhemel: I don't see this website/CustomAggregates.md file?

website/Space Lua/Lua Integrated Query/Grouping.md
```diff
@@ -0,0 +1,525 @@
...
+
+# Examples
+
+All examples below use `index.tag 'tag'`.
+
+## Group by single key
+
+Group all tags by `name`:
+
+```sql
```
zefhemel: Using `sql` is a bit confusing

website/Space Lua/Lua Integrated Query.md
```diff
@@ -78,10 +80,47 @@ Or to select 5 pages tagged with `#meta`:
...
+- `group` — a table (array) of all original items in that group
+
+The `group by` field names are also available as bare variables in `having`, `select`, and `order by`. Use `#group` to get the count of items in a group.
+
+Example:
+
+${query[[
+  from p = index.tag "tag"
+  group by p.name
+  select { name = name, count = #group }
```
zefhemel: This doesn't look right, `name = p.name`?

client/space_lua/aggregates.ts
```diff
@@ -0,0 +1,201 @@
...
+      return fn(sf, ...args);
+    },
+    asString() {
+      return "<builtin aggregate>";
+    },
+  };
+}
+
+// Built-in aggregate specs
+const builtinAggregates: Record<string, AggregateSpec> = {
```
zefhemel, line 48: Is there no way to **not** make aggregates a special type of thing? Can aggregator "functions" not be Lua tables that live in the regular Lua namespace, and have `initialize` and `iterate` callbacks for instance? 

Some may be built-in (implemented in JS perhaps, for performance reasons) and live in a certain namespace, e.g. `liq.sum`, or `agg.sum`? Custom ones then wouldn't need a special aggregate.define, but could really be something along the lines of:

```
agg.myCustomAggregator = {
  initialize = function() ... end,
  iterate = function() ... end,
}
```

client/space_lua/stdlib/string.ts
mjf: But that does `deno fmt <file>`. I have no idea what to do with it. I always first do `eslint` to add these, and `deno fmt` then removes some. I also want them, but I rely on `deno fmt` (or `make fmt`) to have the code always as in the rest of the code base.

website/Space Lua/Lua Integrated Query/Aggregating.md
mjf: But I (and I think most users too) does not want to bind variable for simple-enough queries, I think. It also aligns better with SQL syntax. In SQL you can also do both, with/without binding (at least in Postgres I only refer to if unsure about something).

website/Space Lua/Lua Integrated Query/Aggregating.md
mjf: I would completely drop the `_` stuff. Only with/without binding, no extra `_`, it's so weird.

website/Space Lua/Lua Integrated Query/Grouping.md
mjf: Ah, ok. That was meant to have some syntax highlighting. In fact, I haven't seen the rendered docs yet. I can remove all of these everywhere...

client/space_lua/aggregates.ts
mjf: the `aggregate.define` mimicks closely already existing APIs. I also wanted the aggregates to be easily listed and maintained. To avoid user mistakes, I wanted proper specs for them too because they represent exact contract with exact semantics.

client/space_lua/stdlib/string.ts
zefhemel: Ok let's leave it for now. When we switch to node I'll add prettier or some other formatting tool that will fix these things.

website/Space Lua/Lua Integrated Query/Grouping.md
zefhemel: If you turn it into `lua` and put the proper query[[ stuff around it you should get syntax highlighting for it? If not, you may have to add the new keywords to client/space_lua/parse.ts in the list of keywords under `luaStyleTags`

client/space_lua/stdlib/string.ts
mjf: +1 for the Prettier!

website/Space Lua/Lua Integrated Query/Aggregating.md
mjf: I remove the `_` example.

client/space_lua/aggregates.ts
zefhemel: Yeah, but here you're basically introducing a new Lua language feature: aggregator functions that only are accessible (I assume) within queries and have a separate namespace. You even had to add a special check traversal for this. I don't really like that unless it's absolutely necessary. It also adds potential confusion in case people for instance would define a global variable that clashes with a name of an aggregator function.

Having a need for proper specs and checking I see. Thus far this has mostly been handled with JSON schema, but that's likely not a good match here. I'd almost argue to add types like Luau https://create.roblox.com/docs/luau/type-checking but let's not go there ;)

Honestly though, I do not think (famous last words) that people would add many custom aggregator functions and likely stick to the built in ones. Maybe adding some rudimentary checks (is `initialize` and `iterate` defined) would be sufficient?

website/Space Lua/Lua Integrated Query/Aggregating.md
mjf: Solved by linking to the API.

website/Space Lua/Lua Integrated Query/Grouping.md
mjf: Replaced.

client/space_lua/aggregates.ts
mjf: We will need more APIs smilar to this one asthe LIQ will extend in the future, not only the aggregates, IMHO. Once You established this style, I would keep with it. You can have `sum()` for normal array, but the aggregate for LIQ is not just that. It also eases adding more and more performant aggregates in TS (for performance reasons).

website/Space Lua/Lua Integrated Query.md
mjf: Fixed (both should work).

mjf: I was too fast to respond, sorry. The correct answer is: **no**.

The `select` part operates not over a single record, but over each group (i.e., the set of records sharing the same group key). The select clause’s environment is generally:

- `name` (the group key)
- `group` (the table of records in the group)
- no `p` bound (`p` is not set; it was the row variable during scan, **but you are now at the group level**!)

In the `select { name = name }` the `name` refers to the group key (what you `group by`-ed).

You typically use:

- `name = name` (group key, as in the query), or
- if you want data from the first record (say all grouped `p.name` are the same and non-`nil`): `name = group[1].name`

I hope it's clearer now.

PR #1851, opened 2026-02-23
[Space Lua] Implement `string.pack`, `string.unpack` and `string.packsize`

mjf:
The three binary-packing functions from Lua are now implemented in a dedicated `string_pack.ts` module and wired into `stringApi`.

Format parsing follows the exact option set of Lua:

* signed and unsigned integers of configurable width (`b`/`B`/`h`/`H`/`l`/`L`/`j`/`J`/`T`/`i`/`I`),
* IEEE 754 single and double precision floats (`f`/`d`/`n`),
* fixed-length strings (`c`),
* length-prefixed strings (`s`),
* zero-terminated strings (`z`),
* padding bytes (`x`),
* alignment padding (`X`), and
* endianness/alignment modifiers (`<`/`>`/`=`/`!`).

`string.packsize` rejects variable-length formats (`s` and `z`) as Lua requires.

`string.unpack` returns all unpacked values followed by the next position integer, matching the Lua contract exactly.

Test suite extended in separates `string_pack_test.lua` file. Passes both Lua and Space Lua execution.

### DIFF
--- a/client/space_lua/lua.test.ts
+++ b/client/space_lua/lua.test.ts
@@ -53,6 +53,10 @@ Deno.test("[Lua] String tests", async () => {
   await runLuaTest("./stdlib/string_test.lua");
 });
 
+Deno.test("[Lua] String pack/unpack/packsize tests", async () => {
+  await runLuaTest("./stdlib/string_pack_test.lua");
+});
+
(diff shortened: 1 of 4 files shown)


zefhemel:
Very nice

Issue #1849, opened 2026-02-22
Odd spacing when adding order by


adxsoft:
${query[[from tags.link select _.toPage]]} lists pages with links to them correctly

${query[[from tags.link select _.toPage order by _.toPage]]} show results with a space between each character in the page name

Using sb version 2.4.1-0-g3e1f03b2

mjf:
For me it does exactly the same but even for the simplest:
`${query [[ from tags.link select toPage ]] }`!

If I do:
`${query [[ from tags.link select tostring(toPage) ]]}`
I get the list.

So it seems the `${...}` does something weird. Needs further investigation...

mjf:
I hope #1853 may fix this. @zefhemel ?

PR #1853, opened 2026-02-24
Hotfix: Skip `nil` results in query select to prevent null rows corrupting renderer

mjf:


mjf:
Attempt to fix #1849.

mjf:
Maybe we should also tighten the `lua_widget.ts` condition:

```typescript
  if (
    Array.isArray(result) && result.length > 0 &&
    result[0] !== null && result[0] !== undefined &&
    typeof result[0] === "object"
  ) {
  ```
  
  @zef ?

zefhemel:
I think it's ok like this

PR #1854, opened 2026-02-24
Trim user input from prompts where appropriate

rktjmp:
Re: https://github.com/silverbulletmd/silverbullet/pull/1832

Adds separate checks for when the user clicks "cancel" vs enters an empty string. In most cases we trim the input string before checking if its empty, when doing a prefix rename we allow empty or entirely white space strings to maintain user intent.


rktjmp:
Possibly there is a future bug in there, where if the user does a prefix rename, with the old prefix of "my file" and the new prefix of "", you'd try to create an empty file?

I think it *is* correct to allow an empty new prefix with the intent to *remove* the old prefix.

PR #1858, opened 2026-02-26
[LIQ] Fix `group by`: `order by` with `select` aliases, variable binding in `having` and `nil` column ordering

mjf:
1. Fix `order by` inability to access `select` output aliases and aggregate functions after `group by`.

2. Introduce `LUA_SQL_NULL` sentinel to preserve column ordering when `select` expressions evaluate to `nil`.

   Note: In Lua, setting a table key to `nil` deletes it so rows with `nil` values lose those keys and produced inconsistent insertion order across result tables.

### DIFF
--- a/website/Space Lua/Lua Integrated Query/Aggregating.md
+++ b/website/Space Lua/Lua Integrated Query/Aggregating.md
@@ -15,31 +15,15 @@ Field names used in `group by` are exposed as locals in `having`, `select`, and 
 > The `having` clause acts only on grouped output. For filtering individual items, use `where` prior to grouping.
 
 # Examples
-
-All queries operate on `index.tag 'page'`.
+All example queries operate on `tags.page`, but will work with any query collection. As always, to see the underlying query, hover over the result table and click the _Edit_ button to see the underlying query.
 
 ## Counting with and without binding
-
 Grouping pages by their first tag, and computing the count and aggregate statistics:
 
-**Without binding variable:**
-
-```lua
-query[[from
-  index.tag 'page'
-group by
-  tags[1]
-select {
-  tag = key,
-  total = count(name),
-  min_size = min(size),
-  max_size = max(size),
-  avg_size = avg(size)
-}]]
-```
+**Without binding variable**
 ${query [[
   from
-    index.tag 'page'
+    tags.page
   group by
     tags[1]
   select {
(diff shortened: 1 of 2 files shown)


client/space_lua/query_collection.ts
```diff
@@ -32,10 +32,18 @@ import { executeAggregate, getAggregateSpec } from "./aggregates.ts";
 export { buildItemEnv } from "./query_env.ts";
 import { buildItemEnv } from "./query_env.ts";
 
+// Sentinel value representing SQL NULL in query results.
+export const LUA_SQL_NULL = Symbol.for("silverbullet.sqlNull");
```
zefhemel: I understand the reference to SQL here, but we're not implementing SQL so perhaps not use it? LIQ_NULL maybe?

client/space_lua/query_collection.ts
```diff
@@ -44,16 +52,18 @@ function buildGroupItemEnv(
...
+  itemEnv.setLocal("_", item);
+
   if (item instanceof LuaTable) {
     const keyVal = item.rawGet("key");
     const groupVal = item.rawGet("group");
 
-    if (!objectVariable && groupVal instanceof LuaTable) {
+    // Always unpack first group item's fields as locals, regardless of
+    // objectVariable, so bare field names (e.g. `name`) are accessible
+    // in having/select/order by after group by.
```
zefhemel: Is this really necessary and desired? Personally I will avoid to ever write a query without binding an objectVariable, because I'm afraid any attribute found in an item may shadow a global variable, leading to very weird results (think: a `table` attribute, `js` attribute or whatever that shadows the global var). Is it not possible to just keep those values available via `o.name` or whatever object variable I picked?

client/space_lua/query_collection.ts
```diff
@@ -469,57 +572,164 @@ export async function applyQuery(
...
-      return 0; // All orderBy clauses were equal
-    });
+        return 0; // All orderBy clauses were equal
+      });
+
+      // Reorder both arrays according to sorted indices
+      results = indices.map((i) => results[i]);
+      selectResults = indices.map((i) => selectResults![i]);
+    } else {
+      results = await asyncQuickSort(results, async (a, b) => {
```
zefhemel: This (visually) looks like a lot of code duplication with the if-true case, is this absolutely necessary?

client/space_lua/query_collection.ts
mjf: I was naming it with SQL "NULL ternary logic" in mind (not yet implemented). What about `LIQ_SQL_NULL` denoting that's something that is/behaves like SQL NULL? (Again, not yet.)

client/space_lua/query_collection.ts
mjf: No longer there.

client/space_lua/query_collection.ts
mjf: No longer there.

client/space_lua/query_collection.ts
mjf: Renamed to `LIQ_NULL` as requested.

PR #1859, opened 2026-02-26
Improve Lua widget rendering

mjf:
Improve Lua widget rendering

- **Scalars** (number, string, boolean): Rendered directly as the bullet
  or cell content.

- **Empty table/object**: Rendered as `*(empty table)*`.

- **Lists (arrays)**: Rendered as Markdown lists, with correct handling
  of nested lists or tables as items.

- **Plain objects/`LuaTables` with string keys only**: Rendered as
  a Markdown table.

- **Mixed tables (numeric and string keys)**: Rendered as a single-row
  Markdown table.

Users can still copy widget-rendered Markdown and paste as real content,
materialize it.

Markdown syntax does not allow tables nested deeper than one level (e.g.,
tables inside tables); only lists and tables as immediate list item
content.


zefhemel:
This looks, but it's a bit hard to check. Would this this be a good opportunity to add (at least the beginnings) of a test suite?

mjf:
Complete re-work on the go, reflecting recent changes and with improved semantics and without UX-breaking changes.

PR #1865, opened 2026-03-02
Make ToC widget collapsible

Dobli:
This replaces the current markdown based ToC widget in the `std` library with a dom elements based one. This allows us to wrap the whole list in a `<details>` element that provides a native open/close ability, making the ToC collapsible. It comes with an additional option to define whether the ToC shall be open or collapsed by default.

**Pictures**:

collapsed
open

### DIFF
--- a/libraries/Library/Std/Widgets/Widgets.md
+++ b/libraries/Library/Std/Widgets/Widgets.md
@@ -124,8 +124,9 @@ function widgets.toc(options)
   options = options or config.get("std.widgets.toc")
   options.minHeaders = options.minHeaders or 3
   options.minLevel = options.minLevel or 1
-  options.defaultOpen = options.defaultOpen or nil
   options.header = options.header or "Table of Contents"
+  local defaultOpen = (options.defaultOpen ~= false) or nil
+
   local text = editor.getText()
   local pageName = editor.getCurrentPage()
   local parsedMarkdown = markdown.parseMarkdown(text)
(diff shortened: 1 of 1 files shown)


zefhemel:
I like this, but there are a lot of in-line styles. Could you replace those with classes defined with space-style instead? I think this would be cleaner.

Dobli:
You're right, also allows to easily restyle the elements for further customization. Adapted it, only kept the in-line style to dynamically set the level indents.

zefhemel:
Ok this is getting there, just tried this out myself and two pieces of feedback:

* The background color of the top-bar is transparent now (it was gray before), I think it would be good to get the standard top and bottom widget styling back to what it was for consistency.
* The default collapse state seems to be collapsed. I'm not sure if this is the right default, I'd prefer it to be opened, but also would really like it that this state is persisted in the client. On a space level, so not per page (that would be too complicated). So when I collapse the TOC on one page, it will also have the TOC collapsed on other pages, and vice versa. Or even store this per page? I'm not sure on this one, open to ideas.

Dobli:
I adapted the header to use the grey background as it did before, looks good in both collapsed state as well as open. Can also make further adjustments if necessary.

image

Regarding the state handling, I have the personal preference to have a predictable behaviour when navigating between pages, so I like the current approach with only the default state configurable. I changed the default behaviour to open when not configured now. This also leaves the internal state handling fully to native html elements making the implementation quite simple. Storing (and retrieving) the state would require additional event handling.

In case client side collapsed state handling is required, I will check how to add it or it could be another iteration in case someone feels the need for it, as the new implementation already provides an advantage to the current ToC :smile: .

zefhemel:
Nice, thank you

PR #1867, opened 2026-03-03
[Space Lua] LIQ: Add `using ...` support to `order by`

mjf:
* support for custom comparators in LIQ `order by` clauses via the `using` keyword.

* `using` is mutually exclusive with `asc`/`desc` and accepts either a *named function* or an *inline anonymous function*. The comparator receives two values and returns true if the first should come before the second.

* `nulls first`/ `nulls last` works independently of `using`.

* `using` is now a reserved keyword (`kw`; this cannot be avoided but the chance to collide with user scripts is negligable),

* documentation updated + polished a little

### DIFF
--- a/client/space_lua/query_collection.ts
+++ b/client/space_lua/query_collection.ts
@@ -12,6 +12,7 @@ import type {
 } from "./ast.ts";
 
 import {
+  jsToLuaValue,
   luaCall,
   LuaEnv,
   LuaFunction,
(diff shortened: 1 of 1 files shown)


mjf:
Also includes generated parser file (`make generate`) and, of course, some tests we added too.

mjf:
Fixed #1868 here because of the complex code changes in this PR and because it is part of the `group by` / `order by` serie anyway which the issue in fact refers to.

PR #1870, opened 2026-03-06
Fix table indexer shifting columns when cells are empty

mjf:
For example, with the following table:

```markdown
| a   | b   | c   |
| --- | --- | --- |
| 1   |     | 3   |
| 1   | 2   | 3   |
| a   | b   | c   |
| a   | b   |     |
|     | b   | c   |
| x   |     |     |
|     |     | z   |
```

the following query results in dataset with shifted columns when cells are empty:

```lua
from
  index.tag 'table'
where
  page == 'test'
order by
  pos asc
select {
  a = a, b = b, c = c
}
```

This is now fixed and gives correct results:

| a   | b   | c   |
| --- | --- | --- |
| 1   |     | 3   |
| 1   | 2   | 3   |
| a   | b   | c   |
| a   | b   |     |
|     | b   | c   |
| x   |     |     |
|     |     | z   |

### DIFF
--- a/plugs/index/table.ts
+++ b/plugs/index/table.ts
@@ -49,6 +49,29 @@ function concatChildrenTextsPreserveLinks(nodes: ParseTree[]): string {
   return nodes.map((c) => renderToText(c)).join("").trim();
 }
 
+/**
+ * Ensure a TableRow has a TableCell between every pair of TableDelimiters.
+ * The parser omits TableCell nodes for empty cells; this fills them in.
+ */
+function normalizeTableRow(row: ParseTree): void {
+  const children = row.children!;
+  const normalized: ParseTree[] = [];
+  let lookingForCell = false;
+  for (const child of children) {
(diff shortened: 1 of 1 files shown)


mjf:
# Explanation why this could not be solved in the parser

The Lezer parser operates at the *syntax tree construction* level. It builds `Element` nodes with `from` and `to` byte positions into the source text. A `TableCell` node requires a range of source text to span. For an empty cell like `| |` there is no content between the pipes (only whitespace, which the parser explicitly skips). There is literally nothing to anchor a `TableCell` node to (no `from`, no `to`, no children). Creating a zero-width or whitespace-only node would violate the Lezer tree model where every node maps to a meaningful span of source text. The semantic interpretation "a gap between two delimiters means an empty cell in column" is not the parser's job. That's application-level logic! The renderer already understood this and filled in the gaps. The indexer now does the same, using the same proven normalization (now to be shared by both parser and renderer).

PR #1876, opened 2026-03-09
Improve Lua Widget Rendering

mjf:
# Overview

Lua `${...}` expression results are rendered using two parallel paths:

- **HTML path** (display path) for visual display in the editor and
- **Markdown path** (copy path) for the *copy button* output.

# Value Classification

Every Lua value is classified as one of:

- **Scalar:** number, string, boolean, or `nil`
- **Empty table:** `{}`
- **Array:** a `LuaTable` with only sequential integer keys (no string keys)
- **Array of tables:** an array where every element is a `LuaTable`
- **Record:** a `LuaTable` with any string key (may also have integer keys)

# HTML Rendering (Display Path)

Scalars are rendered as inline Markdown - wiki links, tags, bold, and other inline syntax are resolved, `nil` renders as `<span data-empty>`.

Empty tables render as `<table data-table-empty>`.

An array of tables renders as a multi-row `<table>`. Column headers are the union of all element keys. Each cell value recurses through the same classification.

An array of scalars (or a mix of scalars and tables) renders as `<ul>` with one `<li>` per element. Each item value recurses.

A record renders as a single-row `<table>` with one column per key. Cell values recurse.

Every `<td>` carries `data-table-cell-type` and optionally `data-table-cell-empty`.  Every `<li>` carries `data-list-item-type` and optionally `data-list-item-empty`.  The outermost element carries `data-type`.  These attributes allow CSS styling of rendered widgets.

Nesting is **unlimited**. Tables inside cells, lists inside cells, tables inside list items all render correctly in HTML.

# Markdown Rendering (Copy Path)

Scalars render as plain text. Lua `nil` renders as `nil`. Empty tables render as `*(empty table)*`.

An array of tables renders as a pipe-delimited Markdown table. Nested values in cells are rendered as Lua literals via `toStringAsync()` (e.g. `{1, 2, 3}` or `{a = 1}`), since Markdown table cells cannot contain block-level content.

An array of scalars renders as a Markdown list (`- item`). When a list item is itself multi-line (a sublist) it uses an empty bullet with two-space indented content on the next line:

```
- 1
- 2
-
  - nested 1
  - nested 2
```

Nested tables or arrays **beyond the first list nesting** level are rendered as Lua literals.

A record renders as a single-row Markdown table.

# Column Ordering

Column headers follow the order returned by `LuaTable.keys`: string keys in their insertion order, followed by integer keys in ascending order. For example, `{a = 1, b = 2, 3, d = 4}` produces columns `a | b | d | 1`. This is consequence of `LuaTable` storing string keys and the array part separately (for performance).

# Migration Note

Arrays mixing plain scalars and record-like tables (e.g.  `{1, 2, 3, {a=3}}`) were previously rendered as a single flat table due to an internal JS conversion (`toJS`). They now render as a list, with record elements shown as nested tables - which more accurately reflects their Lua structure. This is the only divergence from the original presentation.

To restore the old tabular layout, wrap all elements as records with a consistent key:

```lua
-- Before (was a flat table, now a list with nested table):
${{1, 2, 3, {a=3}}}

-- After (explicit records, renders as a table):
${{{value=1}, {value=2}, {value=3}, {a=3}}}
```

### DIFF
--- a/client/space_lua/render_lua_html.ts
+++ b/client/space_lua/render_lua_html.ts
@@ -309,6 +309,9 @@ function renderCellContent(
   renderInline: InlineRenderer,
 ): Promise<string> {
   if (isEmpty(v)) return Promise.resolve("");
+  if (isTaggedFloat(v)) {
+    return Promise.resolve(renderInline(luaFormatNumber(v.value, "float")));
+  }
   if (v instanceof LuaTable) {
     if (v.empty()) return Promise.resolve(`<table data-table-empty></table>`);
     return renderLuaTableToHtml(v, renderInline);


mjf:
[Widget Rendering Test Suite](https://github.com/user-attachments/files/25862279/widget_test_suite.md)

PR #1879, opened 2026-03-12
Fix Markdown table rendering and remove deprecated command button remnants

mjf:


zefhemel:
Would it be possible to add some tests at least to the table parser part?

mjf:
Added tests.

PR #1882, opened 2026-03-12
Fix extended task rendering in widgets

mjf:
TL;DR in individual commit messages.

### DIFF
--- a/client/codemirror/lua_widget.ts
+++ b/client/codemirror/lua_widget.ts
@@ -215,6 +215,9 @@ export class LuaWidget extends WidgetType {
         currentName,
         mdTree,
         this.opts.client.clientSystem.spaceLuaEnv,
+        {
+          rewriteTasks: false,
+        },
       );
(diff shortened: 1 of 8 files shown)


zefhemel:
So, actually when a previewed task contains a page reference it DOES work and this is extremely useful, try this:

```
* [ ] Hello

${template.each(query[[from t = tags.task where t.page == editor.getCurrentPage()]], templates.taskItem)}
```
before your patch, actually toggling the checkbox updated the original task too. Now this no longer works because it's disabled. I think it can still be disabled IF no such reference exists though.

mjf:
> before your patch, actually toggling the checkbox updated the original task too. Now this no longer works because it's disabled. I think it can still be disabled IF no such reference exists though.

This is fixed and extended of the extended status tasks. More thinkgs are extended as well (API validation added and `order` field for custom task states).

zefhemel:
Nice improvements!

PR #1886, opened 2026-03-15
fix(client): append file input to body to fix iOS Safari upload dialog

### LINKED ISSUE #1883: File:Upload not working on iPhone/iPad
Hello,

the command File:Upload appears not working in Safari on iPhone and iPad. When selecting the command nothing happens, the Run Command dialog simply disappears. Expecting behavior is that a file/photo picker will pop up and let me select file of photo I want to upload. 

https://github.com/user-attachments/assets/6f8abec8-bf85-4425-af15-3db6aab48047

Copying and pasting and image works fine.

Environment: 
SilverBullet 2.5.1 in Docker
Safari browser on iOS 17

Thanks and cheers,
Petr

Sigmabrogz:
## Summary
Fixes an issue where the file picker dialog does not pop up in iOS/iPadOS Safari when invoking `File:Upload` or using the `Upload: File` command button.

## Problem
Fixes #1883.
On iOS Safari, calling `.click()` on an isolated, detached DOM `<input type="file">` element (created dynamically via `document.createElement`) does not reliably trigger the native OS file picker overlay unless the input element is part of the active DOM tree.

## Solution
Briefly appended the hidden `<input>` element to `document.body` before calling `.click()`, then cleaned it up from the DOM shortly after (1 second delay to ensure the OS has time to present the dialog before removing the source element).

### DIFF
--- a/client/plugos/syscalls/editor.ts
+++ b/client/plugos/syscalls/editor.ts
@@ -284,7 +284,10 @@ export function editorSyscalls(client: Client): SysCallMapping {
           reject(e);
         };
 
+        input.style.display = "none";
+        document.body.appendChild(input);
         input.click();
+        setTimeout(() => document.body.removeChild(input), 1000);
       });
     },
     "editor.flashNotification": (


Sigmabrogz:
This addresses the `File:Upload` prompt silently dropping dead without opening the iOS file picker overlay.

zefhemel:
Awesome. Thanks!

PR #1890, opened 2026-03-16
Fix mobile tasks, annoying caret jumping and empty widgets on navigation

mjf:
* Tapping a task state on mobile could fire twice before the first change finished causing corrupted status text.  Overlapping cycles are now prevented and looking up positions is done **at tap time**.

* Typing on mobile triggered extra decoration rebuilds and cursor resets during IME input making the caret jump around.  These operations are now defered until composition finishes.

* Navigating between pages quickly could show empty widgets because the index had not finished updating after save.  We wait for indexing to complete before loading the next page now.


zefhemel:
Just merged your other PR, now we have a conflict

PR #1891, opened 2026-03-17
[LIQ] Add new aggregate functions, aliases, and queryable aggregate registry

mjf:
* Extend with 13 new built-in aggregates: `product`, `string_agg`, `yaml_agg`, `json_agg`, `bit_and`, `bit_or`, `bit_xor`, `bool_and`, `bool_or`, `stddev_pop`, `stddev_samp`, `var_pop` and `var_samp`.

* Introduce `aggregate.alias` API allowing users to define custom aliases for any aggregate. Standard aliases (`every`, `std`, `stddev` and `variance`) are now defined via this API rather than hardcoded.

* Add `index.aggregates` queryable collection so users can discover all available aggregates directly from LIQ queries.

**UPDATE:**

* Add guards against `LIQ_NULL` leaking in misc. way in many places!

TL;DR in commit messages... :)

### DIFF
--- a/website/Space Lua/Lua Integrated Query/Aggregating.md
+++ b/website/Space Lua/Lua Integrated Query/Aggregating.md
@@ -18,7 +18,21 @@ Field names used in `group by` are exposed as locals in `having`, `select`, and 
 
 All registered aggregate functions — built-in, user-defined, and aliases — can be listed via `index.aggregates()`:
 
-${query[[from index.aggregates() order by name]]}
+${query[[
+  select
+  {
+    Name = '`' .. name .. '`',
+    Description = description,
+    Kind =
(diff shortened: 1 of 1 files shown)


mjf:
This too fixes custom aggregates that were in fact never working.

mjf:
Also fixes table functions not working in intermediate state in aggregates.

mjf:
Switched to draft again, working on `LIQ_NULL` leakage in many places.

mjf:
I believe most real issues are now solved.

mjf:
Ok, I was wrong, now I **believe** we are ok to go... :fireworks:

zefhemel:
Very cool, but could you add some tests for the new aggregates?

PR #1892, opened 2026-03-17
fix: check SB_READ_ONLY for "true" instead of any non-empty value

jbrunner:
Setting SB_READ_ONLY=false incorrectly enabled read-only mode because the check used `!= ""` instead of `== "true"`. This contradicts the documented behavior which states to set the variable to "true" to enable read-only mode.


zefhemel:
Thanks! But I decided to update the docs instead since I already have some sites relying on setting this to `1` 😉

PR #1909, opened 2026-03-25
[LIQ] Unify `from`, `select`, and `group by` clauses to use field lists, implement multi-source cross-join

mjf:
Unify `from`, `select`, and `group by` clauses to use field lists
=================================================
Query clauses `from`, `select`, and `group by` now accept the same field
syntax as Lua table constructors (`name = expr`, bare expr, or `[expr]
= expr`), giving them a consistent grammar and enabling named bindings
everywhere. #feature

What changed:

* New `FieldList` rule in `lua.grammar`. All three clauses now parse
  their arguments through it instead of using expression lists or the
  special `Name "=" exp` pattern.

* AST types `LuaFromClause`, `LuaSelectClause`, and `LuaGroupByClause`
  now carry fields: `LuaTableField[]` instead of a single expression or
  an expression array. The parser, static analysis helpers and the
  evaluator are all updated to work with this new shape.

Named bindings in `select`
--------------------------
`select name, age` now derives column names from bare variable or
property access expressions so the result table has string keys (`name`,
`age`) rather than integer indices. Explicit naming still works.

Named bindings in `group by`
----------------------------
`group by` fields can carry explicit aliases that propagate into the
post-grouping scope. For example:

```sql
group by
  n = name
select {
  label = n
}
```

The alias `n` is now visible in `select`, `having`, and `order by`.
Previously only the raw expression form (`name`) was bound and an
explicit alias was silently lost. #bugfix

A new `LuaGroupByEntry` type pairs each `group by` expression with its
optional alias. The environment builder binds every
declared alias to the corresponding key value, regardless of whether the
group key is a scalar or a multi-key table.

Named bindings in `from`
------------------------
`from p = page` continues to work as before. The clause now also accepts
the full field-list syntax which lays the groundwork for multi-source
from (cross join) in a future change. #todo #feature

Complete LIQ syntax now
-----------------------
Complete LIQ syntax (in Postgres documentation style) could be now
roughly expressed as follows: #todo #documentation

```text
query[[ clause [...] ]]

where clause is one of:

    FROM field [, ...]
    WHERE condition
    GROUP BY field [, ...]
    HAVING condition
    SELECT field [, ...]
    ORDER BY sort_key [, ...]
    LIMIT count [, offset]
    OFFSET start

where field is one of:

    expression
    name = expression
    [expression] = expression

where sort_key is:

    expression [ ASC | DESC ]
               [ NULLS { FIRST | LAST } ]
               [ USING name ]
               [ USING FUNCTION function_body ]

where expression is any valid Lua expression, including:

    aggregate_call
    aggregate_call FILTER ( WHERE condition )
```

Signed-off-by: Matouš Jan Fialka <mjf@mjf.cz>

### DIFF
--- a/client/space_lua/eval.ts
+++ b/client/space_lua/eval.ts
@@ -750,22 +750,110 @@ function fieldsToGroupByEntries(fields: LuaTableField[]): LuaGroupByEntry[] {
   });
 }
 
-function fromFieldsToSource(
-  fields: LuaTableField[],
-  ctx: ASTCtx,
-): { objectVariable?: string; expression: LuaExpression } {
+type FromSource =
+  | { kind: "single"; objectVariable?: string; expression: LuaExpression }
+  | { kind: "cross"; sources: { name: string; expression: LuaExpression }[] };
+
+function fromFieldsToSource(fields: LuaTableField[], ctx: ASTCtx): FromSource {
   if (fields.length === 1) {
     const f = fields[0];
     if (f.type === "ExpressionField") {
-      return { expression: f.value };
+      return { kind: "single", expression: f.value };
     }
     if (f.type === "PropField") {
-      return { objectVariable: f.key, expression: f.value };
+      return { kind: "single", objectVariable: f.key, expression: f.value };
     }
   }
-  return {
-    expression: { type: "TableConstructor", fields, ctx },
-  };
+
+  const sources: { name: string; expression: LuaExpression }[] = [];
+  for (const f of fields) {
+    if (f.type !== "PropField") {
+      throw new LuaRuntimeError("Multi-source 'from' requires named sources", {
+        ref: ctx,
+      } as any);
+    }
+    sources.push({ name: f.key, expression: f.value });
+  }
+  return { kind: "cross", sources };
+}
+
+async function normalizeToArray(
+  collection: LuaValue,
+  sf: LuaStackFrame,
+): Promise<any[]> {
+  if (collection instanceof LuaTable && collection.empty()) {
+    return [];
+  }
+  if (collection instanceof LuaTable) {
+    if (collection.length > 0) {
+      const arr: any[] = [];
+      for (let i = 1; i <= collection.length; i++) {
+        arr.push(collection.rawGet(i));
+      }
+      return arr;
+    }
+    return [collection];
+  }
+  if (Array.isArray(collection)) {
+    return collection;
+  }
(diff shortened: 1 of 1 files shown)


mjf:
Few artifical examples of what works now with the multi-source cross-join:

```pgsql
${query [[
  from
    x = {   1,   2 },
    y = {  10,  20 },
    z = { 100, 200 }
  select {
    sum = x + y + z
  }
]]}
```

|sum|
|--|
|111|
|211|
|121|
|221|
|112|
|212|
|122|
|222|

-----

Or more candy...

```pgsql
${query [[
  from
    x = {   1,   2,   3, },
    y = {  10,  20,  30, },
    z = { 100, 200, 300, }
  group by
    x, y, z
  having
    first(x) == 3  and
    y        == 30 and
    last(z)  == 300
  select
    x,
    y,
    z,
    sum = x + y + z
]]}
```

|x|y|z|sum|
|--|--|--|--|
|3|30|300|333|

-----

Or (more complicated three-way cross-join) to illustrate:

```pgsql
${query [[
  from
    s = {
      { id = 1, name = 'Eva',  },
      { id = 2, name = 'Adam', },
      { id = 3, name = 'John', },
      { id = 4, name = 'Zef',  },
    },
    c = {
      { id = 101, title = 'Mathematics',      },
      { id = 102, title = 'Arts',             },
      { id = 103, title = 'Physics',          },
      { id = 104, title = 'Computer Science', },
      { id = 105, title = 'Literature',       },
    },
    e = {
      { sid = 1, cid = 101, },
      { sid = 1, cid = 103, },
      { sid = 2, cid = 101, },
      { sid = 2, cid = 102, },
      { sid = 2, cid = 103, },
      { sid = 3, cid = 102, },
      { sid = 3, cid = 105, },
      { sid = 4, cid = 101, },
      { sid = 4, cid = 103, },
      { sid = 4, cid = 104, },
    }
  where
    s.id  == e.sid and
    e.cid == c.id
  order by
    s.name
  group by
    s.name
  having
    s.name:match('^A') or
    s.name:match('f$')
  select
    student = s.name:upper(),
    courses = string_agg(c.title, ', '
      order by c.title desc
    )
]]}
```

|student|courses|
|--|--|
|ADAM|Physics, Mathematics, Arts|
|ZEF|Physics, Mathematics, Computer Science|

zefhemel:
This is super cool man!

Issue #1935, opened 2026-04-15
Runtime API - failed to navigate: chrome failed to start


Luminiferous348:
Heya,

I kinda have an issue with the Runtime API. Specifically, the Runtime API just errors out with:
<img width="1799" height="89" alt="Image" src="https://github.com/user-attachments/assets/e3b8d6b9-ffd3-444c-b26f-a6e1ec8f8bc9" />

It also seems to crash quite early into the process. It manages to create the /space/.chrome-data folder, but it then crashes before it can even write anything into the folder:

<img width="573" height="102" alt="Image" src="https://github.com/user-attachments/assets/6c2738e9-c92c-432f-851b-6eeee5e9fa5d" />

As for the normal stdout of silverbullet, it just tells me this:

<img width="1663" height="294" alt="Image" src="https://github.com/user-attachments/assets/080ea48b-9fff-42cb-9101-173471f332fe" />

---

As for my setup: I run Silverbullet in Docker, on the "latest" tag, so "2.6.1-runtime-api". This is my compose:
```
networks:
  default:
    enable_ipv6: True
    ipam:
      config:
        - subnet: fd80:10::/64
    name: traefik-net
services:
  core:
    environment:
      - PUID=568
      - PGID=568
      - SB_NAME=Luminiferous SilverBullet
    expose:
      - 3000
    image: ghcr.io/silverbulletmd/silverbullet:latest
    restart: always
    volumes:
      - /etc/localtime:/etc/localtime:ro
      - /mnt/primary/silverbullet:/space
```

I do have traefik with oauth-proxy2 infront of it, but this shouldn't matter in this case, since I directly running curl from within the docker container. So curl fully bypasses the reverse Proxy and the oauth-proxy, and directly runs against the port 3000 within the container.

As for the storage, well, docker does its docker things and mounts the containers / as overlayFS, and the /space/ folder is a bind mount, pointing to a ZFS partition. I also tried to put chrome-data on /chrome-data (yes, I tried it without creating the folder manually, and with creating the folder beforehand and setting the 568 user to the owner of the /chrome-data folder), and it also doesn't work.  
I also tried to bind mount /space with 777 chmod, and it also doesn't work.
So either /chrome-data can not be on a overlayFS or a ZFS volume, or its not a problem with the /chrome-data folder. At least not as far as I can see.

So yeah... is it possible that the Docker container is just straight up broken? At least the Runtime API Component?

zefhemel:
[...]

PR #1944, opened 2026-04-22
Fixing #1935

Luminiferous348:
Fixes this issue https://github.com/silverbulletmd/silverbullet/issues/1935

By removing the -H flag from adduser, the adduser utility creates a /home/silverbullet folder. This is needed for Chromium, since it doesn't start without a home folder, breaking the Runtime API.

This doesn't affect Chromium, if the Container runs as root, since the root user always has its home under /root. But this fixes things for other users.


zefhemel:
Thanks, you beat me to it, I had more or less this exact fix pending. Thanks!

Issue #1494, opened 2025-08-19
changing to proportional font also makes text within backticks proportional


xkcd386at:
I recently discovered Silverbullet and am in the process of moving my personal wiki to this; very nice tool.

I am using the latest "edge" version as of yesterday or the day before.

What I find is that adding

    ```space-style
    html {
      --editor-font: "Arial" !important;
    }

causes not only the normal text but also the text inside backticks is now proportional.

I have rarely, till now, worked with CSS so I am not quite sure how to achieve this.  However, Firefox's "Inspect" menu seems to indicate that both the normal text *and* the `code` text are controlled by the same variable (editor-font).

Any help appreciated.

aphymi:
You can fix this with a little more CSS by setting the backtick-text's corresponding CSS classes to go back to the monospace font:
```css
html {
  --editor-font: "Arial" !important;
}

.sb-line-fenced-code, .sb-code {
  font-family: "iA-Mono", "Menlo";
}
```

Alternatively, instead of re-writing the `--editor-font` CSS variable, you can just set the font on the specific editor element, then set the code CSS classes back to what they were originally:
```css
#sb-editor .cm-content {
  font-family: Arial;
}

.sb-line-fenced-code, .sb-code {
  font-family: var(--editor-font);
}
```

Does that work for you?

xkcd386at:
**Both** of the solutions you offered worked for me -- thank you so much (and sorry for the delay in trying it out/responding).

(The only caveat is that indentation of second and subsequent lines in a bullet list is now very odd looking...!)

I'm using the second one.  My logic, with my almost-zero knowledge of CSS, is that it seems less "disruptive", with its changes being at a deeper level.

Thanks again and I am closing the issue.

> You can fix this with a little more CSS by setting the backtick-text's corresponding CSS classes to go back to the monospace font:
> 
> html {
>   --editor-font: "Arial" !important;
> }
> 
> .sb-line-fenced-code, .sb-code {
>   font-family: "iA-Mono", "Menlo";
> }
> 
> Alternatively, instead of re-writing the `--editor-font` CSS variable, you can just set the font on the specific editor element, then set the code CSS classes back to what they were originally:
> 
> #sb-editor .cm-content {
>   font-family: Arial;
> }
> 
> .sb-line-fenced-code, .sb-code {
>   font-family: var(--editor-font);
> }
> 
> Does that work for you?

PR #1945, opened 2026-04-23
Wire sb-directive class up to ${...} blocks

Schmavery:
The `sb-directive` class was defined in `client/markdown_parser/customtags.ts` and mapped to `ct.DirectiveTag` in `client/style.ts`, but the parser never actually attached `DirectiveTag` to any node — seems like it wasn't being used?

This makes themes / `space-style` overrides unable to target query bodies, which came up when I was trying to do something like https://github.com/silverbulletmd/silverbullet/issues/1494

### DIFF
--- a/client/markdown_parser/parser.ts
+++ b/client/markdown_parser/parser.ts
@@ -87,7 +87,10 @@ const WikiLink: MarkdownConfig = {
 
 const LuaDirectives: MarkdownConfig = {
   defineNodes: [
-    { name: "LuaDirective" },
+    {
+      name: "LuaDirective",
+      style: { "LuaDirective/...": ct.DirectiveTag },
+    },
     { name: "LuaExpressionDirective" },
     { name: "LuaDirectiveMark", style: ct.DirectiveMarkTag },
   ],


zefhemel:
This makes sense but in practice seems to result in a few styling regressions. The line with a directive seems to be highlighted, and some other issues too. Probably these were latent issues, but it doesn't look good.

Schmavery:
Ah sorry, you're totally right. There were existing styles in editor.scss and colors.scss that I totally forgot about!

Would you rather I remove the styling from the existing `.sb-directive`? Or maybe just add a new one like `.sb-lua-directive-source` on the off change someone was relying on this class somehow.

Alternatively I'm happy to close the PR if you feel like it's too low priority to support this right now. I can work around it with something hacky like this.

```
.cm-line:has(> .sb-keyword,
             > .sb-variableName,
             > .sb-typeName,
             > .sb-string,
             > .sb-number,
             > .sb-operator,
             > .sb-bool,
             > .sb-atom,
             > .sb-directive-mark) {
  font-family: var(--editor-font) !important;
}
```

PR #1949, opened 2026-04-25
Disable splitting to prevent including `import` statements in `service_worker.js`

sirikon:
`import` statements in Service Workers fail in any Firefox version older than 147. Latest Firefox version is 150, but I suggest to disable this setting to make SilverBullet work for latest Firefox ESR, which right now sits at 140 and is the version packaged by Debian stable (13).

Next ESR version that should support the feature is expected for July 21st: https://whattrainisitnow.com/release/?version=esr

### DIFF
--- a/build/build_client.ts
+++ b/build/build_client.ts
@@ -33,7 +33,7 @@ export async function buildClient(): Promise<void> {
     minify: true,
     jsxFactory: "h",
     // metafile: true,
-    splitting: true,
+    splitting: false,
     format: "esm",
     chunkNames: ".client/[name]-[hash]",
     jsx: "automatic",
--- a/build/build_client.ts
+++ b/build/build_client.ts
@@ -14,17 +14,7 @@ export async function buildClient(): Promise<void> {
 
   console.log("Now ESBuilding the client and service workers...");
 
-  const result = await esbuild.build({
-    entryPoints: [
-      {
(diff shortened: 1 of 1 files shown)


zefhemel:
I'm ok with this for the service worker, but if splitting is also disabled for the client we lose a bunch of loading advantages. Can you figure out how to do disable this just for the service worker? Perhaps split this into two separate esbuild steps.

sirikon:
Supposed that there would be no advantage to split dependencies for a single entrypoint if this entrypoint would load everything anyway, just with more roundtrips.

Will look into it later anyway :)

**CORRECTION**: We'll definitely benefit from it even with a single entry point. My bad!: https://esbuild.github.io/api/#splitting
> Code referenced through an asynchronous import() expression will be split off into a separate file and only loaded when that expression is evaluated.

sirikon:
@zefhemel Did some funkyness in there to maintain the behavior while minimizing duplicity. Let me know what you think and if you prefer it done in some other way.

Tested locally and verified that:
- `wget 'http://localhost:3000/.client/client.js'` -> has `import` statements.
- `wget 'http://localhost:3000/service_worker.js'` -> has **no** `import` statements.
- Everything seems to work with a very light manual testing.

zefhemel:
Thanks!

PR #1959, opened 2026-05-01
Add destination property to link objects

MrMugame:
This adds a destination property to `link` objects which is just the resolved full ref including all details, i.e. `@123`, `#etc`, ...

### DIFF
--- a/plugs/index/link.ts
+++ b/plugs/index/link.ts
@@ -12,6 +12,7 @@ import {
 import type { FrontMatter } from "./frontmatter.ts";
 import { updateITags } from "./tags.ts";
 import {
+  encodeRef,
   getNameFromPath,
   isMarkdownPath,
   parseToRef,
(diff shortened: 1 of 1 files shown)


zefhemel:
What do you think of putting the parsed version of the ref in destination? Then you can query against it more deeply. Some other attributes of the link object become effectively duplicates then, but perhaps we can start a deprecation of the "legacy" attributes.

MrMugame:
Yeah, that would make the link objects cleaner, but I see some "issues":
- Users can't directly access the "name", refs only store the path and the `getNameFromPath()` isn't accessible in space lua and I don't know if the distinction name/path is not a little too clunky for most users. The same goes for `isMarkdownPath()`.
- What do we do for external links? We can't really generate a valid ref here, so we have to somehow split up that case. Maybe make it a whole new object? Maybe just make destination an optional. Unsure.

zefhemel:
Ok, let's do it this way then.

PR #1974, opened 2026-05-10
Avoid parsing version file multiple times

CnTeng:
Avoid parsing version file multiple times.


zefhemel:
Handled differently in the soon to be merged Rust backend, thanks though.

PR #1975, opened 2026-05-10
Add UNIX socket support for server

CnTeng:
Add UNIX socket support for server.

More here: https://serverfault.com/questions/124517/what-is-the-difference-between-unix-sockets-and-tcp-ip-sockets

> UNIX domain sockets know that they’re executing on the same system, so they can avoid some checks and operations (like routing); which makes them faster and lighter than IP sockets. So if you plan to communicate with processes on the same host, this is a better option than IP sockets.

### DIFF
--- a/server/cmd/server.go
+++ b/server/cmd/server.go
@@ -49,6 +49,10 @@ func buildConfig(bundledFiles fs.FS, args []string, buildTime string) *server.Se
 		serverConfig.Port = int(port)
 	}
 
+	if os.Getenv("SB_UNIX_SOCKET") != "" {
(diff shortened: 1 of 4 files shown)


zefhemel:
What's the use case for this since this is a web app?

CnTeng:
> What's the use case for this since this is a web app

Almost all reverse proxy applications support use unix socket directly.

https://caddyserver.com/docs/caddyfile/directives/reverse_proxy

https://dev.to/vearutop/using-nginx-as-a-proxy-to-multiple-unix-sockets-3c7a

zefhemel:
Alright

Issue #1956, opened 2026-04-29
Virtual pages only work with one capture group


laurybueno:
The [docs on Virtual Pages](https://silverbullet.md/Virtual%20Pages) mention we can use multiple capture groups in a URL, but I don't think this is the case (at least not anymore).

Trying the example provided:
```lua
virtualPage.define {
  pattern = "lookup:(.+):(.+)",
  run = function(type, id)
    return "# " .. type .. "\nLooking up: " .. id
  end
}
```

Triggers the following error when I navigate to `/lookup:foo:bar`:
```
[Client] Error while dispatching event editor:pageCreating : w: attempt to concatenate a nil value
```

This problem might be related to the change  https://github.com/silverbulletmd/silverbullet/commit/f29f043ff3ae4da54ecdf3230ab59722b0d1b05f.

Issue #1781, opened 2026-01-19
Change virtual pages to not rely on editor:pageCreating event


zefhemel:
Currently virtual pages are hacked on the `editor:pageCreated` event, this is rather ugly (also results in a lot of unnecessary server round trips and 404s), it would be nicer to reimplement this in a different layer. And while at it, also allow virtual pages to appear in the [page picker](https://community.silverbullet.md/t/virtual-page-recent-journal-entries/3778/2?u=zef) (if they don't rely on patterns).

PR #1980, opened 2026-05-13
Resolve virtual pages natively (instead of via editor:pageCreating)

Schmavery:
Partial fix for #1781 (the `editor:pageCreating` half).
Fix for #1956

Move virtual page resolution out of `editor:pageCreating` and into the client. We try to read the real page first, on a miss we check `config.virtualPages`. @zefhemel I hope this is anything like you were imagining 😆 

This fixes the 404-loop that you get on virtual pages.
I'm interested in tacking the second half of #1781 next to see if we can somehow/partially index the virtual pages so that they can show up in the picker and maybe that linking to them doesn't look like a dead link? And not show up in aspiring pages. But wanted to do this part first.

Could also probably pull out the cursor `|^|` extraction bit into a util at this point

### DIFF
--- a/website/Virtual Pages.md
+++ b/website/Virtual Pages.md
@@ -70,6 +70,8 @@ virtualPage.define {
 ```
 
 # How it works under the hood
-Virtual page definitions are stored in the config system under the `virtualPages` key. When a page is being created, SilverBullet fires the `editor:pageCreating` event. The standard library's event listener checks all registered patterns against the page name. If a match is found, the corresponding `run` function is called, and its return value becomes the page content with read-only permissions.
+Virtual page definitions are stored in the config system under the `virtualPages` key. When you navigate to a page, SilverBullet first tries to load it as a real file; only if that lookup misses does it check the registered patterns against the page name. If a pattern matches, the corresponding `run` function is called and its return value becomes the page content with read-only permissions.
+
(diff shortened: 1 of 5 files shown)


website/Virtual Pages.md
```diff
@@ -70,6 +70,8 @@ virtualPage.define {
 ```
 
 # How it works under the hood
-Virtual page definitions are stored in the config system under the `virtualPages` key. When a page is being created, SilverBullet fires the `editor:pageCreating` event. The standard library's event listener checks all registered patterns against the page name. If a match is found, the corresponding `run` function is called, and its return value becomes the page content with read-only permissions.
+Virtual page definitions are stored in the config system under the `virtualPages` key. When you navigate to a page, SilverBullet first tries to load it as a real file; only if that lookup misses does it check the registered patterns against the page name. If a pattern matches, the corresponding `run` function is called and its return value becomes the page content with read-only permissions.
+
+This means a real file always shadows a matching virtual page — for example, creating `tag:my-tag.md` overrides the default tag virtual page just for that one tag.
```
ryanwwest, line 75: Since many OSes don't support colons in filenames, would another example be better? I say this with a grain of salt because IndexedDB's 'files' objectStore itself is fine with this filename.

PR #1981, opened 2026-05-15
More journal robustness

schroederc:
Extend checking e.date to the "previous" case


zefhemel:
Odd, I thought I already did this. Thanks!

PR #1995, opened 2026-05-21
perf: optimize indexing queue performance, add UI-activity throttling, and non-blocking Space Lua script evaluation

gleber:
@zefhemel This is a low effort PR. It's more of a bug report with an AI fix :) Please take it with a huge grain of salt, or just ignore it (but please let me know if you do!). I've vibecoded it using Gemini 3.5 Flash and I do not understand these changes. My intention here is to give you a starting point for whenever you may choose to focus on performance of the web UI.

When SB was loading my humongous space (a separate problem I'll tackle one day), the UI was getting jittery and unresponsive. Every click (even on empty text), most scrolling, any interaction with file picker or command pallet, was getting stuck for 0.1 to even 10 seconds.

I've asked the AI to profile and intercept what the client was doing at load time. And asked it to make UI responsive despite trying to load (sync? also index?..) all of the files.

It does seemingly work, in the sense it makes the text portion of the UI responsive. In the console logs I can see it is detecting changes and queueing indexing of hundreds of files, and the UI is responsive all throughout this time.

Hope it helps you in any way.

## Claude Haiku Explanation of Changes

This PR optimizes SilverBullet's indexing system to **keep the UI responsive even when the space contains a huge number of files**. The core problem: IndexedDB queries were saturating the main thread, freezing the UI while the indexing queue processed thousands of messages. The solution is multi-layered:

---

### **1. In-Memory Queue State Caching (mq.datastore.ts)**

**Problem**: `isQueueEmpty()` was calling `getQueueStats()` on every check, which queried IndexedDB three times (queued, processing, DLQ counts). With large queues, this created constant DB contention.

**Solution**: Add in-memory caches that track queue counts:
- `queuedCounts` and `processingCounts` Maps store current counts
- `ensureQueueInitialized()` loads counts once per queue from DB, then all subsequent operations update the in-memory caches
- After `send()`, `poll()`, `ack()`, `requeueTimeouts()`, and `flushQueue()`, the counts are updated in-memory instead of re-querying

**Example**: The first call to `isQueueEmpty("indexQueue")` hits the DB once to load the counts. Every subsequent call reads from the Map—zero DB overhead.

---

### **2. Event-Driven Queue Waiting (mq.datastore.ts)**

**Problem**: `awaitEmptyQueue()` used a polling loop with `sleep(200)`, waking up every 200ms to check if the queue was empty. For long indexing runs, this hammered IndexedDB thousands of times.

**Solution**: Replace polling with event-driven signalling:

```typescript
// OLD: spin in a loop every 200ms
public async awaitEmptyQueue(queue: string): Promise<void> {
  while (true) {
    if (await this.isQueueEmpty(queue)) break;
    await sleep(200);
  }
}

// NEW: wait for an event signal
public async awaitEmptyQueue(queue: string): Promise<void> {
  if (await this.isQueueEmpty(queue)) return;
  await new Promise<void>((resolve) => {
    const handler = async () => {
      if (await this.isQueueEmpty(queue)) {
        this.eventHook.removeLocalListener(`mq:emptyQueue:${queue}`, handler);
        resolve();
      }
    };
    this.eventHook.addLocalListener(`mq:emptyQueue:${queue}`, handler);
  });
}
```

When the queue worker detects the queue is empty, it dispatches `mq:emptyQueue:${queue}` (already in the code), which wakes up the listener immediately.

---

### **3. Queue Pause Mechanism (mq.datastore.ts)**

**New feature**: Pause/unpause queues without blocking:

```typescript
private pausedQueues = new Set<string>();

public setQueuePaused(queue: string, paused: boolean) {
  if (paused) {
    this.pausedQueues.add(queue);
  } else {
    this.pausedQueues.delete(queue);
    this.wakeupWorker(queue);
  }
}

public isQueuePaused(queue: string): boolean {
  return this.pausedQueues.has(queue);
}
```

In `QueueWorker.run()`, if a queue is paused, the worker sleeps for 500ms instead of processing messages.

---

### **4. Activity-Based Pause/Resume (plugos/hooks/mq.ts)**

**Key insight**: When the user is typing or clicking, pause indexing; when idle for 2 seconds, resume.

```typescript
if (typeof window !== "undefined") {
  let idleTimer: any = null;
  const onActivity = () => {
    // Pause indexing queue when user is active
    this.mq.setQueuePaused("indexQueue", true);
    if (idleTimer) clearTimeout(idleTimer);
    idleTimer = setTimeout(() => {
      this.mq.setQueuePaused("indexQueue", false);
    }, 2000); // Resume indexing after 2s of idleness
  };

  const throttledActivity = throttle(onActivity, 250);

  const events = ["mousedown", "mousemove", "keydown", "scroll", "touchstart"];
  for (const ev of events) {
    window.addEventListener(ev, throttledActivity, { passive: true });
  }
}
```

This ensures indexing **never starves the main thread** when the user is actively editing.

---

### **5. Increased Poll Interval & Inter-Batch Delays (plugos/hooks/mq.ts)**

**Problem**: The index queue defaults to a 1-second poll interval, which hammers IndexedDB. Also, the worker processes batches back-to-back with zero yield.

**Solution**:
- For `indexQueue`, enforce a **minimum 5000ms poll interval** (5 seconds)
  - The wakeup mechanism still delivers messages instantly, so this is just a fallback
- Add an **inter-batch delay of 1000ms** between processing batches
  - This yields the event loop so IDB reads and UI rendering can run

```typescript
const pollInterval =
  queue === "indexQueue"
    ? Math.max(subscriptionDef.pollInterval ?? 0, 5000)
    : subscriptionDef.pollInterval;
const interBatchDelay = queue === "indexQueue" ? 1000 : undefined;
```

---

### **6. Navigation Blocking with Timeout (data/object_index.ts)**

**Problem**: Navigation waits for the index queue to drain, which could freeze the UI for seconds.

**Solution**: Use `awaitEmptyQueueWithTimeout()` with a 2-second cap:

```typescript
public async awaitEmptyQueueWithTimeout(
  queue: string,
  timeoutMs: number,
): Promise<void> {
  if (await this.isQueueEmpty(queue)) return;
  await race([this.awaitEmptyQueue(queue), sleep(timeoutMs)]);
}
```

In `awaitIndexQueueDrain()`:
- Temporarily unpause if needed (to allow the saved page to index immediately)
- Wait up to 2 seconds for the queue to drain
- Restore pause state

```typescript
async awaitIndexQueueDrain(): Promise<void> {
  const wasPaused = this.mq.isQueuePaused("indexQueue");
  if (wasPaused) {
    this.mq.setQueuePaused("indexQueue", false);
  }
  try {
    await this.mq.awaitEmptyQueueWithTimeout("indexQueue", 2000);
  } finally {
    if (wasPaused) {
      this.mq.setQueuePaused("indexQueue", true);
    }
  }
}
```

---

### **7. Batch Index Clearing (data/object_index.ts)**

**New method**: `batchClearFileIndexes()` clears multiple files' indexes in parallel chunks of 128:

```typescript
public async batchClearFileIndexes(files: string[]): Promise<void> {
  const allKeys: KvKey[] = [];
  const chunkSize = 128;
  for (let i = 0; i < files.length; i += chunkSize) {
    const chunk = files.slice(i, i + chunkSize);
    await Promise.all(chunk.map(async (file) => {
      // Strip .md, query prefix, collect keys
    }));
  }
  if (allKeys.length > 0) {
    await this.ds.batchDelete(allKeys);
  }
}
```

This is more efficient than clearing files individually.

---

### **8. Yielding in Space Lua Loading (space_lua.ts)**

When loading all Space Lua scripts, insert a `sleep(0)` between script evaluations:

```typescript
for (const script of allScripts) {
  await sleep(0);  // Yield to let other work run
  const ast = parseLua(script.script, { ref: script.ref });
  await evalStatement(ast, scriptEnv, sf);
}
```

This prevents the Lua evaluation loop from blocking the event loop.

---

### **9. Type Updates (plug-api/types/datastore.ts)**

Add a new `interBatchDelay` option to `MQSubscribeOptions`:

```typescript
export type MQSubscribeOptions = {
  batchSize?: number;
  pollInterval?: number;
  /** Minimum delay in ms between processing batches when queue is non-empty.
   * Yields the event loop so other async work (UI, IDB reads) can run. */
  interBatchDelay?: number;
};
```

---

### **10. Comprehensive Tests**

Two new test suites validate the optimizations:

1. **In-Memory Queue Optimization Test**: Verifies that after initial DB load, `countQuery` is not called again until the queue is flushed
2. **Queue Pause Throttling Test**: Ensures paused queues don't process messages, and unpausing resumes processing

Both use `vi.spyOn()` to track DB calls and confirm the optimization works.

---

## Summary

The PR transforms indexing from a **blocking, DB-heavy operation** into a **responsive, event-driven, pause-aware system**:

| Problem | Solution |
|---------|----------|
| Constant DB queries for queue state | In-memory count caching + single initialization per queue |
| Polling overhead (`isQueueEmpty` every 200ms) | Event-driven signalling + wakeup mechanism |
| Indexing starves main thread | Activity-based pause/resume; 1s inter-batch delays |
| Frozen navigation during drain | 2-second timeout cap on `awaitIndexQueueDrain()` |
| Bursty DB contention | Reduced poll interval (5s fallback) + chunk-based operations |
| Lua loading blocks UI | `sleep(0)` between script evaluations |

**Result**: A massive SilverBullet space can now index in the background without freezing the editor. The user can navigate, edit, and interact freely while indexing happens at reduced priority.

### DIFF
--- a/client/data/mq.datastore.test.ts
+++ b/client/data/mq.datastore.test.ts
@@ -1,12 +1,11 @@
+import type { EventHookT } from "@silverbulletmd/silverbullet/type/manifest";
 import { expect, test, vi } from "vitest";
-import { DataStoreMQ } from "./mq.datastore.ts";
-import { MemoryKvPrimitives } from "./memory_kv_primitives.ts";
-import { DataStore } from "./datastore.ts";
-
 import type { MQMessage } from "../../plug-api/types/datastore.ts";
 import { EventHook } from "../plugos/hooks/event.ts";
 import { System } from "../plugos/system.ts";
-import type { EventHookT } from "@silverbulletmd/silverbullet/type/manifest";
+import { DataStore } from "./datastore.ts";
+import { MemoryKvPrimitives } from "./memory_kv_primitives.ts";
+import { DataStoreMQ } from "./mq.datastore.ts";
 
 test("DataStore MQ", async () => {
   vi.useFakeTimers();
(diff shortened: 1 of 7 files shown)


zefhemel:
Hi, thanks for the effort but  I'm going to hold off on any significant changes to the indexer while @mjf is working on https://github.com/silverbulletmd/silverbullet/pull/1968

What would be more valuable is a detailed bug report with the performance issues you see. It's quite hard to judge this AI assessment, because there's usually a mix of confident bullshit with actual insight and I'd rather dig into it myself (with my own tools) than figure out if the particular model used makes any sense or not.

PR #1996, opened 2026-05-21
feat: fix indexing progress indicator in top bar to show up even at 0%

gleber:
@zefhemel This is a second low effort PR! As before, please take it with a huge grain of salt, or just ignore it (but please let me know if you do!). Done with Gemini 3.5 Flash and I do not understand these changes. Same intention.

When SB was loading my humongous space (a separate problem I'll tackle one day), the UI was not giving me any indicator. But then after a few minutes it suddenly appeared as "index progress: 1%".

I've asked the AI to dig deeper and figure out why progress indicator was missing. This is the result

Hope it helps you in any way.

## AI summary of the work:

- Add indexingQueuingCount field to track pending items *before* they hit the queue, ensuring progress starts at 0% rather than jumping to 1%+
- Track indexing queue size in client-side memory to show progress 0-100% without polling the DB
- Update maxIndexQueueSize eagerly in the iles:changed listener so the full batch size is captured immediately
- Reduce startIndexProgressTracker polling interval from 1000ms to 200ms for smoother, more responsive progress updates
- Remove plug-side updateIndexProgressInUI() timer from plugs/index/queue.ts to avoid conflicting updates with the new client-side tracker
- Add getQueueSizeInMemory() to DataStoreMQ for cheap, non-blocking in-memory queue size reads
- Fix showProgress() in editor_ui.tsx to only set 5s auto-dismiss timer when a percentage is provided (not when clearing)
- Batch ile:changed events into a single iles:changed event in evented_space_primitives.ts for efficient indexing
- Clean up stale client. global references to use 	his. in client.ts
- Fix language_core_test.lua assertEqual to handle nil message argument
- Add unit test for in-memory queue size tracking

### DIFF
--- a/client/client.ts
+++ b/client/client.ts
@@ -146,6 +146,10 @@ export class Client {
   widgetCache!: WidgetCache;
   objectIndex!: ObjectIndex;
 
+  private maxIndexQueueSize = 0;
+  private indexingQueuingCount = 0;
+  private indexProgressInterval: any = null;
+
   constructor(
     private parent: Element,
     public bootConfig: BootConfig,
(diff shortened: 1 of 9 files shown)


zefhemel:
This is way more code than should be required for this and it's hard to fully understand. Can you file this as a bug report instead?

PR #2000, opened 2026-05-25
Fix: Encode extra-ascii space path in HTTP header

### LINKED ISSUE #1988: 'Space folder path different on server' when the command-line argument contains extra-ASCII characters
How to reproduce:
```bash
mkdir ~/Pensées
silverbullet ~/Pensées
```

Visit localhost:3000, you get an alert saying 'Space folder path different on server, reloading the page' and when it reloads the alert comes back in indefinite recursion.

It does not happen when the command-line argument is fully ASCII, even when the absolute path contains extra-ASCII characters. For instance this works appropriately:
```bash
mkdir ~/Pensées
cd ~/Pensées
silverbullet .
```

pimarti:
Fixes #1988.

The space path was not correctly encoded when added to the HTTP header.


zefhemel:
Didn't realize this would trigger a fully resync of spaces because the path is now encoded for all users, this is not great so I'll revert this.

PR #2009, opened 2026-05-31
Fix inverted check in refreshLogin

Schmavery:
I noticed I was being prompted to log in about once a week even with "Remember me" checked. Looking at `refreshLogin` in `server/auth.go`, I think the guard is inverted: it returns early when the `refreshLogin` marker cookie is present, but that cookie is only set when "Remember me" was checked at login, so the refresh never runs for the sessions it was meant to extend.
Kinda seems like the comment didn't match the code in any case.

This flips the check to `== ""` so the function runs when the marker is present. Maybe I'm missing something if this is working for others!


Schmavery:
Looks like the prev PR here may have inverted the behavior to be an early return without inverting the check
https://github.com/silverbulletmd/silverbullet/pull/1804/changes#diff-7ef2fc1dd6fa1adf2d912d72bbc4d106db0873f4baef685f7a5f297a6844c701R261-R263
cc @strboul

zefhemel:
Hah

PR #2020, opened 2026-06-11
Fix: only treat known file extensions as document paths (page names with dots)

### LINKED ISSUE #1957: Wikilinks with dots do not work
Good day!

I have a lot of web-projects I usually refer to by domain names. So in SilverBullet they should be referenced as `[[site.com]]`. But this doesn't seem to work and clicking on such wikilink results in opening `https://mydomain.tld/.fs/site.com` with a 404 error in a new browser window (I have sb installed on my Windows desktop as a PWA).

Should this be considered a bug? Or is it a desired behavior? If so, what are my workarounds?

Thanks!

josh-j:
Fixes #1957

Implements the whitelist approach proposed by @zefhemel in the community thread [Page names with "." (again)](https://community.silverbullet.md/t/page-names-with-again/4083).

## Problem

`normalizePath` in `plug-api/lib/ref.ts` treats **any** dot-suffix as a file extension:

```ts
if (/.+\.[a-zA-Z0-9]+$/.test(path) || path === "") {
  return path as Path;   // treated as a document path, no .md appended
}
```

A page named `site.com` (or `My Notes 2.0`, `something.template`, any web-clipped article titled after its source domain) parses as a *document* ref, so navigation fetches `/.fs/site.com` literally and 404s — the page is unreachable by name even though `site.com.md` exists. In my space (with many imported web clippings titled `… « SiteName.com`) this affected 700+ pages.

## Fix

Per the proposal in the thread: a name ending in a dot-suffix is only a document path when the suffix is a **known file extension**; everything else is a page name that gets `.md` appended as usual. The whitelist covers images, documents, audio/video, web/code, archives, and SilverBullet's own types (`canvas`, `excalidraw`, `conflicted`).

`[[my.template]]` now resolves to `my.template.md`, while `[[diagram.png]]` still resolves to the document.

Making the whitelist configurable (as floated in the thread) would be a natural follow-up; this PR keeps it static to stay small.

## Testing

- Updated the `parseToRef(" .foo")` expectation in `ref.test.ts` (now a page, `" .foo.md"`) and added a case for a domain-suffixed page name.
- `npx vitest run` — 873 passed.
- Deployed on my own instance: previously-404 pages now open by name; document links (png/pdf/etc.) unaffected.

### DIFF
--- a/plug-api/lib/ref.test.ts
+++ b/plug-api/lib/ref.test.ts
@@ -16,7 +16,12 @@ test("parseToRef() default cases", () => {
   expect(parseToRef("foo.md")).toEqual({ path: "foo.md" });
   expect(parseToRef("foo.")).toEqual({ path: "foo..md" });
   expect(parseToRef("foo..")).toEqual({ path: "foo...md" });
-  expect(parseToRef(" .foo")).toEqual({ path: " .foo" });
+  // "foo" is not a known document extension, so this is a page name with a dot
+  expect(parseToRef(" .foo")).toEqual({ path: " .foo.md" });
+  // Page names ending in extension-like suffixes are pages, not documents
+  expect(parseToRef("Clippings/Article StrengthPlanet.com")).toEqual({
+    path: "Clippings/Article StrengthPlanet.com.md",
+  });
   expect(parseToRef("foo[bar")).toEqual({ path: "foo[bar.md" });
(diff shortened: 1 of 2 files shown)


zefhemel:
Thanks for this, but my conclusion (for myself) at the end of was to park this topic and not do this for now.

PR #2022, opened 2026-06-11
Perf: increase space sync concurrency from 3 to 8

josh-j:
## Problem

`SpaceSync.syncFiles` fans out with `processWithConcurrency(..., syncConcurrency)` where `syncConcurrency = 3`. Browsers multiplex these requests over a single HTTP/2 connection, so 3 in flight leaves most of the pipe idle — the initial sync of a larger space is dominated by sequential round-trip latency rather than bandwidth.

## Change

Raise `syncConcurrency` to 8. On my ~1,800-file space this roughly halved initial-sync wall-clock time; the effect grows with client→server latency (reverse proxies, mobile networks). Memory cost is bounded (at most 8 file bodies in flight), and the server side handles the parallelism fine (each request is an independent file read).

## Testing

`npx vitest run` — 873 passed. Running on my own instance through a reverse proxy (Traefik) and from mobile clients with no sync conflicts or regressions observed.

### DIFF
--- a/client/spaces/sync.ts
+++ b/client/spaces/sync.ts
@@ -7,7 +7,9 @@ import type { FileMeta } from "@silverbulletmd/silverbullet/type/index";
 import { notFoundError } from "@silverbulletmd/silverbullet/constants";
 import { processWithConcurrency } from "@silverbulletmd/silverbullet/lib/async";
 
-const syncConcurrency = 3;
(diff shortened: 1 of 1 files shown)


zefhemel:
Thanks, and perhaps you can adapt this with a configurable option somehow (server side variable maybe). There's a reason I picked 3, it's based on not overwhelming especially less powerful mobile devices with too much traffic.

Issue #2016, opened 2026-06-08
Add config to prioritize note name over full path in title and link search when out of space


LanHikari22:
Hi,

I have long note names, and equally long paths to them in my space. For example:

```
lan/2026/proj/003-clusterline-md/entry/000-clusterline-md-getting-started/task/001 create a rust silverbullet plugin
```

The title for this note only shows me `lan/2026/proj/003-clusterline-md/entry/000-clusterli` which is not useful to look at. I would like an option to either truncate the path and only show the title, or in case the name is too long to display, prioritize to show only the note name.

Similarly, when I want to link a note and use [[...]], I see long path names which makes picking a note searched hard.

There could also be an option to make the text multiline instead of truncating.

In the spirit of keeping the default setup minimal and pushing changes to plugs, we can also dispatch an event for a plug to modify the path that appears in the title, and issue an error if multiple event responses come back to disallow multiple plugs of a conflicting implementation. This is a suggestion inspired from having read this part in [client/content_manager.ts](https://github.com/silverbulletmd/silverbullet/blob/342686b2fc79df2487e1ab36b49216958edcaee4/client/content_manager.ts#L317).

Thanks,
Mohammed

PR #2023, opened 2026-06-14
Allow title text to be configurable via service and add opt in base name only option

LanHikari22:
Hi,

Per #2016, I implemented a service that allows us to customize the title bar to include only the name of the page and not the full path.  This can be overridden by other plugs that implement the service at higher priority.  

I also ran `make fmt` as instructed in `CONTRIBUTING.md` and this ended up changing 118 files. I did this in its own commit so please advise if I need to revert this.

Thanks,
Mohammed

---

[Internal dev notes](https://codeberg.org/deltatraced/deltatraced/src/commit/1abdd2cbf4813fefe66fc5e80bc8f2ad95fe471b/lan/2026/topic/oss/000%20OSS%20Contrib/task/000%20SB%20Option%20to%20only%20have%20base%20name%20in%20title/entry/001%20Overview%20SB%20Option%20to%20only%20have%20base%20name%20in%20title.md)


zefhemel:
I know the fmt hasn't been run for a while and will do this soon, but the current state of this PR makes it impossible to review.

PR #2057, opened 2026-07-19
Fix transclusions of custom syntaxes not closing html tags

### LINKED ISSUE #2025: Syntax' renderHtml doesn't close elements properly
I've written a custom syntax that renders discord conversations like this:

Image

But I've noticed that runs the renderWidget flow, and if I put that on a page then render that page as part of a transclusion, it uses the renderHtml flow. In that flow it looks like this:

Image

The code for the two flows is essentially identical though:

```
syntax.define {
  name = "Discord",
  startMarker = "^::discord$",
  endMarker = "^::$",
  mode = "block",
  renderClass = "sb-discord-preview",
  renderWidget = function(body, pageName)
    return widget.htmlBlock(renderDiscordConversation(body))
  end,
  renderHtml = function(body, pageName)
    return renderDiscordConversation(body)
  end
}
```

By looking at the DOM, I see in the widget flow the elements all look how I'd expect. Of particular note, I have a header and two messages as siblings of each other:

Image

But in the html flow, it seems the elements aren't being closed properly, and each sibling is actually nested within the previous element:

Image

thepaperpilot:
Fixes #2025

### DIFF
--- a/client/codemirror/inline_content.ts
+++ b/client/codemirror/inline_content.ts
@@ -15,17 +15,20 @@ import {
   expandMarkdown,
   readTransclusionContent,
 } from "../markdown_renderer/inline.ts";
+import { renderMarkdownToHtml } from "../markdown_renderer/markdown_render.ts";
 import {
   isLocalURL,
   resolveMarkdownLink,
 } from "@silverbulletmd/silverbullet/lib/resolve";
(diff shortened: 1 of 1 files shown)


zefhemel:
Thanks, this looks like a reasonable solution.

PR #2062, opened 2026-07-22
Fix shell disabled when SB_SHELL_BACKEND=local (#2058)

### LINKED ISSUE #2058: shell.run() returns "Not supported" even with SB_SHELL_BACKEND=local (2.9.0)
I'm running SilverBullet in Docker and have a few Space Lua scripts that need to run local commands via `shell.run()`. Space is the normal local `/space` folder, nothing unusual there.

I set `SB_SHELL_BACKEND=local` in my compose file, but `shell.run()` still comes back with `Not supported`. I double-checked and the running process really does have the variable set — `/proc/<pid>/environ` shows `SB_SHELL_BACKEND=local`, so it's not getting dropped somewhere. I also tried hitting `/.shell` directly and got the same thing, so I don't think it's a Lua or plug problem on my end.

Version:

```
2.9.0-0-g72bba941-2026-06-11T07-04-08Z
```

Compose snippet:

```yaml
environment:
  SB_SHELL_BACKEND: local
```

It's not just one command — even a plain `echo` fails. To reproduce, just run this in a page:

```lua
local result = shell.run("echo", {"hello"})
```

I'd expect it to run `echo hello` and give me the output, since `local` is documented as the backend that enables shell commands. Instead I get:

```
stderr: Not supported
code: 1
```

Same result from a direct `POST /.shell`:

```json
{"cmd":"echo","args":["hello"]}
```

Am I missing a step, or is something off with the `local` backend in 2.9.0? Docs I was going off of:

- https://v2.silverbullet.md/API/shell
- https://v2.silverbullet.md/Install/Configuration

Thanks.

bong-u:
Fixes #2058.

`ShellConfig::parse` enabled shell running only when `SB_SHELL_BACKEND` was
unset (`enabled = backend.is_none() && !read_only`), so the documented
`SB_SHELL_BACKEND=local` set `backend.is_some()` and disabled the shell.

Switch to allow-list semantics matching the docs (defaults to `local`, `off`
to disable): enabled only for `local` (the default when unset, case-insensitive);
any other value fails safe to disabled; read-only always disables.

Added regression tests; `cargo test`/`fmt`/`clippy` pass.

### DIFF
--- a/server/src/shell.rs
+++ b/server/src/shell.rs
@@ -8,9 +8,9 @@ pub struct ShellConfig {
 }
 
 impl ShellConfig {
-    /// Build from the environment. Shell running is enabled only when
-    /// `SB_SHELL_BACKEND` is unset AND the space is not read-only;
-    /// `SB_SHELL_WHITELIST` is a space-separated allow-list.
+    /// Build from the environment. `SB_SHELL_BACKEND` defaults to `local`
+    /// (enabled); any other value disables shell running. `SB_SHELL_WHITELIST`
+    /// is a space-separated allow-list.
     pub fn from_env(read_only: bool) -> Self {
         let backend = std::env::var("SB_SHELL_BACKEND")
             .ok()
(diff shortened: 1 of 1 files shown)


zefhemel:
Thanks for this. I'm redoing this part of the code a fair bit, let me merge those changes with this and see where it lands ok?

bong-u:
> Thanks for this. I'm redoing this part of the code a fair bit, let me merge those changes with this and see where it lands ok?

Sure, sounds good. thanks!

zefhemel:
Ok, looks like what you did here is immediately usable so merging it after all. Thanks!

PR #2067, opened 2026-07-24
Add openssh-client to docker image

### LINKED ISSUE #2066: Can't use git autosync with ssh because Docker is missing openssh-client
# Summary
Absolutely new user here. :) I am running silverbullet docker in unraid. I want to enable git autosync, but use ssh key instead of a github PAT. 

# Expected
* git pull/push work from within the app (through git plug) using ssh keys

# Actual
* git pull/push does not work as there is no sshclient within docker.

# Workaround
I am running the git push/pull from the host right now using userscripts (cron like).

neupsh:
Alpine's git package doesn't pull in an ssh client, so git autosync over ssh fails inside the container with `ssh: not found`. Adds `openssh-client` alongside the existing packages.

Tested locally: with `~/.ssh` mounted in, `git clone`/`ls-remote` over ssh work fine after this change; without it they fail with the missing ssh error.

Closes #2066

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ FROM alpine:latest
 # binary (silverbullet-amd64 / silverbullet-arm64, built on the CI runner).
 ARG TARGETARCH
 
-RUN apk add --no-cache git curl bash tini
+RUN apk add --no-cache git curl bash tini openssh-client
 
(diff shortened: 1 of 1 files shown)


neupsh:
hi @zefhemel , would love to have this merged if you can review and find it a non-issue. TIA.

PR #2072, opened 2026-07-29
Fix Runtime API startup when authentication is enabled

TiltAbricot:
## Summary

- seed the server-managed headless authentication token as a same-origin, HTTP-only session cookie before navigating Chromium
- keep query-token authorization for compatibility with older runtime clients
- scope the cookie to the active space prefix and remove the token from the navigation URL
- make the Runtime API end-to-end test run with authentication enabled and verify anonymous requests remain unauthorized

## Problem

With `SB_USER` enabled, the initial headless page load was authorized by the `token` query parameter, but subsequent client requests did not carry it. The client was redirected to `/.auth`, the CDP bridge disconnected, and Runtime API calls returned `bridge_unavailable`.

## Validation

- `make fmt`
- `npm run check`
- `npx biome lint .`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `npx vitest run` (1,162 passed; 5 skipped)
- `cargo test --workspace --all-features`
- authenticated Chromium end-to-end test: `/.runtime/lua` returns `{"result":2}` while the anonymous request returns HTTP 401

### DIFF
--- a/bin/silverbullet/tests/runtime_e2e.rs
+++ b/bin/silverbullet/tests/runtime_e2e.rs
@@ -23,6 +23,18 @@ impl Drop for Server {
 mod common;
 use common::free_port;
 
+const AUTH_TOKEN: &str = "runtime-e2e-token";
+
+fn chrome_available() -> bool {
+    std::env::var("SB_CHROME_PATH")
+        .ok()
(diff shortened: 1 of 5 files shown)


zefhemel:
Thanks I ended up folding this in with the recent refactor to reuse a single Chrome instance across spaces.

PR #2077, opened 2026-08-01
Fix: proper progress status update

funbiscuit:
This PR fixes several problems with progress status widget:

1. Sync and index showProgress fire at the same time leading to sync progress blink (because index calls hide progress)
2. Index progress updates once a second which is not as smooth as sync progress which shows each percent
3. Index progress might get stuck under 100 because of 1s sleep and indexing is done after sleep finished. So UI shows 98 and clears only after 5s timeout

Also I made progress type required since UI needs it anyway. And added extra api function hideProgress so we don't need to send undefined to mark "no progress" (syscall is reused).

This is how it looked before (all three bugs are present):

https://github.com/user-attachments/assets/072b2971-d9ea-4920-b915-0aab7c78049f

And this is now (no blinking, index is updated smoothly and hidden as soon as index is completed):

https://github.com/user-attachments/assets/5f603b9d-d395-4f7d-82e8-2d8c74153a26

### DIFF
--- a/client/editor_ui.tsx
+++ b/client/editor_ui.tsx
@@ -111,7 +111,13 @@ export class MainUI {
   }
 
   // Progress circle handling
-  private progressTimeout?: ReturnType<typeof setTimeout>;
+  private progressMap = new Map<
+    "index" | "sync",
+    {
+      percentage: number;
+      timeout: ReturnType<typeof setTimeout>;
+    }
+  >();
 
   viewDispatch: (action: Action) => void = () => {};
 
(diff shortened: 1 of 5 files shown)


zefhemel:
Right. It's one of those things I've also noticed but never got around to doing something about. Thanks!

funbiscuit:
Changed order of `percentage` and `type` parameters in signatures (so `npm run check` passes). I think it's more "semantically correct" to pass type first and then optional progress. Looks like all places are internal API so I think it's okay to make such changes.

PR #2086, opened 2026-08-08
Run lint/fmt in CI

funbiscuit:
## Problem

CI currently runs type-checks and tests but skips lint and formatting for both client and server. So
formatting-breaking commits land on `main` regularly.
`make check` already covered everything locally (but not client formatting), but CI only ran a
subset (`npm run check` + tests) and reimplemented it inline, so the two drifted apart.

| Check                      | `make check` | `ci.yml` |
|----------------------------|--------------|----------|
| `tsc --noEmit`             | yes          | yes      |
| `biome lint`               | yes          | **no**   |
| `biome format`             | **no**       | **no**   |
| `cargo fmt --check`        | yes          | **no**   |
| `cargo clippy -D warnings` | yes          | **no**   |

## Solution

- Added `npx biome format .` to the `make check` (since cargo fmt was already there)
- Added separate `lint` job that runs `make check` in `ci.yml`
- Added `cache: 'npm'` to all `setup-node` steps (it's a bit faster with it on subsequent runs)
- Ran `make fmt` and committed changes to separate commit

## Suggestion

For this to actually block bad commits, the `lint` job needs to be marked as a **required status check**
under **Settings → Branches → branch protection rules → `main`**.
Without that it's advisory only. The existing `test-*` jobs likely want the same rule.

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -20,6 +20,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version-file: ".nvmrc"
+          cache: 'npm'
 
       - name: Setup Rust (linux musl + windows-gnu + freebsd targets)
         uses: dtolnay/rust-toolchain@stable
@@ -102,6 +103,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version-file: ".nvmrc"
+          cache: 'npm'
 
       - name: Setup Rust
         uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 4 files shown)


funbiscuit:
Rebased commits onto new main. `make fmt` already includes `npx biome format .` but changes to ci (new `lint` job that checks formatting/lints) still needed. And second commit again reformats existing code - it drifted again a little bit.

I also noticed there is `biome check` command but it is not mentioned anywhere. And running it results in a lot of complaints so I didn't fix those. Let me know if you think this should also be added to lint job and Makefile

zefhemel:
Implemented on main, differently

Issue #2087, opened 2026-08-10
SpaceLua API/Template: Markdown Highlight/mark does not render in tasks/checkbox items


SohamG:
# Reproduce (v2.10)
## Space Lua
```space-lua
example={}
example.t1 = template.new([==[ 
* [ ] # ${a}
]==])
example.t2 = template.new([==[ 
* [ ] _${a}_
]==])
example.t3 = template.new([==[ 
* [ ] **${a}**
]==])
example.t4 = template.new([==[ 
* [ ] ~~${a}~~
]==])
example.t5 = template.new([==[ 
* [ ] <mark>${a}</mark>
]==])
example.t6 = template.new([==[ 
* [ ] ==${a}==
]==])
example.t7 = template.new([==[ 
* [ ] *${a}*
]==])
example.t8 = template.new([==[ 
* [ ] `${a}`
]==])
```
## Queries
‘\#‘: ${example.t1 {a="foo"} }
‘_’: ${example.t2 {a="foo"} }
‘\*\*‘: ${example.t3 {a="foo"} }
‘~~‘: ${example.t4 {a="foo"} }
‘mark tag’: ${example.t5 {a="foo"} }
‘\=\=’: ${example.t6 {a="foo"} }
‘\*’: ${example.t7 {a="foo"} }
‘\`’: ${example.t8 {a="foo"} }

## Outcome

<img width="1080" height="803" alt="Image" src="https://github.com/user-attachments/assets/6e7dd19b-9538-4e26-9c2f-d66fca005f76" />

# Expected
Text wrapped with `==` renders with a highlighted background (same as <mark>mark</mark> tag) 

<img width="191" height="46" alt="Image" src="https://github.com/user-attachments/assets/6bdd5e0f-f444-4cb1-abfd-07e496ca1eb3" />

# Observations
- Using `<mark>` and == works when not in query
- Using the tag renders the tag verbatim
-  Since other inline markdown styles are supported, this IMO is an inconsistency

# setup
Linux. firefox 153.0.1 running silverbullet v2.10 from gh release  locally/accessing from localhost. Reproduced on v2.9.0 from repo package.

zefhemel:
I think this was just fixed, let me know if not

PR #2088, opened 2026-08-10
Fix: highlight not rendering in task items

funbiscuit:
## Problem

Described in #2087.

`==text==` is rendered correctly in the editor (CodeMirror live preview) but appears as unstyled plain text in query
results.

Root cause: Markdown renderer emitted `<span class="highlight">`, but no `.highlight` CSS rule actually exists.
Editor uses `.sb-highlight`.

## Solution

1. Changed `class: "highlight"` to `class: "sb-highlight"` so `==` now works in queries.
2. Added call to `groupInlineHtml()` in `Task` renderer case similar to how `Paragraph` does it.
3. Added small unit test that checks that `Task` renders correctly with inline HTML.

### DIFF
--- a/client/markdown_renderer/markdown_render.test.ts
+++ b/client/markdown_renderer/markdown_render.test.ts
@@ -177,6 +177,17 @@ test("Unmatched HTML tags render as literal text", () => {
   expect(html).toEqual('<span class="p">text &lt;b&gt;unclosed</span>');
 });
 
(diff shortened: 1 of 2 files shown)


SohamG:
Thanks I appreciate the quick action! I'll let someone more experienced with the codebase comment on the patch and hope it makes it to a silverbullet near me soon :)

PR #2089, opened 2026-08-17
Fix: stop applying workspace .gitignore rules

### LINKED ISSUE #2074: Version 2.10 (presumably erroneosly) applies ignore rules from `.gitignore` file
Hi,

first off, massive thanks for this great project! It is admittedly one of those I didn't even have a purpose for before spinning it up. I just liked the whole idea so much that I did and since then find excuses to use SilverBullet whenever possible. :)

The issue I'm encountering started when upgrading from 2.9.0 to 2.10.0. I didn't try any of the `edge` builds.

**Update:** Ok turns out the below was merely a symptom. The real cause is the `.gitignore` file in the root of my space. Before 2.10, it wasn’t interpreted by SilverBullet (and I suspect, it shouldn’t be now, either?). Among others, `Library/`, `_plug/`, and `Repositories/` are in there to be excluded from git versioning. As soon as I remove this file, old libraries show up and are persistent.

~~With 2.10.0, none of the installed libraries  show up or work. Same is true for one old plug I have among the files, although I caught wind those are falling out of support.~~
~~I went ahead and re-installed them via configuration manager. Many times, actually. Sometimes, the library will show up as installed, at least briefly, sometimes not. Sometimes it works (e.g. SilverSearch, btw.) for a short while. I cannot exactly say what has to happen for a library to "vanish" again, but they all do so eventually. Meaning library functionality and commands are not available any longer. Often even after a simple page change or reload. When they do, they are gone from configuration manager, as well.~~

~~I checked with `find` and a 24h-in-the-past limit to see if any file on the server changes when installing a library in 2.10: none of the (already installed, but defunct) library files is touched. Only when fully deleting the `Library` folder are new files written. Although this doesn't change the described behavior.~~
~~In the course of this, I also tried removing all "old" libraries and plugs to see whether it makes any difference - it does not.~~

~~I'm running a pretty standard install, `docker` on `arm64`. Nothing fancy in my library root, only files created with SilverBullet and a couple of attached images. I'm running the image as a non-root user.~~
~~When switching back to 2.9.0, everything is working again instantly.~~

~~I checked and re-checked, but I cannot find any mention of necessary manual migrations between those two versions. Did I miss something?~~

atirna:
## Problem

A `.gitignore` in a space root was being applied alongside `SB_SPACE_IGNORE`, making workspace files such as `Library/` disappear from SilverBullet.

## Fix

Use only the configured `SB_SPACE_IGNORE` patterns for SilverBullet visibility. Workspace `.gitignore` files remain available for Git without hiding space files.

## Tests

- Added regression coverage for a root `.gitignore` while retaining configured ignore behavior.
- `npm run build`
- `make test`
- `make check`

Fixes #2074


zefhemel:
sorry, missed this one and had already fixed this here: https://github.com/silverbulletmd/silverbullet/commit/1294a3a86a5b91c6978fd09c4d32baaea163839b

PR #2116, opened 2026-09-13
Index anchors inside table cells

### LINKED ISSUE #2104: Anchors inside table cells are parsed but never indexed, so `[[$anchor]]` reports "Anchor not found"
### Version

2.10.0 (`2b2a7c7`)

### What happens

A `$anchor` placed inside a markdown table cell is **parsed but never indexed**, so `[[$anchor-name]]` reports `Anchor not found`. Anchors in headings, paragraphs and list items work correctly.

### Minimal reproduction

```markdown
| Case | Content |
| :--- | :--- |
| in a cell | $Anchor-In-Table some text |

$Anchor-In-Paragraph a paragraph hosting an anchor.

- [ ] a list item $Anchor-In-Listitem
```

Then query the index:

```lua
from a = index.tag "anchor" where a.page == "<that page>"
```

**Expected:** three anchors.
**Actual:** two — `Anchor-In-Paragraph` and `Anchor-In-Listitem`. The table one is absent, and `[[$Anchor-In-Table]]` fails to resolve.

### Other notes
The anchor is correctly displayed as a formatted anchor inside the table, and repeating the same anchor in a page twice does give a "Duplicate anchor "$xyz" — defined 2 times on this page.", so it's otherwise seemingly handled correctly as an anchor in that context.

### Why (as far as I can tell)

The anchor *does* parse inside a cell — `NamedAnchor` is an inline node, and running `markdown.parseMarkdown` over the example above yields `NamedAnchor` nested inside `Table`. So this is not a parser limitation.

It is never **collected**. `collectAnchor()` is invoked from exactly three places, none of which visits table cells:

- `plugs/index/header.ts:46` — `ATXHeading`
- `plugs/index/paragraph.ts:53` — `Paragraph`
- `plugs/index/item.ts:147` — `ListItem`

Because the anchor is missing from the index, `index.resolveAnchor` returns `missing` and `client/client.ts:943` reports `Anchor not found` — i.e. the failure presents as a navigation problem but is really an indexing gap.

### Possible fix

`plugs/index/table.ts` already indexes table rows as objects (`TableRowObject` carrying `page` and `pos`), so a host with a position already exists — a fourth `collectAnchor()` call site there would follow the existing pattern.

I'm happy to open a PR if that approach looks right to you, or if you'd prefer the anchor to attach to something other than the row object.

anandghegde:
Fixes #2104.

@ArchieBinnie, thanks for the detailed report. You offered to open a PR for this, and I don't want to step on that. This follows the approach you proposed (a fourth `collectAnchor()` call site in `plugs/index/table.ts`). If you already have a patch in progress, I'm happy to close this in favor of yours.

## Problem

A `$anchor` inside a table cell is parsed as `NamedAnchor` but never collected. So there's no `anchor` record, and `[[$Anchor-In-Table]]` reports "Anchor not found". Anchors in headers, paragraphs and list items work, because `collectAnchor()` is only called from `header.ts`, `paragraph.ts` and `item.ts`.

## Fix

`indexTables` already emits one `table` object per row, with `page` and `range`. The row now collects its first anchor with `collectAnchor(row)` and uses the anchor name as its `ref`, the same "first anchor wins" pattern as paragraphs and items. `appendAnchorRecords` then emits the `anchor` record (`hostTag: "table"`), and `resolveAnchor` finds the host row via `getObjectByRef(page, "table", name)` and returns its range.

Rows without an anchor keep their `Page@pos` ref.

## Tests

- Added `resolves an anchor inside a table cell (#2104)` to `plugs/index/resolve_anchor.test.ts`. It uses the table from the issue and goes through `indexMarkdown` and `resolveAnchor`, checking `hostTag` is `table` and the range is the anchored row. It fails on `main` (`result.ok` is false) and passes with this change.
- `npm run build`, then `npx vitest run`: 191 test files, 2577 tests pass (5 skipped).
- `npx tsc --noEmit` and `biome format` are clean.

## Notes

- An anchor in the table **header** row isn't a `TableRow`, so it still isn't indexed. That seemed like an odd place for an anchor, so I left it alone.
- The anchor stays in the cell's column value (e.g. `content` is still `$Anchor-In-Table some text`). Paragraphs and items strip the anchor from their displayed text. I kept this change to the indexing gap, but can strip it from cell values too if you prefer.

This PR was written with Claude Code (Opus). I reviewed the change and ran the tests above.

https://claude.ai/code/session_01M8rNAK3JEZQNVxrnQ8SVGG


ArchieBinnie:
@anandghegde no concerns about stepping on toes - I was waiting on any response to the bug report's options, and didn't get round to coding a fix on my side, so I'm happy if yours is merged.

anandghegde:
Thanks @ArchieBinnie — appreciated, and thanks for the report that pinned down the missing `collectAnchor()` call site.

Flagging the two open questions from the description for whoever reviews, since both are deliberate and easy to change:

- An anchor in a table **header** row still isn't indexed; a header isn't a `TableRow`, and it seemed an unlikely place for an anchor.
- The anchor text stays in the cell's column value, where paragraphs and list items strip it from their displayed text. I kept this PR to the indexing gap, but stripping it from cell values is a small follow-up if that inconsistency is worth closing.

Issue #2105, opened 2026-08-31
Navigator does not get auto focused on launch to enable typing


LogeshG5:
In edge release, once navigator is launched, textbox does not get autofocused. Once Page picker is launched, you have to click on the textbox to start typing.

zefhemel:
Is this still happening? I can't reproduce.

a-newton:
It's still happening for me on edge (tested just now in Waterfox and Vivaldi).

I've also noticed that if the phone's on screen keyboard is visible at the point when the navigator is opened (either via the two finger click gesture or via a menu button) then the on screen keyboard remains open but the input focus remains on the page hidden behind the navigator modal. This means that if you forget to refocus the modal input text box then you start to accidentally write over the contents of the page underneath.

zefhemel:
Hang on are we all talking about mobile browsers here?

a-newton:
Yes, it's only happening on mobile for me.
Sorry, I was sure it mentioned it was a mobile issue in the original post 🤦

LogeshG5:
### Environment Information

* **SilverBullet Version:** `2.11.0-3-gb7548912-2026-09-17T13-25-29Z`

#### Desktop
* **OS:** Windows
* **Browsers Tested:** Edge, Chrome, Firefox

#### Mobile
* **Device / OS:** Android
* **Browsers Tested:** Chrome

#### Screenshots

1. Upon opening (Cursor still in editor and not in input field), 

<img width="529" height="228" alt="Image" src="https://github.com/user-attachments/assets/4098389a-6051-46e9-8145-b93fd3884774" />

2. After I click on the input field, 

<img width="529" height="228" alt="Image" src="https://github.com/user-attachments/assets/105a1cc0-c628-4ba4-bee0-47cfba77d916" />

PR #2118, opened 2026-09-14
fix: focus the page picker after paint-gated reveal (#2105)

Keyne:
The navigator modal opens with .sb-modal-paint-pending (opacity 0, pointer-events: none). Focusing the filter during that window is dropped by Chromium, so Cmd-K required a click before typing. Wait until the pending class is gone, then focus.


zefhemel:
Fixed as part of #2144

PR #2144, opened 2026-09-19
fix: focus modal filters after keyboard shortcuts on tablets

### LINKED ISSUE #2131: No longer auto-focuses on the modal after using a keyboard shortcut
I updated to `2.11.0` today, and now when I use any keyboard shortcut to load any modal (control + /, K, P, etc), it no longer focuses on that text input for the modal.

So now you use the keyboard shortcut, then have to manually click / touch the modal text input.

dyk1454683243-sudo:
**Fixes #2131**.

After 2.11.0, keyboard shortcuts that open a modal (Cmd/Ctrl+/, page picker, `editor.filterBox`) no longer put the caret in the filter on iPad and Android tablets with a hardware keyboard. `focusInput` and `FilterList` no-op on every `isMobileDevice()` (coarse primary pointer), which those tablets report even when a keyboard is attached.

This keeps the original tap-open skip on pure-touch phones: programmatic `.focus()` leaves the field focused with no on-screen keyboard, and no tap can recover one. A modifier chord (the shortcut that opened the modal, or any earlier one this session) is treated as a physical keyboard, and the filter is focused and given `autofocus`.

## Coordination with #2118

#2118 (for #2105) retries navigator focus after the paint-gated reveal. This PR does **not** take that on: it only changes the mobile-skip / autofocus path. `shouldFocusModalFilter()` lives in `client/lib/mobile.ts`, which #2118 does not touch, so the two can land independently. After both merge, `focusInput` should keep this guard and #2118's retry.

## Test plan

- [x] `npx vitest run client/lib/mobile.test.ts` — coarse pointer skips focus; Cmd/Ctrl/Alt chords enable it; plain typing and IME composition do not.
- [x] `npx biome lint` on touched files
- Desktop Chromium e2e for “type immediately after Cmd-K” is left to #2118 (paint-pending). This change is the tablet/keyboard skip.
- Manual: iPad or Android tablet + keyboard → Cmd+/ (and Cmd+K) → type immediately without tapping the filter. Phone, tap-open a picker → filter should not steal focus / swallow the OSK.


zefhemel:
Just tested this on my phone with external keyboard attached and it works great!

